PR #6, opened 2016-12-28
Write custom image field

slomo:
- reads image properties from settings
   - setups validators accordingly
   - generates help texts automatically
- setup our own upload widget


slomo:
I suspect the used markup should be reworked, because it relies quite a bit on how it works in euth. I would nevertheless merge it first and improve on it later on. We could also think about adding some css here.

PR #11, opened 2017-01-11
add comments

vellip:


adhocracy4/comments/services.py
```diff
@@ -0,0 +1,8 @@
+from .models import Comment
+
+
+def delete_comments(contenttype, pk):
+    comments = Comment.objects.all().filter(
+        content_type=contenttype, object_pk=pk)
+    for comment in comments:
+        comment.delete()
```
xi: What is this used for?

vellip: @MagdaN you know that?

MagdaN: @xi the code gets all comments for a resource and deletes them, e.g. it is used when an idea is deleted it is used to delete all comments of that idea

xi: Ok. But where is it called? More specifically: In which cases is this used and in which cases is the code in adhocracy4/comments/signals.py used?

2e2a: signaly.py registers a delete handler for the post_delete signal.

However, @slomo has created a generic function for this:
`generics.setup_delete_signals(settings.A4_COMMENTABLES, Comment)`

I'll create a pull request.

PR #15, opened 2017-01-16
declare js dependencies

xi:
I copied all dependencies that are actually used from euth-wagtail


MagdaN:
@vellip can you take a look at this?

PR #23, opened 2017-01-25
comment markup improvements

xi:
While trying to style the a4 comments for meinberlin I found some issues with the markup:

- `<a href="#">` was used instead of `<button>`
- `class="icon fa-*"` was used instead of a `<i class="fa fa-*">` child (see [CSS in euth_wagtail](https://github.com/liqd/euth_wagtail/blob/40dd0be976073e92ca97a9041abcdc5c6a8c2350/euth_wagtail/assets/scss/components/_comments.scss#L73))
- `aria-hidden="true"` was used for essential elements

### DIFF
--- a/adhocracy4/comments/static/comments/CommentManageDropdown.jsx
+++ b/adhocracy4/comments/static/comments/CommentManageDropdown.jsx
@@ -12,15 +12,15 @@ const CommentManageDropdown = (props) => {
         <ul className="dropdown-menu">
           {props.renderModeratorOptions && [
             <li key="1">
-              <button onClick={props.toggleEdit} aria-hidden="true">{django.gettext('Edit')}</button>
+              <button onClick={props.toggleEdit}>{django.gettext('Edit')}</button>
             </li>,
             <li className="divider" key="2" />,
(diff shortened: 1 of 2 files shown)


MagdaN:
@xi did you test this? This might break some stuff, e.g. using buttons instead of anchors might cause some problems.

xi:
Sorry, I should have said this in the original description: I could not test it as comments do not work yet in a4-meinberlin.

I don't expect any breakage in terms of functionality, but the styling in euth_wagtail will probably need to be adapted.

MagdaN:
@xi I remember having some functional issues with using buttons instead of anchors, maybe you just test this

xi:
Ok, I did the test. Works fine for me.

xi:
I added some more classes to make styling easier:

- `dropdown-item` (which also exists in bootstrap4) (c25db1d)
- `comment-reply-button` (6d74c81)

slomo:
works on opin (even without using `dropdown-item` but with `comment-reply-button`) see https://github.com/liqd/euth_wagtail/tree/2017-02-adopt-to-new-comment-markup

PR #24, opened 2017-01-30
make project moderators optional

2e2a:


slomo:
Can you elaborate why this was needed, and why it is a good idea? I think there might be a few other spots  in the code that rely on the moderator always being present.

2e2a:
> Can you elaborate why this was needed, and why it is a good idea? I think there might be a few other spots in the code that rely on the moderator always being present.
 
This was a simplification to allow projects being managed by initiators or even admins, e.g. if a project is not created/managed by a customer, but by us.

MagdaN:
@slomo maybe I was too quick merging this, I assumed it would work for us, as we always set the creator of a project as moderator.

PR #25, opened 2017-01-30
improve rating markup

xi:
- hide icons for assistive technologies
- use simpler markup (before it used a lot of bootstrap helpers for a rather simple design)
- use the `disabled` property instead of the custom `is-read-only` class

Note that there are still some accessibility issues. The `isSelected` state and the overall semantics of the component are not yet properly exposed. We should definitely come back to that at some point.

I tested this with a4-meinberlin and it worked fine. (It needs CSS adjustments though)


slomo:
Nice simplification of markup. I'll have to check if styling can be still done on opin (with bootstrap) before merging.

PR #26, opened 2017-01-31
add generic sort view

2e2a:
Extends the ordering functionality of the Django `generic.ListView`:

- Applies sort from the request parameter
- Adds supported orderings dict list
- Provides function get the ordering name

### DIFF
--- a/adhocracy4/contrib/apps.py
+++ b/adhocracy4/contrib/apps.py
@@ -1,6 +1,6 @@
 from django.apps import AppConfig
 
 
-class OrganisationsConfig(AppConfig):
+class ContribConfig(AppConfig):
     name = 'adhocracy4.contrib'
     label = 'a4contrib'
--- a/adhocracy4/contrib/apps.py
+++ b/adhocracy4/contrib/apps.py
@@ -1,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class OrganisationsConfig(AppConfig):
+    name = 'adhocracy4.contrib'
+    label = 'a4contrib'
--- a/adhocracy4/contrib/views.py
+++ b/adhocracy4/contrib/views.py
@@ -24,7 +24,7 @@ class SortableListView(generic.ListView):
     """Add sorting via request parameter.
 
     Fields:
-        ordering: List containing the signle default ordering string
+        ordering: List containing the single default ordering string
         orderings_supported: List of possible ordering tuples (ordering,
             ordering name)
     """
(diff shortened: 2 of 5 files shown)


2e2a:
will fix test coverage.

2e2a:
> will fix test coverage.

Would need to setup integration tests for this. I suggest to merge this and postpone the tests.

adhocracy4/contrib/views.py
```diff
@@ -18,3 +19,20 @@ def raise_exception(self):
...
         """
         return self.request.user.is_authenticated()
+
+
+class SortableListView(generic.ListView):
+    """
+    ordering_current: List of possible ordering tuples (ordering,
+        ordering name).
+    """
+    orderings_supported = []
```
xi: Why is this not a dict in the first place?

adhocracy4/contrib/views.py
```diff
@@ -18,3 +19,20 @@ def raise_exception(self):
         limited by the current phase.
         """
         return self.request.user.is_authenticated()
+
+
+class SortableListView(generic.ListView):
+    """
+    ordering_current: List of possible ordering tuples (ordering,
+        ordering name).
+    """
```
xi: I don't understand how this docstring is related to the code. `ordering_current` does not exist, or does it?

adhocracy4/contrib/views.py
```diff
@@ -18,3 +19,20 @@ def raise_exception(self):
...
+    orderings_supported = []
+
+    def get_ordering(self):
+        ordering = self.request.GET.get('ordering')
+        if ordering and ordering in dict(self.orderings_supported):
+            self.ordering = [ordering]
+        return self.ordering
```
xi: Getter functions should not have side effects. I think we should either set `self.ordering` in the beginning of the request (`__init__`?) or not set it at all and always use `get_ordering()`.

adhocracy4/contrib/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class OrganisationsConfig(AppConfig):
```
xi: Should be `ContribConfig`

adhocracy4/contrib/views.py
2e2a: The list is used in the sort tamplate to iterate over available sortings.

2e2a:
reflected comments and added some tests

adhocracy4/contrib/views.py
```diff
@@ -22,17 +21,22 @@ def raise_exception(self):
 
 
 class SortableListView(generic.ListView):
+    """Add sorting via request parameter.
+
+    Fields:
+        ordering: List containing the signle default ordering string
```
xi: typo: single

xi:
@MagdaN This just adds new functionality, so it should not break anything in euth_wagtail. However, you might want to consider using this for euth_wagtail.

PR #27, opened 2017-02-01
implement SortableListView.get_ordering()

xi:
see https://docs.djangoproject.com/en/1.10/ref/class-based-views/mixins-multiple-object/#django.views.generic.list.MultipleObjectMixin.get_ordering


2e2a:
lgtm

PR #34, opened 2017-02-09
add date input widgets

2e2a:
Copy `euth_wagtail/contrib/widgets` to core

**Incomplete:** Javascript and dependencies are missing.

### DIFF
--- a/adhocracy4/contrib/widgets.py
+++ b/adhocracy4/contrib/widgets.py
@@ -1,0 +1,54 @@
+from django.contrib.staticfiles.storage import staticfiles_storage
+from django.forms import widgets
+from django.utils import formats
+from django.utils.safestring import mark_safe
+from django.utils.translation import get_language
+
+
+class DateInput(widgets.DateInput):
+    class Media:
+        js = (staticfiles_storage.url('flatpickr.min.js'),
+              staticfiles_storage.url('js/dateTimeInput.js'))
+        css = {'all': [staticfiles_storage.url('flatpickr.min.css')]}
+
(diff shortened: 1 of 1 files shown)


MagdaN:
I fixed some stuff (basically using local times in browser but UTC in DE) for this in our release2.0 branch. I would suggest to use this https://github.com/liqd/euth_wagtail/commit/8e026bffccd920528a38df43afc075e46d3af5d1

Also newest version of Flatpickr should be used for proper Safari Support.

2e2a:
> I fixed some stuff (basically using local times in browser but UTC in BE) for this in our release2.0 branch. I would suggest to use this liqd/euth_wagtail@8e026bf
>
> Also newest version of Flatpickr should be used for proper Safari Support.

Thanks for the hint, will consider it, when we start using it.

adhocracy4/contrib/widgets.py
```diff
@@ -0,0 +1,54 @@
...
+    class Media:
+        js = (staticfiles_storage.url('flatpickr.min.js'),
+              staticfiles_storage.url('js/dateTimeInput.js'))
+        css = {'all': [staticfiles_storage.url('flatpickr.min.css')]}
+
+    input_type = 'text'
+    format_index = 0
+
+    # becomes a public value in Django 1.10
+    def _format_value(self, value):
```
slomo: I am not sure if and why we needed to override this.

2e2a:
Closing this for now. Might reopen it when implementing the date picker.

PR #36, opened 2017-02-17
Send emails one by one to each receiver

slomo:
As I see it the behaviour of sending mails with multiple to entries is never wanted on a a platform. So I think we should change how sending email works.


2e2a:
lgtm

PR #37, opened 2017-02-21
Categories app

2e2a:
Add the following models:
- `Category`: categories that belong to a module.
- `Categorizable`: Add a category field to another model, e.g. Item

### DIFF
--- a/adhocracy4/categories/admin.py
+++ b/adhocracy4/categories/admin.py
@@ -1,0 +1,6 @@
+from django.contrib import admin
+
+from . import models
+
+
+admin.site.register(models.Category)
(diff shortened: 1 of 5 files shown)


MagdaN:
Looks good, I just have one issue which came to my mind just know: How can we translate a category?

2e2a:
> Looks good, I just have one issue which came to my mind just know: How can we translate a category?

It might not need translation, as it is user generated content.

2e2a:
Need to discuss how to enable/configure it.

2e2a:
Not extending UserGeneratedModel anymore, as the creator can be received from `module.project`.

PR #39, opened 2017-03-06
Filtered list view

2e2a:
Defines a view that integrates `django-filter`.
For an usage example see: https://github.com/liqd/a4-meinberlin/pull/86


xi:
Does this deprecate `SortableListView`?

PR #40, opened 2017-03-08
Add archive flag to project

2e2a:


xi:
What is the exact semantics of this? How does it relate to "all phases of all modules have ended"?

2e2a:
> What is the exact semantics of this? How does it relate to "all phases of all modules have ended"?

This relation is not defined in core and can be decided in each project, e.g. by showing this setting only if participation is over.

xi:
I don't think it is a good idea to have flags in core that do not have clear semantics.

From your answer I guess the intended semantics are roughly "Even if a project has completed all of its participation, it may still be active in other ways. This flag allows to distinguish these projects from truly archived ones."

2e2a:
Yes, that's way to define it.

MagdaN:
LGTM please merge, if you are fine with it

PR #41, opened 2017-03-08
Order projects by creation date per default

2e2a:


2e2a:
Found a way to adjust the queryset in the filter. No need for this change.

PR #45, opened 2017-03-13
Remove sort view

2e2a:
Use `FitleredListView` instead.


slomo:
I see that there are some test for sorting with django filters. But I don't understand how sorting is configured (eg. can I limit somehow which fields are available for sorting?).

2e2a:
> I see that there are some test for sorting with django filters. But I don't understand how sorting is configured (eg. can I limit somehow which fields are available for sorting?).

The tests are deleted as part of this pull request. I would suggest to merge it and add tests fir the `FitleredListView` separately.

PR #46, opened 2017-03-14
Module feature detection

2e2a:
Add a helper function to check if any phase of a module provides a certain feature.

Add a templatetag to check if the module of an item has a feature in its phases. Using item instead of module here, as it directly provides the module and the item class.

### DIFF
--- a/adhocracy4/modules/templatetags/module_tags.py
+++ b/adhocracy4/modules/templatetags/module_tags.py
@@ -1,0 +1,8 @@
+from django import template
+
+register = template.Library()
+
+
+@register.assignment_tag
+def itemHasFeature(item, feature):
+    return item.module.has_feature(feature, item.__class__)
--- a/adhocracy4/modules/templatetags/module_tags.py
+++ b/adhocracy4/modules/templatetags/module_tags.py
@@ -3,6 +3,6 @@ from django import template
 register = template.Library()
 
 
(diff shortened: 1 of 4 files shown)


adhocracy4/modules/templatetags/module_tags.py
```diff
@@ -0,0 +1,8 @@
+from django import template
+
+register = template.Library()
+
+
+@register.assignment_tag
+def itemHasFeature(item, feature):
```
slomo: Is there any particular reason why this in camelCase?

slomo: I am wondering whether we want to have this as an assignment tag. Real world usage would probably look like this:

    {% item_has_feature idea 'crud' as idea_has_crud %}
    {% if idea_has_crud %}
    <a href="/edit">Edit idea</a>
    {% endif %}

Alternatives would be a filter:

    {% if idea | has_feature:"crud" %}
    <a href="/edit">Edit idea</a>
    {% endif %}

Or a custom tag (definitely not lazy mans choice):
  
    {% has_feature idea 'crud'  %}
    <a href="/edit">Edit idea</a>
    {% end_has_feature %}

2e2a: pushed an update.

PR #48, opened 2017-03-17
Finalise email porting from OPIN to A4

slomo:
When the initial email code was added to A4 the original classes in OPIN were not deleted. During the last week I changed  (maybe improved) the emails in OPIN. This pull-request proposes the same changes here:

### features

- get rid of the `send_email_with_template` function (was legacy should be replaced with class based emails)
- fix sending of emails to multiple receivers (it was present in A4 but broken)
   - send one email per receiver
-  allow to override the render process for the email
- introduce single template mixin
   - allows to specify the complete email (subject, text, html) in one template
   - enforces more structured approach to emails (still can be changed on per platform basis)
   - requires base templates for [html](https://github.com/liqd/a4-opin/blob/master/euth_wagtail/templates/email_base.html), [text](https://github.com/liqd/a4-opin/blob/master/euth_wagtail/templates/email_base.txt) and [subject](https://github.com/liqd/a4-opin/blob/master/euth_wagtail/templates/email_base.subject)
   - less templates for individual messages (see [example](https://github.com/liqd/a4-opin/blob/master/euth/memberships/templates/emails/invite.en.email) in opin)
- allow projects to configure email global email mixin with `A4_EMAIL_BASE_MIXIN` config variable
- email is now send with user object in context (if possible, doesn't work for invite)
- remove links to `django-admin` from report_moderators email (they might anyway not be able to see it)

### missing features

- change locale used for sending email (e.g get language setting from user profile)
   - easier solution might be to force a language through config right now
- async sending
   - especially important for emails with many receivers (like event stream notifications)
- handle links/italic/bold/underline in text mails


slomo:
@xi @2e2a @vellip In case you also want to switch to the SingleTemplateMixin style email, I would probably remove the config option for email base mixin and while at it also include the app name in the email path as another breaking change.

vellip:
We don't use emails in meinBerlin yet, so you can do that. Single Template for mails sounds great!

2e2a:
> @xi @2e2a @vellip In case you also want to switch to the SingleTemplateMixin style email, I would probably remove the config option for email base mixin and while at it also include the app name in the email path as another breaking change.

SingleTemplateMixin is sufficient for us, you can remove the option for the base mixin.

slomo:
I removed the configurable EmailBaseMixin and therefore also the old reports templates. If needed I could also commit the report templates from opin. Should I also update the that the email templates path contains the app name similar to the "normal" templates?

2e2a:
Works fine for us.

slomo:
Also works for us.

PR #49, opened 2017-03-17
Only call thumbnailer if imagefield is not none

MagdaN:
Fixes https://github.com/liqd/a4-opin/issues/758


slomo:
I think the issue is rather in a4-opin since it doesn't give ImageField but an property as paramter (which is checking the image field for emptiness).

MagdaN:
But it also does not hurt to add this line of code here?

PR #50, opened 2017-03-20
Move generic views and mixins to dedicated package

2e2a:


2e2a:
Choosing another approach.

PR #52, opened 2017-03-28
Document the current phase model

slomo:
In this pull request:

 - document the original modelling/constraints of phases
 - describe what has bee implemented
 - list upcoming challenges

In **future** pull requests:

 - combine plan and current state into a short term vision / documentation of desired model
 - create a long term vision from challenges, further discussion and product vision workshop

### DIFF
--- a/docs/phases.md
+++ b/docs/phases.md
@@ -1,94 +1,102 @@
-# Phases #
+# Phases
 
-## Original design ##
+## Original design
 
-  **Definition:** A phase is fixed time interval inside the linear time line of
-  a particpation project. It enables participants to interact in certain
-  way with the participation projects content.
+### Definition
 
-  **Details:**
-  - each app defines phases (eg ideas/phases.py)
-  - each phase belongs to a module (indirectly to a project)
-  - each module only has phases from one app
-  - each phase has a initiator configruable start and end date
-      - phases are active if start date <= now < end date
-      - never two phases active in the same project
-      - optional as long as project is in draft
-  - if active is phases:
-    - provides a view for (all?) module content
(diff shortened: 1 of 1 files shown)


xi:
I did some gardening. Sounds like a good summary to me. A bit rough on the edges, but AFAIU it will be extended later.

slomo:
It is definitely just a first version. Nevetheless I would be interrested which topics are particular *rough on the edges*.

docs/phases.md
```diff
@@ -0,0 +1,102 @@
...
+-   introduce a new dynamic way for initiators to combine phases
+    -   possibility to create custom blueprints
+    -   two step process
+    -   should support multiple modules, if implemented (see below)
+
+-   unclear concept for multiple modules
+    -   all modules need to be accessible from project detail view
+    -   modules should be integrated in the timeline
+    -   probably multiple active modules at the same time
+    -   very complex ui/ux issue
```
xi: Once the "jump to content of previous phase" issue is fixed, multiple modules yould not be an issue anymore. But "multiple phases (from different modules) at the same time" would still be one.

xi:
> Nevetheless I would be interrested which topics are particular rough on the edges.

Thw whole text is very dense and might not be very helpful for people who are not familiar with the topic. This is especially true for the "Upcomin Challanges" section.

slomo:
I think it is not even a text. It is intentionally just a list of bullet points. I'd prefer to keep it that way in future revisions, because that makes it easier to maintain and a more useful tool for every day work.

2e2a:
lgtm

PR #53, opened 2017-03-28
mark comment author links as external for embed

xi:
see https://github.com/liqd/a4-meinberlin/pull/113

### DIFF
--- a/adhocracy4/comments/static/comments/Comment.jsx
+++ b/adhocracy4/comments/static/comments/Comment.jsx
@@ -122,7 +122,7 @@ var Comment = React.createClass({
         {this.renderDeleteModal()}
         <h3 className={this.props.is_deleted ? 'comment-deleted-author' : 'comment-author'}>
           {this.props.is_deleted ? this.props.user_name
-            : <a href={`/profile/${this.props.user_name}`}>{this.props.user_name}</a>
+            : <a href={`/profile/${this.props.user_name}`} data-embed-target="external">{this.props.user_name}</a>
           }
           {moderatorLabel}
         </h3>


MagdaN:
As embedding is not in core - I think this is not optimal. Another solution would be to add 'data-embed-target="internal"' to all links that are supposed to open inside embed view.

slomo:
I agree with @MagdaN that this is definitely a design smell right now. Especially since it is not clear from the adhcoracy4 code or documentation why this is needed.

As mentioned in the Scrum of Scrums the restrictions mentioned in the [embed document](https://github.com/liqd/a4-meinberlin/blob/master/docs/embed.md#restrictions-on-the-overall-application) also need apply for core (at least if the embed is expected to work with core ui). So we should document those restrictions in core, before merging this.

MagdaN:
Since it is not clear if we need embedding in other projects, I am not sure, if am happy with having those restrictions in core.

2e2a:
I think moving embedding to core should be considered an option for the next future, as there are projects planed that will require it. Then we will also have the relevant documentation here. Not sure if we then need the extra documentation now.

And even if we do not move embedding to core, core still has to implement the implied restrictions, or change the embedding concept, which is not really an option at this point.

MagdaN:
OK, if you all agree, that changing the concept is not an option, then lets merge it

slomo:
At least the result of the scrum of scrums was, that we are very likely to have embedding in multiple projects in the ¿near? future. I think @rittermo or Jana have the better overview concerning this.

MagdaN:
I thought the mein berlin case was supposed to be a test case to check if if this is a suitable solution, now it is a concept that can't be changed.

xi:
This is just a temporary workaround. I created #59 for a long-term solution.

Issue #31, opened 2017-02-07
API permissions do not consider phases


2e2a:
When posting a comment or rate, `object_pk` and `content_type_id`can be set to any project item. The phase permission of the project are not being checked.

slomo:
### problem statement

- currently two permission systems in use (drf permissions & django-rules)
- django-rules has good integration with a4 model
   - checks phases if participant creates new resource
   - checks phase if participant updates own resource
   - prevents information leakage and modification of content in private projects
   - enables all operations for admins, initiator and moderators
- drf permissions has not so good integration
   - enables moderator to update all resources in own project
   - checks that user can only modify own resource
   - does **not** prevent information leakage or modification of content in private projects
     - but there are no APIs to get content from any project (so leakage isn't an issue)
     - modification is (eg. I can comment/rate in any private project if I know a ressource id)
   - does **not** check phases for participant operations

### solution

> combine drf permissions and django-rules

- django-rules needs to know to which module/project a resource belongs
   - because project might be private or in wrong phase
- easy for modifying existing content (get content_object of comment/rate etc)
- hard for creating content
   - only encoded in body, which is parsed after permission checks
- hard for listing content
   - only encoded in optional filter parameters

⇒ **subproblem**: we need to know context 
  - always add module to all api requests (as path part or non-optional parameter)
  - allow each API endpoint to specify its own context (eg `object_pk` and `content_type` for comments/rates)

PR #56, opened 2017-04-03
APIs for generic foreign keys

slomo:
This pull requests is the first of two trying to address #31. As mentioned in the ticket there are not enough informations in the comment API to implement proper rules for creation and listing of comment. The latter is currently solved by disabling listing in all relevant APIs (comment, reports, ratings).

In this pull-request the comment and rating api are prefixed with the content-type and object_pk of their subject.

`/comments/<pk>` ⇒ `/contenttypes/<content_type_pk>/objects/<object_pk>/comments/<pk>`
`/ratings/<pk>` ⇒ `/contenttypes/<content_type_pk>/objects/<object_pk>/ratings/<pk>`

Disadvantages:

- still not solved for reports (might not be required)
- for updating, deleting and detail view the context is not required (but still given)
- long urls
- custom router

There will be a second pull-request for introducing the permissions themself.

### DIFF
--- a/adhocracy4/api/mixins.py
+++ b/adhocracy4/api/mixins.py
@@ -1,0 +1,47 @@
+from django.contrib.contenttypes.models import ContentType
+from django.http import Http404
+from django.shortcuts import get_object_or_404
+
+
+class ContentTypeMixin:
+    """
+    Should be used in combination with ContentTypeRouter to fetch the
+    decode content_type and object_pk of an request.
+
+    Currently only numeric object_pk are supported.
+    """
+    content_type_filter = []
+
+    def dispatch(self, request, *args, **kwargs):
+        content_type = kwargs.get('content_type', '')
+        object_pk = kwargs.get('object_pk', '')
+
+        if not content_type.isdigit() or not object_pk.isdigit():
+            raise Http404
+        else:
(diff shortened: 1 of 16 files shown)


slomo:
Another source of ugliness: Comment reply has another url than comment it self. But replies are still included in the original comment. Example below with content type if comment = 7 and 23 for ideas.

- `/contenttypes/23/obejcts/45/comments` to comment idea 45 => creates comment with id 22
- `/contenttypes/7/objects/22/comments` to reply to the comment above

slomo:
It could also make more sense to use the `app.model` notation instead of the numeric values.

MagdaN:
Its all quite complicated, but I guess it is a suitable solution for the problem.

2e2a:
lgtm

slomo:
Requires to change the way comment and rating APIs are included.

2e2a:
Looks good to me and works fine.

We probably should add `a4_comments.comment` to the `content_type_filter`of comments and ratings directly in the api views.

2e2a:
>  We probably should add a4_comments.comment to the content_type_filterof comments and ratings directly in the api views.

After a discussion we decided to keep `a4_comments.comment` in the settings.

PR #57, opened 2017-04-03
Generic map elements

2e2a:
Copy of generic map elements from a4-opin:
- widgets
- javascript
- setting model


adhocracy4/maps/fields.py
```diff
@@ -0,0 +1,41 @@
...
+from django.utils.translation import ugettext_lazy as _
+from jsonfield.fields import JSONField, JSONFormField
+
+
+class GeoJSONFormField(JSONFormField):
+
+    def to_python(self, value):
+        empty_featureset = '{"type":"FeatureCollection","features":[]}'
+        if (value not in django_validators.EMPTY_VALUES and not
+                value == empty_featureset):
```
xi: Is this check repliable? (e.g. the order of keys could be different)

adhocracy4/maps/fields.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def formfield(self, **kwargs):
+        error_messages = {'required': self.required_message}
+        error_messages.update(kwargs.get('error_messages', {}))
+        kwargs['error_messages'] = error_messages
+        return super(GeoJSONField, self).formfield(**kwargs)
+
+
+class GeometryField(GeoJSONField):
+    pass
```
xi: What is the point of this class?

adhocracy4/maps/mixins.py
```diff
@@ -0,0 +1,49 @@
...
+class MapItemListMixin(object):
+
+    def dump_geojson(self):
+        result = {}
+        result['type'] = 'FeatureCollection'
+        feature_list = []
+
+        for item in super().get_queryset():
+
+            url = ''
```
xi: should be renamed to `image_url`

adhocracy4/maps/mixins.py
```diff
@@ -0,0 +1,49 @@
...
+            point_dict = item.point
+            point_dict['properties'] = properties
+            feature_list.append(point_dict)
+
+        result['features'] = feature_list
+        return result
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['mapideas_json'] = self.dump_geojson()
```
xi: the key is confusing because this is not actually JSON, or is it?

adhocracy4/maps/static/a4maps/map_choose_point.js
```diff
@@ -0,0 +1,101 @@
+function createMap (L, baseurl, name) {
+  var basemap = baseurl + '{z}/{x}/{y}.png'
+  var osmAttrib = '&copy; <a href="http://openstreetmap.org/copyright">OpenStreetMap</a> contributors'
```
xi: As far as I remember, the attribution changes with `baseurl` (e.g.  [mapbox](https://www.mapbox.com/help/attribution/#text-attribution) requires a specific attribution).

adhocracy4/maps/fields.py
slomo: Clarity about what is the default Geometry of GeoJSONField (this code comes from a generic geojson library) analogous to the Point and MultiPolygonField.

adhocracy4/modules/views.py
```diff
@@ -41,6 +41,8 @@ def form_valid(self, form):
     def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
         kwargs['module'] = self.module
+        if self.module.settings_instance:
+            kwargs['settings_instance'] = self.module.settings_instance
```
xi: The name is slightly confusing. Maybe "module_settings_instance" is clearer. I am also not sure what the "_instance" means.

adhocracy4/maps/static/a4maps/map_choose_point.js
```diff
@@ -0,0 +1,101 @@
...
+    var yi = polyPoints[i].lng
+    var xj = polyPoints[j].lat
+    var yj = polyPoints[j].lng
+
+    var intersect = ((yi > y) !== (yj > y)) &&
+        (x < (xj - xi) * (y - yi) / (yj - yi) + xi)
+    if (intersect) inside = !inside
+  }
+  return inside
+}
```
xi: I guess this is taken from [adhocracy3](https://github.com/liqd/adhocracy3/blob/master/src/adhocracy_frontend/adhocracy_frontend/static/js/Packages/Core/Mapping/MappingUtils.ts). Maybe it would be worth to also copy the comment from there.

adhocracy4/maps/templates/a4maps/map_choose_point_widget.html
```diff
@@ -0,0 +1,10 @@
+<script type="text/javascript">
+    var baseurl = "{{ map_url }}";
+    var bbox = {{ bbox }};
+    var name = "{{ name }}";
+    var point = {% if point %} {{ point | safe }} {% else %} undefined {% endif %};
+    var polygon = {{ polygon | safe }};
+</script>
```
xi: I would prefer not to use global variables for this.

adhocracy4/maps/fields.py
```diff
@@ -0,0 +1,41 @@
...
+    def to_python(self, value):
+        empty_featureset = '{"type":"FeatureCollection","features":[]}'
+        if (value not in django_validators.EMPTY_VALUES and not
+                value == empty_featureset):
+            return super().to_python(value)
+        else:
+            return None
+
+
+class GeoJSONField(JSONField):
```
xi: `JSONField` sets `help_text` to "Enter valid JSON" by default. This is not helpful in our case. I am not sure how  we can avoid this tough.

slomo: You can overwrite this in the [form](https://github.com/liqd/a4-opin/blob/master/euth/maps/forms.py) if needed. Which according to the [django docs](https://docs.djangoproject.com/en/1.8/topics/forms/modelforms/#overriding-the-default-fields) can happen even in nicer syntax.

If that is not suitable we can probably also change the error message here.

adhocracy4/maps/static/a4maps/map_choose_point.js
slomo: I think it comes rather from [stack overflow](http://stackoverflow.com/questions/31790344/determine-if-a-point-reside-inside-a-leaflet-polygon). But since this is the same algorithm  and probably even the same original source, lets add the comment.

adhocracy4/maps/mixins.py
slomo: It seems not to be json. But on the other hand it is directly outputted to the template in our [idea listing](https://github.com/liqd/a4-opin/blob/master/euth/maps/templates/euth_maps/mapidea_list.html#L7). Also it is marked safe and I think it is [not safe](https://opin-dev.liqd.net/en/projects/asdasdasd-3/). Security issues coming up!

adhocracy4/maps/mixins.py
```diff
@@ -0,0 +1,49 @@
...
+        for item in super().get_queryset():
+
+            url = ''
+
+            if hasattr(item, 'image') and item.image:
+                image = get_thumbnailer(item.image)['map_thumbnail']
+                url = image.url
+
+            properties = {
+                'name': item.name,
```
slomo: At on `name` here (and maybe also with `slug`, `image` and `url`) there should be a call to escape.

xi:
I started fixing similar issues to #61 in 313d918. Unfortunately, I ran out of time. So someone will have to finish that.

xi:
Luckily I found time to finish my changes after all. Note that my changes are breaking:

- before: Define a global `name` variable and have a `<div>` with id `map_{name}`. Additional parameters are defined as global variables.
- after: have a `<div>` with `data-map="{choose_point|choose_polygon|display_point|display_points}"`. Additional parameters are defined as data attributes.

As a side effect, this also allows to have more than one map on a page.

adhocracy4/maps/templates/a4maps/map_choose_point_widget.html
```diff
@@ -1,10 +1,11 @@
-<script type="text/javascript">
-    var baseurl = "{{ map_url }}";
-    var bbox = {{ bbox }};
-    var name = "{{ name }}";
-    var point = {% if point %} {{ point | safe }} {% else %} undefined {% endif %};
```
xi: I removed checking if point is set here. This still needs to be restored.

adhocracy4/maps/templates/a4maps/map_choose_polygon_widget.html
```diff
@@ -1,9 +1,10 @@
-<script type="text/javascript">
-    var baseurl = "{{ map_url }}";
-    var bbox = {{ bbox }};
-    var name = "{{ name }}";
-    var polygon = {% if polygon %} {{ polygon | safe }} {% else %} undefined {% endif %};
```
xi: I removed checking if polygon is set here. This still needs to be restored.

adhocracy4/maps/widgets.py
```diff
@@ -22,9 +22,9 @@ def render(self, name, value, attrs):
 
         context = {
             'map_url': settings.BASE_MAP,
-            'bbox': json.dumps(settings.MAP_BOUNDING_BOX),
```
slomo: Are you sure that you want to remove this? You are outputting now python maps/lists directly into the template. This will usually be the [string value](https://docs.djangoproject.com/en/1.10/ref/templates/language/#variables), which might be the reason that works for regular maps.

adhocracy4/maps/widgets.py
```diff
@@ -22,9 +22,9 @@ def render(self, name, value, attrs):
...
+            'bbox': settings.MAP_BOUNDING_BOX,
             'name': name,
-            'polygon': json.dumps(value)
```
slomo: See as above.

adhocracy4/maps/widgets.py
slomo: I checked again in out code. And it seems that the setting value is stored as String. I would rather like to change this fact, to have this match the polygon.

slomo: Also most other A4 settings are prefixed with A4. Should we change it for this one?

slomo:
LGTM. There is still a lot of stuff to do. I added a list i n #64, feel free to contribute.

PR #60, opened 2017-04-04
Add rules to REST api

### LINKED ISSUE #31: API permissions do not consider phases

slomo:
Fixes #31 

 - add rules for comments and rates
 - enforce those rules from the api classes

### DIFF
--- a/adhocracy4/api/permissions.py
+++ b/adhocracy4/api/permissions.py
@@ -84,10 +84,10 @@ class RulesPermission(permissions.BasePermission):
         model_cls = self.get_model_cls(request, view)
         perm = self.get_rule(request, model_cls, rules_method_map)
 
-        if not request.user.has_perm(perm, obj):
-            raise PermissionDenied
+        if request.user.has_perm(perm, obj):
+            return True
 
-        return True
+        raise PermissionDenied
 
 
 class ViewSetRulesPermission(RulesPermission):
--- a/adhocracy4/api/permissions.py
+++ b/adhocracy4/api/permissions.py
@@ -38,7 +38,7 @@ class RulesPermission(permissions.BasePermission):
         GET='{app_label}.view_{model_name}',
         OPTIONS='{app_label}.view_{model_name}',
         HEAD='{app_label}.view_{model_name}',
-        POST='{app_label}.create_{model_name}',
+        POST='{app_label}.add_{model_name}',
         PUT='{app_label}.change_{model_name}',
         PATCH='{app_label}.change_{model_name}',
         DELETE='{app_label}.delete_{model_name}',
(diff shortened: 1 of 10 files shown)


slomo:
Blocked because requires #60

slomo:
:warning: Rebased on current master :warning:

adhocracy4/api/permissions.py
```diff
@@ -19,3 +22,85 @@ def has_object_permission(self, request, view, obj):
...
+        else:
+            rules_method_map = self.default_rules_method_map
+
+        model_cls = self.get_model_cls(request, view)
+        perm = self.get_rule(request, model_cls, rules_method_map)
+
+        if not request.user.has_perm(perm, obj):
+            raise PermissionDenied
+
+        return True
```
2e2a: Maybe it is safer to return `PermissionDenied` per default and `True` if `user.has_perm`?

adhocracy4/api/permissions.py
```diff
@@ -19,3 +22,85 @@ def has_object_permission(self, request, view, obj):
...
+    """
+
+    default_rules_method_map = RulesMethodMap(
+        GET='{app_label}.view_{model_name}',
+        OPTIONS='{app_label}.view_{model_name}',
+        HEAD='{app_label}.view_{model_name}',
+        POST='{app_label}.create_{model_name}',
+        PUT='{app_label}.change_{model_name}',
+        PATCH='{app_label}.change_{model_name}',
+        DELETE='{app_label}.delete_{model_name}',
```
2e2a: Maybe use crud terms (read, create, update, delete)?

2e2a:
Nice integration!

adhocracy4/api/permissions.py
slomo: You mean as verbs in those permissions. I took those from [DRF DjangoModelPermission class](https://github.com/encode/django-rest-framework/blob/master/rest_framework/permissions.py#L94), which probably choose those names because the are the [default permission names](https://docs.djangoproject.com/en/1.11/ref/models/options/#default-permissions).

That was more informal than a counter argument. I like the CRUD wording as well, but the other one is well established in the framework. In any case, we should rework our existing permissions to match the naming shema that we choose here :weight_lifting_woman:.

2e2a: > That was more informal than a counter argument. I like the CRUD wording as well, but the other one is well established in the framework. In any case, we should rework our existing permissions to match the naming shema that we choose here :weight_lifting_woman:.

Yes, we should use the verbs we choose here in all permission.

slomo:
@2e2a Fixed comments.

2e2a:
lgtm

PR #61, opened 2017-04-05
avoid using the safe filter

xi:
This fixes some cross-site-scripting vulnarabilities. No changes in the individual projects should be required.

### DIFF
--- a/adhocracy4/comments/static/comments/react_comments.jsx
+++ b/adhocracy4/comments/static/comments/react_comments.jsx
@@ -2,6 +2,8 @@ var CommentBox = require('./CommentBox')
 var React = require('react')
 var ReactDOM = require('react-dom')
 
-module.exports.renderComment = function (mountpoint, props) {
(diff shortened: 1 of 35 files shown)


slomo:
Could you have look how I solved the same issue in https://github.com/liqd/a4-opin/pull/775/commits/39051f22667f4ac01427f9364d987d982cf46b55 and give me some feedback about the solution?

slomo:
Why the splitting of the comment into an inclusion tag? It seems to me that this just unnecessarily distributes logic between to files (it has been an inclusion tag earlier in opin/euth-wagtail, but I rewrote it to be in a single file).

xi:
> Why the splitting of the comment into an inclusion tag?

Mostly for consistency. I also think that separating view and controller is good practice.

xi:
> Could you have look how I solved the same issue in liqd/a4-opin@39051f2 and give me some feedback about the solution?

I think the solution is fine. I thought about implementing that as a second iteration.

The only downside I can see is that it does not work well with embedding yet. It would be nice to know how to fix that first.

PR #63, opened 2017-04-06
Fix packing of reports app

### LINKED ISSUE #62: Packing doesn't include report templates
Currently report emails are broken because, the templates of the app are not included.

slomo:
Fixes #62


2e2a:
lgtm

Issue #64, opened 2017-04-07
Improvements to Maps


slomo:
- [x] make attribution configurable
- [ ] validate GeoJSON (in GeoJSON field)
- ~~rethink if we actually want GeoJSON, or store Points and MultiPolygons in Leaftlet representation~~
  - we want to stick to GeoJSON (decided in Scrum of Scrums)
  - was only an option to simplify validation
- [x] document `isMarkerInsidePolygon` algorithm properly
- [x] replace view mixins for list and detail with template tag to use in detail and list view
- [x] rename display_points attribute from point to points
- [x] don't use map_url and base_url as synonyms
- [x] prefix all settings value with A4 or A4_MAPS
- [ ] set a better default help text for `GeoJSONField`
- [ ] do not hardcode height
   - add as optional parameter to template tag and widget, with old value as default, and none if None
- [ ] consistent usage of form controls in different templatetags/widgets (display_points currently uses a custom solution)
- [ ] Show rates in pins only if the rating feature is enabled

fuzzylogic2000:
partly done

PR #66, opened 2017-04-10
Refactor maps

xi:
This fixes some of the points from #64.

See the individual commit messages for breaking changes.


adhocracy4/maps/mixins.py
```diff
@@ -47,7 +47,7 @@ def dump_geojson(self):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context['mapitems_json'] = self.dump_geojson()
-        context['map_url'] = settings.BASE_MAP
+        context['baseurl'] = settings.MAP_BASEURL
         context['polygon'] = json.dumps(self.module.settings_instance.polygon)
```
MagdaN: Is this the URL to the tile server? If yes, I find it confusing to call this base_url.

xi: It is the name that was used before in most places. I could not think of a better alternative yet.

xi:
I will also implement the "replace view mixins for list and detail with template tag to use in detail and list view" refactoring. This mitigates the risk of naming collisions in the context.

adhocracy4/maps/mixins.py
MagdaN: I think the name should somehow describe, that it is about maps. base_url sounds like host name.

xi: it is always used in the context of maps, so repeating that it is about maps should not be required. That is of course not true for the view mixins. But I am still working on converting them to template tags.

MagdaN: You do what you want, I find it missleading, but you get your way.

tests/maps/test_maps_templatetags.py
```diff
@@ -0,0 +1,253 @@
...
+
+    match = re.match(expected, helpers.render_template(template, context))
+    assert match
+    _point = match.group('point')
+    assert json.loads(unescape(_point)) == point
+    _polygon = match.group('polygon')
+    assert json.loads(unescape(_polygon)) == area_settings.polygon
+
+
+# @pytest.fixture
```
vellip: Shall these stay commented out?

MagdaN:
I am not sure if I like, that the map height is set in CSS, as that way the map does not work out of the box. I would prefer to pass the height to the widget or set a min-height with inline styles.

xi:
> I am not sure if I like, that the map height is set in CSS, as that way the map does not work out of the box.

I reverted my changes for now as I did not find a good way to make it work out of the box and allow simple styling with CSS at the same time. I would prefer merging this PR soon and work on the height separately.

PR #67, opened 2017-04-11
Add maps templates and static content to pypackage

slomo:
Since this is already the second PR with that change and it seems that we will add more and more templates or static content we should address this more broadly.

I tried already to add all template directories and static dir to the package, but failed :sob:. Maybe someone can assist me while I start another attempt?


2e2a:
> I tried already to add all template directories and static dir to the package, but failed :sob:. Maybe someone can assist me while I start another attempt?

I can try to help. This is something we will always forget.

PR #68, opened 2017-04-11
Add an ItemForm which should be used with Item*Views

kleingeist:
Item*Views are initializing forms with the additional `module` and
`settings_instance` keyword arguments. The ItemForm will take those from
the arguments dict and store them as object attributes.


kleingeist:
Need to evaluate form / view centric logic for creation

kleingeist:
This PR will be closed until we decided how to handle form and view interaction.

PR #70, opened 2017-04-11
fix bug with tiles

vellip:
When the map is hidden initially and leaflet initialises tiles don't show properly sometimes. See https://github.com/tombatossals/angular-leaflet-directive/issues/49

Something that also occurs on opin: 
screen shot 2017-04-11 at 12 13 02


vellip:
I just now noticed the lines below!

wrss2:
I have thats same problem in Angular 2 and Material 2 Tabs. Try to reset after initialization not work invalidateSize();.


image

vellip:
@wrss2 We're doing that already, I fixed it in https://github.com/liqd/adhocracy4/pull/71. Thanks for the contribution anyway!

PR #71, opened 2017-04-11
fix bug with initially hidden map

vellip:
The check that was done to do the `map.invalidateSize()` was too broad. When clicking 'result' first and then switching to the middle tab, the `mapVisible` was set already to true. Also: when resizing while not in that particular tab the bug would occur again as well. Now it's just on every click. We can observe if this will have any performance issues but I doubt it.

screen shot 2017-04-11 at 12 13 02


xi:
I don't think this will negatively impact performance. From the leaflet docs:

> Checks if the map container size changed and updates the map if so

So it should not do anything if the size did not change.

wrss2:
i have solution 
    setTimeout(function () {
      map.invalidateSize();
    }, 0);

PR #72, opened 2017-04-12
Add related factories to project and modules

slomo:
Ensure that each factory created project also has a module and a phase.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ help:
 
 install:
 	npm install
-	$(VIRTUAL_ENV)/bin/python3 setup.py development
+	$(VIRTUAL_ENV)/bin/pip install -r requirements/dev.txt
 
 lint:
 	. $(VIRTUAL_ENV)/bin/activate && node_modules/.bin/polylint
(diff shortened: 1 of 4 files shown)


slomo:
@xi @MagdaN  @2e2a Do we want to have this. It requires to fix all those test cases which assume that an project has initially no phases or modules. I started to fix a few, but it requires probably 2 to 3 more hours.

2e2a:
@xi @MagdaN @2e2a Do we want to have this. It requires to fix all those test cases which assume that an project has initially no phases or modules. I started to fix a few, but it requires probably 2 to 3 more hours.

I thinks its nice to have this, as we otherwise have to connect projects, phases and modules manually in many tests.

2e2a:
After some further evaluation with @kleingeist I think that these factories do not help a lot, as the phases are project specific.

Instead we are using helper functions:
https://github.com/liqd/a4-meinberlin/blob/master/tests/helpers.py#L11


I would suggest to close this pull request and not do the extra work of fixing the tests.

slomo:
There is a similar helper/context manager in A4 https://github.com/liqd/adhocracy4/blob/master/tests/helpers.py#L8, which we also used in a copied version in opin and ae. Maybe we should promote it to `adhcoracy4.test.helpers`, otherwise I agree.

PR #73, opened 2017-04-12
do not feature archived projects

mkind:
Do not feature archived projects anymore. This might become obsolet if we want to enable the user whether to filter or not.

### DIFF
--- a/adhocracy4/projects/models.py
+++ b/adhocracy4/projects/models.py
@@ -17,7 +17,7 @@ class ProjectManager(models.Manager):
         return self.get(name=name)
 
     def featured(self):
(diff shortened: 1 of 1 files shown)


slomo:
I think that those properties/querysets are only relevant for a4-opin (an only used in one place each). So maybe we should remove them from adhocracy4?

xi:
We chose to manually pick the featured projects via wagtail. So yes, you can remove it. But I also do not see any harm in leaving it in, e.g. as an example.

MagdaN:
yes, it can be removed

2e2a:
lgtm

PR #74, opened 2017-04-12
draft: polygon presets

xi:
This is a first draft for polygon presets. There are several possible UI concepts for this:

1. Show a select by default. One option is "custom". If this option is selected, the map widget is shown.
2. Show a select and a button. When the button is clicked, the select is disabled and the map widget is shown.
3. Both map and select are shown. When a user selects a preset in the select, it is loaded into the map (replacing the previous content).

We expect that in the case of meinberlin, about 70% of projects will use an unaltered polygon preset. So it should be a prominent option to use a preset while also making it easy to define a custom polygon.

This pull requests implements the third option. It seems to be the most usable (e.g. it allows to edit a preset after selection) and is also quite easy to implement.

Still to do:

-   get presets from the backend
-   only include the select if there are any presets
-   ask the user for confirmation before replacing the content (required?)
-   add a label/help text to the select

### DIFF
--- a/adhocracy4/maps/static/a4maps/map_choose_polygon.js
+++ b/adhocracy4/maps/static/a4maps/map_choose_polygon.js
@@ -99,5 +99,19 @@ window.jQuery(document).ready(function () {
         mapVisible = true
       }
     })
+
+    $('#select_' + name).on('change', function (event) {
+      var geoJson = event.target.value
+      drawnItems.clearLayers()
+      if (geoJson) {
+        var group = L.geoJson(JSON.parse(geoJson), {
+          style: polygonStyle
+        })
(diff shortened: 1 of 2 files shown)


vellip:
After talking with @MagdaN and @slomo about it we decided that this feature should be contained in the meinberlin repo for now. While there's interest in this feature the actual benefit of it for the OPIN project is not clear yet nor is the `<select/>` solution viable enough (There'd be a hell lot of `<option/>`s in the select) to make it in the core. But that's all we got at least in this iteration.

We'd extend the MapChoosePolygonWidget and do the JavaScript there. We just might have to expose the Leaflet `map` object for an API to access the map and draw on it. Maybe we can do that via jQuery.data.

PR #79, opened 2017-04-24
Consistent permission naming

2e2a:
In https://github.com/liqd/adhocracy4/pull/60 we decided to use the permission terms view, add, change, delete.

Renaming the respective predicates for consistency.

### DIFF
--- a/adhocracy4/modules/predicates.py
+++ b/adhocracy4/modules/predicates.py
@@ -46,13 +46,13 @@ def is_allowed_view_item(user, item):
             is_public_context(user, item))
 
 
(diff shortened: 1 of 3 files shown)


slomo:
Looks greate we need to change 3 lines in OPIN to make this happen. What do you think about adding a permission called `create_project` and replace all usages of `inititate_project`? And if that is useful (for sake of consistency) should we do it right now in the PR?

2e2a:
Makes sense, I'll add it.

slomo:
I included this in https://github.com/liqd/a4-opin/pull/809, works nicely.

PR #81, opened 2017-04-25
Allow default filter values

2e2a:
Copy from https://github.com/liqd/a4-opin/commit/c9a12dc9851f35ecec39a7950eeb1552c81a7fe9

### DIFF
--- a/tests/filter/test_filter_views.py
+++ b/tests/filter/test_filter_views.py
@@ -1,0 +1,70 @@
+import pytest
+
+from tests.apps.questions import models as question_models
+
+from adhocracy4.filters.filters import DefaultsFilterSet
+from adhocracy4.modules.views import ItemListView
+
+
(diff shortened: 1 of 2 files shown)


tests/filter/test_filter_views.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@pytest.mark.django_db
+def test_default_filter(rf, question_list_view, phase, question_factory):
+    project = phase.module.project
+    request = rf.get('/questions')
+    q1 = question_factory()
+    q1.text = 'x'
+    q1.save()
+    q2 = question_factory()
```
slomo: you can override fields directly with the factory (eg. `q2 = question_factory(text='x')`) w/o using save afterwards.

2e2a: Good to know. Will fix it.

PR #82, opened 2017-04-25
refactor url replace logic and update documents api url

vellip:
You'll need to update the call to documents endpoints. see: https://github.com/liqd/a4-meinberlin/pull/254/commits/04c6cc29d243c4a258034828305c4ec7718b9ab8 

Also adds a generic url replace logic.

Assuming you want to post to an endpoint's url `api/contenttypes/$contentTypeId/objects/$objectPk/comments/` you now make a call to `api.comments.change` with an object `urlReplaces` in data. JS will look for the keys `contentTypeId` and `objectPk` in that object and replace the placeholders in the url with the particular value in `urlReplaces`.

### DIFF
--- a/adhocracy4/ratings/static/ratings/react_ratings.jsx
+++ b/adhocracy4/ratings/static/ratings/react_ratings.jsx
@@ -8,8 +8,10 @@ var classnames = require('classnames')
 var RatingBox = React.createClass({
   handleRatingCreate: function (number) {
     api.rating.add({
-      object_pk: this.props.objectId,
-      content_type: this.props.contentType,
+      urlReplaces: {
+        objectPk: this.props.objectId,
+        contentTypeId: this.props.contentType,
+      },
       value: number
     }).done(function (data) {
       this.setState({
@@ -23,8 +25,10 @@ var RatingBox = React.createClass({
   },
   handleRatingModify: function (number, id) {
     api.rating.change({
-      object_pk: this.props.objectId,
-      content_type: this.props.contentType,
+      urlReplaces: {
+        objectPk: this.props.objectId,
+        contentTypeId: this.props.contentType,
+      },
       value: number
     }, id)
       .done(function (data) {
--- a/adhocracy4/static/api.js
+++ b/adhocracy4/static/api.js
@@ -12,11 +12,8 @@ var baseURL = '/api/'
 var api = (function () {
   var urls = {
     report: baseURL + 'reports/',
-    document: baseURL + 'documents/',
-    follow: baseURL + 'follows/'
-  }
-
-  var genericUrls = {
+    document: baseURL + '/modules/$moduleId/documents/',
+    follow: baseURL + 'follows/',
     comment: baseURL + 'contenttypes/$contentTypeId/objects/$objectPk/comments/',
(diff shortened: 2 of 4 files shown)


adhocracy4/static/api.js
```diff
@@ -31,16 +28,13 @@ var api = (function () {
...
-    } else {
-      url = genericUrls[endpoint]
-        .replace('$contentTypeId', data['content_type'])
-        .replace('$objectPk', data['object_pk'])
+    var url = urls[endpoint]
+    if (data.urlReplaces) {
+      url = url.replace(/\$(\w+?)\b/g, (match, group) => {
+        return data.urlReplaces[group]
+      })
       data = $.extend({}, data)
```
vellip: why was this needed?

slomo: Because the caller passes you an object and we start deleting values from it. And it could be that the caller wants to do something else with the object afterwards. Than she would probably wonder what happend to those keys.

adhocracy4/ratings/static/ratings/react_ratings.jsx
```diff
@@ -23,8 +25,10 @@ var RatingBox = React.createClass({
   },
   handleRatingModify: function (number, id) {
     api.rating.change({
-      object_pk: this.props.objectId,
-      content_type: this.props.contentType,
+      urlReplaces: {
```
slomo: Why do the components suddenly require knowledge about the url. I thougt the `api.rating` and `api.comments` where meant to abstract that away.

vellip: They required knowledge already before. Previously they had to be specified in `data` too and were deleted afterwards. Now they still have to be in data but the values are grouped together.

adhocracy4/static/api.js
```diff
@@ -12,11 +12,8 @@ var baseURL = '/api/'
 var api = (function () {
   var urls = {
     report: baseURL + 'reports/',
-    document: baseURL + 'documents/',
-    follow: baseURL + 'follows/'
-  }
-
-  var genericUrls = {
+    document: baseURL + '/modules/$moduleId/documents/',
```
xi: Why is here a leading slash?

adhocracy4/ratings/static/ratings/react_ratings.jsx
xi: I think both options have value. @vellip's approach is more explicit, the old approach had nice separation of abstraction layers. For me, both is fine.

slomo: Okay I could agree that before they were passed as data while beeing used in the url. The client has already an additional parameter for `id` which also goes into the url. So why not stop passing any url parts as data then, and instead pass them as additional url paramter.

```
api.rating.change(
   {'id': id, 'object_pk': object_pk, 'content_type': content_type},
   {'value': value}
)
```

While at it you could probably also specify the url more directly. Also to reduce the complexity and functionality in the api client.

```
api.change(
   ['contentTypes', content_type, 'objects', object_pk, 'ratings', id],
   {'value': value}
)
```

PR #83, opened 2017-04-27
Actions app

2e2a:
Mostly a copy from `a4-opin` with small refactorings.


adhocracy4/actions/management/commands/timed_system_actions.py
```diff
@@ -0,0 +1,31 @@
+from django.core.management.base import BaseCommand
```
slomo: Should it be named `create_system_actions`?

adhocracy4/actions/signals.py
```diff
@@ -0,0 +1,53 @@
...
+def _has_project(instance):
+    return hasattr(instance, 'project') \
+           and instance.project.__class__ is Project
+
+
+def _extract_target(instance):
+    target = None
+    if hasattr(instance, 'content_object'):
+        target = instance.content_object
+    elif _has_project(instance):
```
slomo: Following many other places (like our rules for instance) we could check here only for `hasattr(instance, 'project')`

adhocracy4/actions/signals.py
```diff
@@ -0,0 +1,53 @@
...
+        target=target,
+    )
+
+    if _has_project(instance):
+        action.project = instance.project
+
+    action.save()
+
+
+if hasattr(settings,'A4_ACTIONABLES'):
```
slomo: I am not in favour of having this optional. If you don't want to use actions simply don't enable the app.

adhocracy4/actions/verbs.py
```diff
@@ -0,0 +1,12 @@
+# http://activitystrea.ms/registry/verbs/
+
+CREATE = 'created'
```
slomo: Left hand values are in pastense to keep it compatible to an older version of OPIN actions.

adhocracy4/actions/verbs.py
```diff
@@ -0,0 +1,12 @@
+# http://activitystrea.ms/registry/verbs/
```
slomo: This whole modules could be replaced by an python enum. Which is simply didn't knew when I implemented this.

    Verbs(enum.Enum):
      CREATE = 'create'
      ADD = 'add'
      UPDATE = 'updated'
      COMPLETE = 'complete'

I am not sure if that really is an improvement, since the `Enum` is rather new (introduced in Python 3.4). There is no proper Django Support (like an Enum field).

adhocracy4/follows/templatetags/react_follows.py
```diff
@@ -0,0 +1,12 @@
+from django import template
+from django.utils.html import format_html
+
+register = template.Library()
+
+
+@register.simple_tag()
+def react_follows(project):
+    return format_html(
+        '<span data-euth-widget="follows" data-project={project}></span>',
```
slomo: This widget won't work embedded, because it is not triggered by inline scripts. Either we have to fix the global javascript hook for embedding issue or it needs to changed to something along those lines:

    return format_html(
           '<span id="{mountpoint}" data-project="{project_slug}"></span>'
           '<script>window.adhocracy4.renderFollow(document.getElementById("{mountpoint}"))</script>',
           project_slug=project.slug,
           mountpoint="follow-button-" + project.slug,
    )

Also you need to hookup this in [index.js](https://github.com/liqd/adhocracy4/blob/master/index.js).

2e2a:
Applied comments

slomo:
Should I merge it directly or integrate it first on opin, to see if there are any further integration issues?

2e2a:
> Should I merge it directly or integrate it first on opin, to see if there are any further integration issues?

We need to implement the notifications next week and would need it to be merged then. If you have time until then, it would be good to check if the integration works.

slomo:
I'll try to do it tomorrow, otherwise I will merge.

slomo:
Found not time last week. Will merge now instead.

PR #87, opened 2017-05-03
add project follows

xi:
This ports the follows app from a4-opin to adhocracy4 core. There are some important things to note:

-   It is currently only possible to follow projects. In the future it might be a requirement to follow other things, e.g. items. It is not yet clear how this will work exactly. It will probably be implemented as a separate (but related) feature. We think that using generic foreign keys is overkill here.

-   The react component does not currently support the case where the user is not logged in. For now, the recommended bahvior in this case is not to show the component. See https://github.com/liqd/a4-opin/issues/629 and https://github.com/liqd/a4-opin/issues/627 for related issues in a4-opin.

-   There is some discussion as to whether the API of this feature should be changed. This pull request keeps the current API. The related issue https://github.com/liqd/a4-opin/issues/510 should be moved to this repository.

-   The API does not currently check project membership on private projects. The recommended way for this is to check membership when sending notifications. It might be a good idea to check it here though.

-   The HTML code may need to be adapted in order to be stylable in different projects.

### DIFF
--- a/adhocracy4/api/mixins.py
+++ b/adhocracy4/api/mixins.py
@@ -1,6 +1,9 @@
 from django.contrib.contenttypes.models import ContentType
 from django.http import Http404
 from django.shortcuts import get_object_or_404
+from rest_framework import status
+from rest_framework.request import clone_request
+from rest_framework.response import Response
 
 from adhocracy4.modules import models as module_models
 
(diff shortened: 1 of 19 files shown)


2e2a:
Added tests from a4-opin.

2e2a:
We should consider renaming the `Follows` model to `FollowsProject`.

2e2a:
https://github.com/liqd/adhocracy4/issues/91

PR #93, opened 2017-05-09
return whether a project can be archived

mkind:
* return whether a project is archivable
* a project is archivable if not yet archived and if already finished


2e2a:
lgtm

PR #96, opened 2017-05-16
Enable signals for inheriting projects.

kleingeist:
Currently the signals are registered explicitely for the Project base model. In
meinBerlin we inherit from the base model to enable external projects and
bplans. As Django signals do not take inheritance into account when handling
signals, it is required to register signals without binding to a sender and
check if the senders type is a subclass of Project manually.

If, for example for performance reasons, a general signal receiver is not wanted in the core, we could add explicit signal handlers for two inherited project types we had to implement.

### DIFF
--- a/adhocracy4/projects/signals.py
+++ b/adhocracy4/projects/signals.py
@@ -6,18 +6,21 @@ from adhocracy4.images import services
 from .models import Project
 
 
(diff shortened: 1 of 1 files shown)


2e2a:
@MagdaN @slomo: what do you think?

MagdaN:
LGTM

PR #97, opened 2017-05-17
Remove the category related field name.

kleingeist:
If concrete categorizable Items in different apps have the same Modelname, the related field reference on the category clashes. (For example `budgeting.models.Proposal` and `kiezkasse.models.Proposal` would bost result in the same `Category.proposal_set`)
As there is no real usecase to get the related items of a specific type from a Category object i propose to remove the backwards field reference


2e2a:
makes sense.

PR #98, opened 2017-05-17
Add a paging aware filterset.

kleingeist:
By default FilterSets from django_filter are keeping every non filter
argument in the query string. This results a problem when paging is used
in addition to filters (see liqd/a4-meinberlin#403)
It is desired to reset the page to zero when applying a filter. This PR
introduces a PagedFilterSet which removes the page argument from the
query string.

### DIFF
--- a/adhocracy4/filters/filters.py
+++ b/adhocracy4/filters/filters.py
@@ -2,7 +2,22 @@ import django_filters
 from django.http import QueryDict
 
 
-class DefaultsFilterSet(django_filters.FilterSet):
+class PagedFilterSet(django_filters.FilterSet):
+    """Removes page parameters from the query when applying filters."""
+    page_kwarg = 'page'
+
+    def __init__(self, data, *args, **kwargs):
+        if self.page_kwarg in data:
+            try:
+                data.pop(self.page_kwarg)
+            except AttributeError:
+                # Create a mutable copy
+                data = data.copy()
+                data.pop(self.page_kwarg)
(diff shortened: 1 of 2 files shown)


adhocracy4/filters/filters.py
```diff
@@ -2,7 +2,22 @@
...
+    page_kwarg = 'page'
+
+    def __init__(self, data, *args, **kwargs):
+        if self.page_kwarg in data:
+            try:
+                data.pop(self.page_kwarg)
+            except AttributeError:
+                # Create a mutable copy
+                data = data.copy()
+                data.pop(self.page_kwarg)
```
MagdaN: Why is that needed?

kleingeist: Django is using a custom `QueryDict` which may be immutable. Immutable QueryDicts are throwing AttributeErrors if it is tried to modify them.

slomo: I think the query dict will always be immutable by default (so I would only implement that case) and there is also `QueryDict.mutable`, which is probably more than catching an attribute error.

kleingeist: @slomo what do you mean by "there is also QueryDict.mutable"
i did consider the mutable case as default because i assumed the `PagedFilterSet` will be called mainly as the parent class of `DefaultsFilterSet` which is explicitly using a mutable QueryDict
unfortunately the QueryDict is using the private `_mutable` attribute for its state and has no public method to get it's state. i did consider _underscore attributes as not suitable for use

PR #101, opened 2017-05-29
add some translatable verbose_names

xi:
I had to switch to `ugettext_lazy` in one place. I am not completely sure about the consequences though.


MagdaN:
This should not break anything, I guess, what do you think @slomo?

slomo:
Makes perfectly sense. We should watch out for more places like those, they tend to show up from time to time. In OPIN the translations are so incomplete that we rarely catch them.

PR #103, opened 2017-05-30
Fix DefaultsFilterSet

kleingeist:
QueryDicts are inheriting from MultiValueDicts which act different then dicts
on updates, as:
  "update() extends rather than replaces existing key lists."

This commit will set the default filter value only if it has not been manually
set yet.

Fixes liqd/a4-meinberlin#490


2e2a:
lgtm

PR #106, opened 2017-06-08
Add overwriteable get_languages method to emails

kleingeist:
To allow for translated emails the overwriteable `get_languages` method
is introduced. It gets the receiver as an attribute an may be
overwritten to send Emails according to (not yet available) user
settings.
Currently the currently set language is used, which is set from the
browsers language settings if the call results from a user request or
from `settings.LANGUAGE_CODE`.

### DIFF
--- a/adhocracy4/emails/__init__.py
+++ b/adhocracy4/emails/__init__.py
@@ -3,7 +3,7 @@ from django.contrib.sites import models as site_models
 from django.core.mail.message import EmailMultiAlternatives
 from django.template import Context
 from django.template.loader import select_template
-from django.utils.translation import get_language
+from django.utils import translation
 
 from . import mixins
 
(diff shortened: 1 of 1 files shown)


2e2a:
@MagdaN, @slomo: could you take a look

slomo:
LGTM. I didn't test it OPIN but I guess it should work as expected without further changes.

MagdaN:
Hallo Alexej,kannst bitte noch einmal diese NRW-Karte updaten (einfach Pullen), die unter Downloads liegt und dann Moritz Bescheid geben. Es scheint dringend zu sein.Liebe Grüße und vielen DankMagdaAm 19.06.2017 8:46 vorm. schrieb Alexej Starschenko <notifications@github.com>:@MagdaN, @slomo: could you take a look

PR #107, opened 2017-06-14
update leaflet to 1.0.x

xi:
https://github.com/Leaflet/Leaflet/blob/master/CHANGELOG.md

This updates leaflet to 1.0.3. All changes I made should be backwards compatible.

I testes all four map types and found no further issues. leaflet.draw also seemed to work fine with the new version.

I encountered on strange issue: Yesterday, when ever I was on a page that had leaflet loaded, I was unable to scroll the page. However, today I could not reprodce the issue anymore.

See also https://github.com/liqd/a4-meinberlin/pull/544 for further changes that might be required for individual platforms.

### DIFF
--- a/adhocracy4/maps/static/a4maps/map_choose_point.js
+++ b/adhocracy4/maps/static/a4maps/map_choose_point.js
@@ -35,19 +35,36 @@ function createMarker ($, L, newlatln, oldlatln, basePolygon, map, name) {
   return marker
 }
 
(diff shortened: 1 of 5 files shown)


slomo:
I am currently lacking the time to try this implement this properly in OPIN. Depending on how urgent this is for you I would either merge it without updating OPIN or try to do it this or next week.

PR #108, opened 2017-06-14
Add START verb and signal start of projects

kleingeist:
The start of a project ist defined with an Action with the verb START
and the project as the object.
A START action will be created if a project first phase starts within
the next hour.


2e2a:
lgtm

PR #109, opened 2017-06-15
Delete Actions if the related object is deleted

kleingeist:
As we are loosing all the information when the object of an action is
deleted, the action itself may be deleted, too.

### DIFF
--- a/adhocracy4/actions/signals.py
+++ b/adhocracy4/actions/signals.py
@@ -7,7 +7,7 @@ from django.db.models.signals import post_delete, post_save
 from .models import Action
 from .verbs import Verbs
 
-
+# Actions resulting from the create_system_actions management call
 SYSTEM_ACTIONABLES = (
     ('a4phases', 'Phase'),
     ('a4projects', 'Project')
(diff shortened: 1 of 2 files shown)


adhocracy4/actions/signals.py
```diff
@@ -1,11 +1,19 @@
...
 from django.conf import settings
-from django.db.models.signals import post_save
+from django.contrib.contenttypes.models import ContentType
+from django.db.models.signals import post_delete, post_save
 
 from .models import Action
 from .verbs import Verbs
 
 
+SYSTEM_ACTIONABLES = (
```
kleingeist: if it is not created due to a setting in A4_ACTIONABLES but due to a call of the management function 'create_system_actions'

kleingeist:
> Here we introduce another signal and before doing that I would like to think about, whether we can't come up with better solutions. One that I can come up with is "simply" filtering the action with deleted objects or subjects from the view.

Unfortunately filtering in SQL is not possible due to generic foreign keys. The tablenames used for joins have to be static and cannot be resolved from other db fields

PR #112, opened 2017-06-21
Set the START project actions timestamp to the actual phases start_date

kleingeist:
The START action should include the actual objects start timestamp. If a project start is rescheduled the original START action has to be adapted.

This is different from the SCHEDULE action, where the timestamp is in relation to the action itself.
In opin the SCHEDULE action is not used to send the "ends in 24 hours/soon" email, but a COMPLETE action which is not in a4 core yet.
While i agree to use  the phase.end_date as the timestamp for the COMPLETE action, i'd argue to keep the time when the action was created as the timestamp for the SCHEDULE action.

@xi @slomo  what is your opinion on

The START action and the modifications to SCHEDULE were initially introduces in #108

### DIFF
--- a/adhocracy4/actions/management/commands/create_system_actions.py
+++ b/adhocracy4/actions/management/commands/create_system_actions.py
@@ -54,13 +54,17 @@ class Command(BaseCommand):
                     obj_object_id=project.id
                 ).first()
 
-                # If the first phases start has been modified and moved ahead
-                # of the previous action, a new project start action is created
-                if not existing_action \
-                        or existing_action.timestamp < phase.start_date:
-
+                if not existing_action:
                     Action.objects.create(
                         project=project,
                         verb=Verbs.START.value,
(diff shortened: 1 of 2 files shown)


xi:
For me the basic question is this: Does an `Action` object represent the action itself, or is it *about* an action. In the first case, the `Action` object should have the timestamp when it actually happened. In the latter case, it should be the time when it was created (which can be before or after the actual action).

slomo:
I think we discussed this briefly offline, but I can't remember if we had any further insights.

2e2a:
To me It makes more sense, if the timestamp would mean the time of the action object (not the action). However this introduces a certain non-linearity in the event stream, due to our cron-job based creation of some events.

I'm fine with both solutions:
- Keep it as is. Schedule having the timestamp of the action creation, keeping the actions linear
- Use the object time in the action and sort the actions differently

PR #113, opened 2017-06-21
Async mail sending

2e2a:
Usage:
- add `background_task` to installed apps
- `manage.py makemigrations background_task`
- `manage.py migrate`

Run:
python manage.py process_tasks

TODO:
- [x] make sure email logic still works in async
- [x] tests


kleingeist:
In the scrumofscrums it was decided to send mails asynchronous by default, except for certain account related emails like registration, confirmation, password reset.

kleingeist:
to ensure account emails are send synchronously see the changes made for meinBerlin:
https://github.com/liqd/a4-meinberlin/pull/598/commits/249aa72902725d245cce8b2bb1d8f4af715caccf

adhocracy4/test/helpers.py
```diff
@@ -33,3 +37,29 @@ def redirect_target(response):
 def render_template(string, context=None):
     context = Context(context or {})
     return Template(string).render(context)
+
+
+@contextmanager
+def skip_background_mail():
```
slomo: Could you add a comment that briefly summarizes what the aim of this context manager is? I am not sure if I understood it correctly, but if so I would probably write "Re import email tasks module with background_task decorator mocked away. The new implementation calls the function synchronously and checks all arguments for json serilizability".

You probably already checked it, but there is no means provideded by background_tasks to write such tests?

In addition I was also thinking whether we shouldn't make this more generic, so that one can basically give the module as a parameter here.

kleingeist:
> I think this looks good to merge. I would just have one questions towards API design. The whole email classes are implemented in this class based email pattern. The async feature now doesn't fit exactly into that pattern. Have you thought about having an async mixin that basically overrides send or introducing async as a class property that decides wether async sending is used or not.
>
> To put this in more context. Currently all email options are decided if you define your email class and not be the caller who actually sends your email. Now this is a feature where the caller has the burden to select the right option.

i agree using a Mixin / SyncEmail class would make sense in the context of the current implementation. Note that SOS decided to make Async the default case, so we would use the SyncEmail as Base for account emails

adhocracy4/test/helpers.py
kleingeist: the problem is the validation if the input is json serializable.
before i did implement a skip_decorator(decorator, module) context which replaced the given decorator used in the given module with an identity decorator. But after a review feedback it was decided to check that the input is serializable. although we were not 100% sure about it, as we define in the doc, that the input has to be serializable and we use internal knowledge of the thirdparty module by checking the serializibility with json.
what is your opinion on that? should we skip the serialization test and rely on the definition of valid input for the test?

slomo: So it was generic for all decorators not only for the background-task decorator. My intention was to only make it generic for the background-task decorator. So it seems to me there are those options worth to discuss:

1. Move the entire thing as it is from helpers to the email test package, because it is only useful there
2. Make a generic replace decorator Y in module X with function Z. Where X = backgroundtask Y = adhcoracy4.emails and Z= a function that checks json serilisabilityof arguments, in our current use case.

Anyhow these are only mere cosmetic changes. The docstring is more important for me.

PS: Arbeiten in der Freizeit ist der Feind der Arbeiterschaft :factory:

kleingeist: > 1. Move the entire thing as it is from helpers to the email test package, because it is only useful there

i disagree as the skip_background_task context is required to test if emails are/would sent in other modules. see for example the `test_report_api.py` [1]
because i expect this to be a common pattern i think it is nice to have a context that is named explicitly and does not require further knowledge about the the EMail implementation. 

Another option would be to globally skip background tasks for all tests and just enable them if they are explicitly needed. Although i am not entirely sure this is doable.

[1] https://github.com/liqd/adhocracy4/pull/113/files#diff-460ed8a9f44094f159330c5b76202e04R56

slomo: > i disagree as the skip_background_task context is required to test if emails are/would sent in other modules. see for example the test_report_api.py

Oh that was not obvious to me and I missed that point.

>  because i expect this to be a common pattern i think it is nice to have a context that is named explicitly [...]

Okay so it is basically a `force_sync_email_sending` context?

Towards the option of disabling that in test. Couldn't I instead test whether the email job was enqueued properly? I am a bit in favour of solutions that do not require another context manager, because a lot of our test already use a context manager (either `freeze_time` or `with_phase`) and I am not a huge fan of nesting them to much.

slomo:
So as I see it this is incomplete. Right @kleingeist ?

kleingeist:
@slomo i did refactor the code to use a SyncEmailMixin rather than a separate send_sync method.
furthermore i did disable all background tasks in the root conftest.py. Note that this has to be done for every project (ae, mb, opin), too

Making the Email class pluggable was not possible, as the concrete Email implementations would have to inherit from a class that is only available via the settings object. It would  have been possible to do something like Djangos Meta definitions but that seemed overkill for our usecase.

slomo:
only for informational purposes: In https://github.com/liqd/adhocracy4/commit/e8bc62bc37e1bd4bb5b71fa1b084ecbb29217d67 I removed the previously existing mechanism to dynamically configure the base email template. But I can't access whether that would still work with the current state of the projects.

slomo:
What is the status here?

PR #114, opened 2017-06-21
some first thoughts about testing

mkind:
I appreciate comments.

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -1,0 +1,98 @@
+Testing Strategies
+==================
+
+General Concepts
+------------------
+
+* security vs. reliability vs. performance tests
+* black-box vs. white-box
+
+Further Reading
+* [CMU: Software Testing](https://users.ece.cmu.edu/~koopman/des_s99/sw_testing/)
+* [OWASP: Security Testing Web Applications throughout Automated Software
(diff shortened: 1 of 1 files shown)


xi:
Nice writeup so far. However, I don't think this is fit to be included in this repo.

So far this is a summary of generic best practices for testing. I would prefer to link to some external resources for this. Instead, this document should concentrate on the specific aspects of testing in the context of adhocray4.

docs/testing.md
```diff
@@ -0,0 +1,98 @@
...
+
+Coverage
+--------
+
+The code coverage measures how many lines of code are executed during
+testing. It's a good idea to maximize coverage. It's not a good idea
+to rely on the coverage. The coverage does not give any advice about
+logical issues or cases, that are not considered in your code. However,
+there should be a good reason for a low coverage. Having no time is
+not a good reason.
```
xi: I don't agree with the last sentence. Having no time even may be the only reason for low coverage. Mostly anything can be done with enough time, including refactoring the whole application in order to improve testability.

As an example: Our experience with browser-based integration tests in adhocracy3 was mostly that they were extremely unreliable. We could have spend a lot of time and effort to improve that, but other things were more impotant.

slomo:
What about this guy? It is the last open PR on the repo currently.

MagdaN:
Lets merge this

mkind:
go for it

PR #115, opened 2017-06-28
Refactor ProjectMixin

kleingeist:
- ProjectMixin not flexible enough, we had to "simulate" ProjectMixins behavior in a lot of places
- Goal to simplify and find a simple invariant: Every view in the context of a project should have set the project as an view attribute (and on the request.project)
- The active module and phase can be derived from the project. This holds as long as only a single phase _and_ module is active
  If we have to allow for multiple active modules/phases we would have to replace the ProjectContextDispatcher with an equivalent ModuleContextDispatcher
- The ProjectContextDispatcher allows for simpler unit tests as it is easy to mock the project context 

Note: Currently this PR is breaking as ProjectMixin has been removed. But it could be kept and marked deprecated to allow for backwards compatibility 

Note: We dropped the Mixin suffix for classes that overwrite the dispatch method, as they are dependent on the order of the dispatch method execution and thus on the order of the declaration of parent classes.


kleingeist:
closed in favor for #123 and liqd/a4-meinberlin#583

PR #116, opened 2017-06-28
Check if project is live in module predicates

2e2a:
Before the module predicates did allow to change, rate, et al. items  of unpublished projects if the phase did allow it.


slomo:
LGTM

PR #118, opened 2017-06-29
add tile image to project

2e2a:
The aspects of the tile image and the heroimage is too different. Cropping the heroimage to the tile does not work in general.

### DIFF
--- a/adhocracy4/projects/migrations/0008_project_tile_image.py
+++ b/adhocracy4/projects/migrations/0008_project_tile_image.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import adhocracy4.images.fields
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('a4projects', '0007_add_verbose_names'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='project',
+            name='tile_image',
+            field=adhocracy4.images.fields.ConfiguredImageField('tileimage', verbose_name='Tile image', blank=True, upload_to='projects/tiles', help_prefix='The image will be shown in the project tile.'),
+        ),
+    ]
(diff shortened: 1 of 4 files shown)


kleingeist:
just a note: old images (due to modifications/deletions) are currently not automatically deleted. for projectes a signal is used to delete them (see [0])
some time ago i did rewrite the delete images signals to act on every image field found (see [1]) but those changes are not merges yet.
@slomo did vote against signals and would prefer a cronjob which deletes old images regularly. 


[0] https://github.com/liqd/adhocracy4/blob/master/adhocracy4/projects/signals.py
[1] https://github.com/liqd/adhocracy4/compare/2017-06-jd-move-images-signals

slomo:
Since this just adds additional fields, I would merge this. I have just two additional points here: 

 1. Has this been taken though some design process for adhocracy4 in general. Or asked differently should we adopt other platforms here? ( @damusp )
 2.  Even though I am on a personal vendetta with signals, we should probably delete those files somehow and keep it consistent. So I would say we should right now also add/extend signals for the tile image field. (Was that also your point? @kleingeist )

2e2a:
> Has this been taken though some design process for adhocracy4 in general. Or asked differently should we adopt other platforms here? ( @damusp )

No, this was not a design decision, but a reaction to customer complains.
But we can discuss this for other projects, as the problem is there too.

2e2a:
> Even though I am on a personal vendetta with signals, we should probably delete those files somehow and keep it consistent. So I would say we should right now also add/extend signals for the tile image field. (Was that also your point? @kleingeist )

Will add the delete signals.

kleingeist:
> Even though I am on a personal vendetta with signals, we should probably delete those files somehow and keep it consistent. So I would say we should right now also add/extend signals for the tile image field. (Was that also your point? @kleingeist )

@slomo yes, i linked to my branch which registers signals for all ConfiguredImageFields on django initialization.
@2e2a, @slomo what is your opinion on the general approach? i'd apparently would prefer it over the explicit registration of delete handlers for every ConfiguredImaged.

PR #123, opened 2017-07-05
Add last_active_module property to projects

kleingeist:
This simply adds a new last_active_module property which returns the active phases module, or if non is active the last active phases module

### DIFF
--- a/adhocracy4/projects/models.py
+++ b/adhocracy4/projects/models.py
@@ -145,6 +145,10 @@ class Project(base.TimeStampedModel):
     def is_private(self):
         return not self.is_public
 
+    @property
+    def modules(self):
+        return self.module_set
+
     @functional.cached_property
     def active_phase(self):
         return self.phases\
--- a/adhocracy4/projects/models.py
+++ b/adhocracy4/projects/models.py
@@ -145,10 +145,6 @@ class Project(base.TimeStampedModel):
     def is_private(self):
         return not self.is_public
 
-    @property
-    def modules(self):
-        return self.module_set
-
     @functional.cached_property
     def active_module(self):
         """Return the module of the currently active or last past phase."""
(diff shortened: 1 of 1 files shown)


adhocracy4/projects/models.py
```diff
@@ -145,6 +145,18 @@ def other_projects(self):
     def is_private(self):
         return not self.is_public
 
+    @property
+    def modules(self):
+        return self.module_set
```
kleingeist: not sure if it would be better to return self.module_set.all() ?

adhocracy4/projects/models.py
```diff
@@ -145,6 +145,18 @@ def other_projects(self):
...
+
+    @functional.cached_property
+    def active_module(self):
```
slomo: Should we name that differently, because of the difference between an active module and an active phase in the proposed implementation? Like `last_active_module'?

adhocracy4/projects/models.py
slomo: If it is meant for external usage I would definitely return a queryset and not a related manager. In a addition one could also ask if we need the proposed abstraction right now.

kleingeist:
@slomo i did rename and remove the modules property
the modules property was added to have an equivalent to the phases property and define some kind of expected interface. but from a functional perspective it is not needed

slomo:
I will merge this now. I think the modules property is not flawed by definition. It can help to decouple more logic from Djang, as long as we only use it as iterable. So if you think it is useful I see no harm in introducing it, while in my opinion it should definitely be a queryset and not a related manager.

kleingeist:
@slomo i just realized  that is was a related manager when is was looking at my own pull request :(
as i was only using it in one place i don't mind adding it later if we need it

PR #124, opened 2017-07-07
free text filter

fuzzylogic2000:


fuzzylogic2000:
This filter has to be configured with the model fields to search on. It uses Q-objects and the lookup expression icontains to filter for the search value. Does not use DB specific features or external search providers.

kleingeist:
lgtm
i think it is nice to be able to search by related fields, too, due to the usage of Q fields (`user__name`)

PR #126, opened 2017-07-07
Ka 2017 06 move filter widgets to a4

fuzzylogic2000:
The DropdownLinkWidget existed in the same version in meinBerlin und OPIN (and now also AE). To use the widget from A4, just change the import (from 'from apps.contrib import widgets' to 'from adhocracy4.filters import widgets') and move the templates to templates/adhocracy4/widgets/ in some general folder.

The TextInputWidget is needed to use the FreeTextFilter.


fuzzylogic2000:
I couldn't find the tests (in OPIN or meinBerlin) to move/copy to A4. Could someone point me there? @slomo Or give me some hints how to test these? @mkind

slomo:
I guess no one before test this on a modular scale, so basically you have to come up with your own tests. Testing the filters seems already to be solved so I assume there is only the widgets left to test. Similar test could be the test for the [image upload widget](https://github.com/liqd/adhocracy4/blob/master/tests/images/test_image_widgets.py) or the (rate template tag)[https://github.com/liqd/adhocracy4/blob/master/tests/ratings/test_rating_templatetags.py]

xi:
I am fine with including `DropdownLinkWidget` in core. I would even prefer to add some HTML, if only for documentation on how to correctly use it (e.g. ARIA attributes).

I do not yet understand the purpose of `TextInputWidget`. Could you explain what it does and why it is needed?

Since the only job of widgets is to render HTML, I guess the only sensible way to test them is to call `render()` and check the output. (See for example [this django test](https://github.com/django/django/blob/master/tests/admin_widgets/tests.py#L253))

fuzzylogic2000:
The TextInputWidget puts out a TextInput-field, which has all the information needed for the filter. It gets its label and value in the same way as the DropdownLinkWidget and it also gets the url parameters of the other filters, so that these are not lost. I couldn't use any ot the existing widgets of the django filters or django, because there either wasn't a suitable one or it doesn't give me the filter-functionality, I need.

If that widget has to prove itself first, we could also just add it later (or not at all).

Thanks for pointing me to the tests!

adhocracy4/filters/templates/a4filters/widgets/text_input.html
```diff
@@ -0,0 +1,15 @@
...
+    </label>
+    <br>
+    <form class="search-filter" action="" method="get" id="{{ id }}">
+        {% for par_name, value in url_par.items %}
+            {% if par_name != name %}
+                <input type="hidden" name="{{ par_name }}" value="{{ value }}">
+            {% endif %}
+        {% endfor %}
+        <input class="search-filter-input" id="{{ id }}" type="text" name="{{ name }}" value="{{ value_label }}">
+        <button class="search-filter-btn btn" type="submit"><i class="fa fa-search" aria-hidden="true"></i></button>
```
xi: The button needs some text. I would change `aria-hidden="true"` to `aria-label="{{ trans 'Search' }}"` and maybe also add `title="{{ trans 'Search' }}"`

adhocracy4/filters/widgets.py
```diff
@@ -0,0 +1,96 @@
...
+        if not hasattr(self, 'data'):
+            self.data = {}
+        if value is None:
+            value = ''
+
+        _id = attrs.pop('id')
+
+        return render_to_string(self.template, {
+            'id': _id,
+            'value_label': value,
```
xi: Why is this called `value_label`? Why not `value`?

adhocracy4/filters/widgets.py
```diff
@@ -0,0 +1,96 @@
...
+class TextInputWidget(TextInput):
+    """Extend to get a text input into your filter.
+
+    Put 'adhocracy4.filters.apps.FiltersConfig' into your
+    settings.
+
+    To use your own template, overwrite the given one.
+    """
+    label = None
+    right = False
```
xi: I noticed that you use this to change the layout. I think we should avoid that and handle layout separately. Note that the `right` attribute on `DropdownLinkWidget` still makes sende because you may need it to decided where to draw the dropdown box.

adhocracy4/filters/widgets.py
```diff
@@ -0,0 +1,96 @@
...
+        })
+
+
+class TextInputWidget(TextInput):
+    """Extend to get a text input into your filter.
+
+    Put 'adhocracy4.filters.apps.FiltersConfig' into your
+    settings.
+
+    To use your own template, overwrite the given one.
```
xi: It took me some time to understand the purpose of this widget, so maybe some more explanation could be useful. Maybe something like this:

"This widget renders a complete <form> element. It automatically creates hidden fields for all other filters so they are preserved on submit."

I would also rename this widget to `FreetextFilterWidget`. We can still rename it if we ever find a second use case.

adhocracy4/filters/widgets.py
fuzzylogic2000: Ha! I thought, it was supposed to be used like that, because the sorting seemed to be right-aligned in all designs. Will have to change that then.

adhocracy4/filters/widgets.py
fuzzylogic2000: Thanks! Lots of good points!

adhocracy4/filters/widgets.py
xi: If this is better for you we can leave it in for an easy transition. But should note it down for refactoring.

adhocracy4/filters/templates/a4filters/widgets/text_input.html
xi: (the `title` should be on the `<button>`)

adhocracy4/filters/widgets.py
fuzzylogic2000: I removed it from the widget, but still use it in the html. You can change that in the html, and we can overwrite that template for now and change it later.

tests/filter/test_filter_widgets.py
```diff
@@ -0,0 +1,31 @@
...
+    assert ('<li><a href="?test_filter=test-val1">test-label1</a></li>\n'
+            '<li><a href="?test_filter=test-val2">test-label2</a></li>'
+            in html)
+
+
+def test_render_text_input_widget():
+    widget = widgets.TextInputWidget()
+    html = widget.render('test_filter', 'test_val1', attrs={'id': 'test_id'})
+
+    assert ('<label for="test_id" class="filter-label">') in html
```
kleingeist: @xi i think it would be nice to have tests that are covering the retention of selected filters. those would be hidden fields for the TextInputWidget and Links for the Dropdown.
i would also match for `<input class="search-filter-input" id="{{ id }}" type="text" name="{{ name }}" value="{{ value_label }}">` instead of or additionally to the label

xi:
Nice!

PR #127, opened 2017-07-10
overwrite polygon default help text

xi:
the default help text is "enter valid JSON". It can not be blank (see https://github.com/dmkoch/django-jsonfield/issues/95)

see also https://github.com/liqd/a4-meinberlin/issues/503


kleingeist:
@xi changing the help_text / verbose_name will probably trigger a new migration which is missing from the pull request

PR #128, opened 2017-07-11
Allow to override kwargs in FilteredListViews

slomo:
- passes the queryset from the list view to the filter instead of the filter using `$model.objects.all()`
    - allows to prefilter the queryset with the `queryset` option of the view.
 - allows to pass additional arguments to the FilterSet initializer from views extending the FilteredListViews
    - use case was to filter ideas on a profile based on which users profile I am currently viewing

I am pretty confident that the first feature should not make any issues, but still marking this breaking.

### DIFF
--- a/adhocracy4/filters/filters.py
+++ b/adhocracy4/filters/filters.py
@@ -25,8 +25,8 @@ class DefaultsFilterSet(PagedFilterSet):
 
     defaults = None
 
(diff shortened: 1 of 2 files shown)


kleingeist:
lgtm :+1: 
the changes are non breaking for meinBerlin. i would merge if you give the okay for opin and ae

PR #130, opened 2017-07-14
Extend actions properties and queryset

slomo:
-  extend queryset with `exlude_update` and `filter_public` (used in mB and AE)
- make type and icon properties of Action (removes need for Proxy model)
   -  allow each project to configure its icons and types
- change default ordering to descending by creation time

Example usage: https://github.com/liqd/a4-advocate-europe/pull/187


2e2a:
lgtm

PR #131, opened 2017-07-17
Add a property to access a modules first phases start date

kleingeist:


2e2a:
LGTM. Merging this for now.

In general, we need to decide what kind of properties we want to have.

PR #135, opened 2017-07-20
Remove item views

2e2a:
These item views are not generic enough to be in core.
Meinberlin uses abstract idea views instead.


kleingeist:
lgtm.

PR #141, opened 2017-07-25
Ensure django 1.11 compatibility

slomo:
- split `adhcoracy4.test.factories` into multiple package => requires changes to tests
- allow any Django version >=1.8
- remove explicit usage of template.Context, will most certainly require some changes to emails
    eg.  https://github.com/liqd/a4-meinberlin/blob/master/meinberlin/apps/cms/emails.py#L59


slomo:
The last point might not apply to meinBerlin, because with a quick search I could only find uses of the dict api of the Context object.

setup.py
```diff
@@ -13,7 +13,7 @@
       include_package_data=True,
       install_requires = [
           'bleach',
-          'Django >=1.8, <1.9',
+          'Django >=1.8',
```
kleingeist: in requirements/base.txt django is still pinned to Django==1.8.17

kleingeist:
i did a first try on django 1.11 with meinberlin and product. unfortunately we have to do some fixes for meinberlin before we can upgrade.

furthermore i did fix some more errors on a4. as i didn't know if it is okay to push to @slomo 's PR i created a new branch which can be compared to this one at the following url and merged to this PR if you are okay with it
https://github.com/liqd/adhocracy4/compare/ym-2017-07-ensure-django-1.11-compatibility...2017-07-jd-ensure-django-1.11-compatibility

kleingeist:
addendum: i did not test if the changes made a4 in this (and my) PR would break backwards compatibility. maybe we could ensure django 1.11 compatibility in core while keeping 1.8 in meinberlin until the next sprint

slomo:
@kleingeist It is a bit hard to see which commits on that compare actually concern Django 1.11 compat. I could only figure out f6a9839acc951be00dbbe4512ce128de62852c80,  4b356d10ce91376aab6d6b5e9afd96f94d79dee2 and probably the version updates. So I would propose you add the minimal working set to this PR (I guess in general you are allowed to ammend peoples PR). If you want to you can also add the other commits restructering deps and so, but I don't see them strictly to be concerned with Django 1.11 compat.

By static analysis :smile: I would say your changes should also work in Django 1.8. Most definitely I would keep compat for django 1.8 and 1.11 (maybe we should make that clear in setup.py). At least until all projects (opin. product and mB) have switched.

kleingeist:
@slomo most of the commits seen in the diff cam from the master which was merged in my django 1.11 branch. i did merge the to your branch too. furthermore i did rebase and squash the requirement files updates:

the relevant commits for compatibility are regarding flatatt and get_fields which are both compatible with 1.11 and 1.8.

i am in favor of merging

PR #142, opened 2017-07-25
Change markers for a better overview

Navigator28:
- deleted shadows
- riseOnHover for a better overview

It would be even better for the overview if there would be the possibility to change the picture while focus an marker. But this is a not necessary or very important feature.


Navigator28:
Reason for closing: Shadows are a feature which is probably necessary for other projects.

PR #144, opened 2017-07-25
Support for multiple modules

### LINKED ISSUE #148: Unifiy past/future phases logic
Currently there is the `PhasesQuerySet` which defines `active_phases` and `finished_phases`, the `Project` model with `future_phases` and `past_phases` and the `Module` model with `future_phases` and `past_phases` properties that are doing very similiar things
I think we should unify the properties to use the `PhasesQuerySet` and extend the QuerySet with default filters and a future phases method

2e2a:
Support multiple modules:
- Define active module in project
- Define ModuleDetailView
- Copy properties from `Project` to `Module` (backwards compatibility is kept)
- Adapt rule predicates to modules
- Extended documentation with basic modules description (docs should be restructured later)

Fixes #148


adhocracy4/modules/views.py
```diff
@@ -0,0 +1,45 @@
...
+class ModuleDetailView(rules_mixins.PermissionRequiredMixin,
+                       generic.DetailView):
+    model = models.Module
+    permission_required = 'a4projects.view_project'
+    slug_url_kwarg = 'module_slug'
+
+    def dispatch(self, request, *args, **kwargs):
+        module = self.get_object()
+        if module.is_active:
+            return ModuleRedirectView.as_view()(request, *args, **kwargs)
```
xi: Note: This only works if there is only one active module at any time. This could be changed by checking "is this the module that is currently displayed on the project page" instead.

2e2a:
>  Note: This only works if there is only one active module at any time. This could be changed by checking "is this the module that is currently displayed on the project page" instead.
 
Defined the properties`active_module`(Project) and `is_active_in_project` (Module) to resolve this. Only one module is considered active in the project context and is beeing redirected to project detail.

adhocracy4/modules/models.py
```diff
@@ -25,6 +31,42 @@ def settings_instance(self):
...
+    def is_active_in_project(self):
+        return self == self.project.active_module
+
+    @property
+    def active_phase(self):
+        return self.phase_set \
+            .active_phases() \
+            .first()
+
+    @functional.cached_property
```
kleingeist: not sure if that makes sense, as `all()` returns a `QuerySet` which will be lazy evalauated on the db when it is iterated

```
type(Module.objects.first().phase_set.all())
Out[16]: adhocracy4.phases.models.PhasesQuerySet
```

adhocracy4/modules/models.py
```diff
@@ -25,6 +31,42 @@ def settings_instance(self):
...
+    def phases(self):
+        return self.phase_set.all()
+
+    @property
+    def future_phases(self):
+        phases = self.phase_set.filter(models.Q(start_date__gt=timezone.now())
+                                       | models.Q(start_date=None))
```
kleingeist: This is copied from Project but i am not sure why a phase without a start date should be a future phase. Does anyone has an idea?

kleingeist:
I am not sure when we should use cached properties. does anyone has a general advice on how to decide.

Somehow related i realized we could have a weird state if some request would be handled just when a phase ends, as the active phase etc. are dependent on timezone.now() which changes over the course of a request. i know this is a very rare case and i am not sure if we should consider working on it

adhocracy4/modules/views.py
```diff
@@ -0,0 +1,45 @@
...
+class ModuleDetailView(rules_mixins.PermissionRequiredMixin,
+                       generic.DetailView):
+    model = models.Module
+    permission_required = 'a4projects.view_project'
+    slug_url_kwarg = 'module_slug'
+
+    def dispatch(self, request, *args, **kwargs):
+        module = self.get_object()
+        if module.is_active_in_project:
+            return ModuleRedirectView.as_view()(request, *args, **kwargs)
```
kleingeist: why did you use the RedirectView? I guess it's a leftover from a previous version. but we could just return directly via
```
from django.http.response import HttpResponseRedirect
...
    if module.is_active_in_project:
        url = reverse('project-detail', kwargs={'slug': module.project.slug})
        return HttpResponseRedirect(url)
```

adhocracy4/modules/models.py
xi: Maybe the semantics of a phase without start_date is "planned, but not yet scheduled". How does `start_date__gt` handle `None`?

kleingeist:
some more notes:

- `Project.active_phase` and `Project.active_module` are following different semantic, but i think they should act the same
    - `active_phase` returns the active phase with the smallest weight
    - `active_module` returns the active module which ends first
- `adhocracy4/phases/predicates.py: has_feature_active` will break with multiple active phases. we would have to test if the phase is in the list of active_phases (which we don't have yet as a property)
- `create_system_actions` is not ready to handle projects with multiple phases (that maybe could even end all within 24hours)

adhocracy4/modules/models.py
slomo: Semantically that was exactly the idea, that non-dated phases are future phases because they are expected to be scheduled.

In SQL comparisons with `Null` are supposed to be `unknown`, which is neither true nor false. So they would not match the select criteria.

2e2a:
Pushed some updates reflecting the comments.

adhocracy4/modules/models.py
```diff
@@ -25,6 +32,32 @@ def settings_instance(self):
                 return getattr(self, setting)
 
     @property
+    def is_active(self):
+        return self == self.project.last_active_module
```
xi: I don't like the wording here. For me, a module is "active" when it has an active phase. This also how we use the term in the meinBerlin frontend. I have some ideas for alternatives:

- In the frontend, we also have the concept of a "selected" module (i.e. the module that the user is currently looking at). So maybe `is_selected_by_default`. (I also think that this property is only relevant for UI)

- Maybe call it by what it is: `is_last_active`.

- Maybe just leave it out completely. `module == project.last_active_module` is not that much to type and perfectly understandable..

PR #145, opened 2017-07-26
Add reply_to

Navigator28:
Adds reply_to so we (as a receiver of the email) are able to send an emails back.


adhocracy4/emails/__init__.py
```diff
@@ -15,6 +15,9 @@ class EmailBase:
     fallback_language = 'en'
     for_moderator = False
 
+    def get_reply_to(self):
+        return []
```
kleingeist: i'd prefer to return `None` by default, as this is the current default value for EmailMultiAlternatives

https://github.com/django/django/blob/1.8.18/django/core/mail/message.py#L407

kleingeist:
lgtm

PR #147, opened 2017-07-27
Fix background task decorator patching

kleingeist:
Since models.py and signals (from app.ready()) are loaded before the
conftest.py is evaluated it is required to reload the module from which
the patch decorator is used.
This PR extends the patch_background_task_decorator to allow for a list
of module names where the task decorator is used and which have to be
reloaded to apply the patch.

P.S. Until now the patching did work because core emails are only loaded
from the DRF api.py which is loaded after the conftest.py is evaluated.
But in meinBerlin we realised that the emails send from models/signals
were still handled by background_task

### DIFF
--- a/adhocracy4/test/helpers.py
+++ b/adhocracy4/test/helpers.py
@@ -1,3 +1,4 @@
+import importlib
 import json
 import os
 from unittest import mock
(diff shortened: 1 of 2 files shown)


2e2a:
lgtm

PR #151, opened 2017-08-08
add guidelines

vellip:


docs/guidelines.md
```diff
@@ -0,0 +1,21 @@
...
+## What features will be added to A4
+ - the feature has to be planned in more than one project
+ - don't add a feature to core which relies on code not apparent in core
+ - has been discussed in Scrum of Scrums
+ - feature should be sufficiently stable
+
+## Who reviews pull requests
+ - if the change is breaking (check & label your PR):
+   - Review has to be done from another team than the one providing.
+ - if the change adds a new feature:
```
slomo: Should we also not here that we want to label PR with `feature` in that case?

vellip: Makes sense, done!

PR #153, opened 2017-08-09
Support for multiple modules

### LINKED ISSUE #148: Unifiy past/future phases logic

kleingeist:
Support multiple modules:
- Define last active phase in project as the the last started past or currently active phase
- Define last active module in project as the module of the last active phase
- Define ModuleDetailView and ModuleDispatcher
- Copy properties from `Project` to `Module` 
- Adapt rule predicates to modules
- Extended documentation with basic modules description (docs should be restructured later)

Fixes #148

This PR replaces #144 as we refactored again with @xi and wanted to keep the commits clean
see #144 for the previous discussion


kleingeist:
This PR is as far as i can tell **non breaking**

* `Project.past_phases` order had been changed, but it does not occur as  in our projects:
```
./multi_git_grep.sh past_phases
a4-advocate-europe
==========


a4-meinberlin
==========
meinberlin/apps/contrib/predicates.py:    return project.active_phase or len(project.past_phases) > 0
meinberlin/apps/projects/templates/meinberlin_projects/includes/timeline.html:                        {% if not project.past_phases %}


a4-opin
==========


a4-product
==========

```

* `Project.active_phase` and `Project.days_left` has been changed and marked deprecated, but their behaviour stays the same in projects without overlapping phases

* `PhaseDispatchMixin` has been updated to pass the module in addition to the project. But as the counterpart `ProjectMixin` has been adapted, too this is a non-breaking change as long as it is guaranteed, that every PhaseContent.view inherits from `ProjectMixin`

* rules have  been changed, but so have the predicates using them
```
./multi_git_grep.sh has_feature_active
adhocracy4
==========
adhocracy4/comments/rules.py:    return phase_predicates.has_feature_active(
adhocracy4/phases/predicates.py:def has_feature_active(module, model, feature):
adhocracy4/phases/predicates.py:    return has_feature_active(item.module, item.__class__, 'crud')
adhocracy4/phases/predicates.py:        return has_feature_active(module, item_class, 'crud')
adhocracy4/phases/predicates.py:    return has_feature_active(item.module, item.__class__, 'comment')
adhocracy4/phases/predicates.py:    return has_feature_active(item.module, item.__class__, 'rate')
adhocracy4/ratings/rules.py:    return phase_predicates.has_feature_active(


a4-advocate-europe
==========


a4-meinberlin
==========


a4-opin
==========


a4-product
==========


```

adhocracy4/phases/models.py
```diff
@@ -20,6 +20,29 @@ def active_phases(self):
     def finished_phases(self):
         return self.filter(end_date__lte=timezone.now())
 
+    def past_phases(self):
```
slomo: I think this queryset API has now at least two problems: One they use different wording for features that sound very similar. One is <temporal_adjective>_phases  and the other <adverb>_phases (eg. past_phases vs finished_phases). And the second issue, those API endpoints partly duplicate code but also differ slightly in semantics (especially ordering). 

I think we should clear this up. Not saying we should do it right now, but it needs to be tracked somewhere.

While at it, we could also think whether we choose a smart default ordering for phases. But I guess we have to look back again why we choose it.

adhocracy4/projects/mixins.py
```diff
@@ -1,29 +1,59 @@
...
+
+
 class ProjectMixin(generic.base.ContextMixin):
     def dispatch(self, *args, **kwargs):
         self.project = kwargs['project']
-        self.phase = self.project.active_phase or self.project.past_phases[0]
-        self.module = self.phase.module if self.phase else None
+        self.module = kwargs['module']
+        self.phase = self.module.last_active_phase if self.module else None
+        # Workaround for filters
```
slomo: In what sense is that a workarround? Do we expect to remove it and if so how would we do it?

adhocracy4/projects/models.py
```diff
@@ -135,36 +135,71 @@ def has_member(self, user):
...
-                   .active_phases()\
-                   .first()
+        """
+        Return the currently active phase.
+
+        The currently active phase is defined as the phase that out of all
+        currently active phases started last. This is analogous to the last
+        active phase.
+
+        Attention: This method is _deprecated_ as multiple phases may be
```
slomo: Should specify or own deprecation decorator or use the one from django here, to make this more clear?

adhocracy4/projects/models.py
```diff
@@ -135,36 +135,71 @@ def has_member(self, user):
...
+        active at the same time.
```
slomo: The updated document in the docs state, that there shall be only one active module, hence only one active phase.

docs/phases.md
```diff
@@ -1,21 +1,23 @@
-# Phases
```
slomo: I guess the file now could also be called phases_and_modules.md

docs/phases.md
```diff
@@ -60,43 +58,41 @@ Below are listed the differences of what is currently implemented in a4 and opin
...
-        -   extend project to store release date of results
-        -   integration into timeline
-
 -   jump to content of previous phase
     -   needed because offline phases (OPIN only) hide the module content
     -   transition generic offline phases to events
         -   events still visible in timeline
     -   only module specific offline phases
 
--   hide features if not usable in a module
```
slomo: Is that generically solved?

adhocracy4/modules/migrations/0003_rm_unique_name.py
```diff
@@ -0,0 +1,25 @@
...
+    operations = [
+        migrations.AlterField(
+            model_name='module',
+            name='description',
+            field=models.CharField(blank=True, help_text='This short description will appear on the header of the module and project detail pages. It should briefly state the goal of the module in max. 250 chars.', max_length=250, null=True, verbose_name='Short description of the module'),
+        ),
+        migrations.AlterField(
+            model_name='module',
+            name='name',
+            field=models.CharField(help_text='This title will appear in the timeline and the header on the module and project detail pages. It should be max. 512 characters long', max_length=512, verbose_name='Title of the module'),
```
kleingeist: good point. i was thinking from the meinberlin/product perspective. but i am not sure if the (new) timeline should be in core. i guess it's to specific for our different projects. i think we could remove the helptext.

adhocracy4/phases/models.py
kleingeist: good point. we somehow missed the duplication between finished and past phases. btw. finished_phases is not in any project, just in the a4 tests. i think i'd be save to remove it.

we had many discussions about the ordering. 
@xi wished for a consistent api regarding ordering (by setting Meta.ordering = 'start_date')
while i tried to keep backwards compatibility. Changing the default ordering would require to adapt the dashboard code to explicit order on the software defined weight.
i think it would be a good idea but wanted to hear back form you what you think about breaking

adhocracy4/projects/mixins.py
kleingeist: i was confused why we need to set it to the request in addition to the view, but @2e2a explained it was due to the limitness of django filters where only the request is available and no other arguments may be passed:
https://github.com/liqd/a4-meinberlin/blob/d075131d8f3d2b6f85073ddd81cf514a20e1b1a9/meinberlin/apps/contrib/filters.py#L12

maybe documenting it as "workaround" is a bad choice. i justed wanted a reminder on why it was set. should i try to further document it

adhocracy4/phases/models.py
xi: I am not sure, but maybe we missed something: Do we have a problem once start/end date become optional due to the dashboard overhaul?

adhocracy4/projects/models.py
kleingeist: unfortunately django has no deprecated decorator :( 
we could use warning.deprecated() but they won't be shown by default except for djangos RemovedInNextVersion (afaik). And they wouldn't be more apparent than a comment when using the method.

we could add a custom decorator like: https://stackoverflow.com/a/30253848 but it seemed overkill for the two methods. I'd rather like to keep the code free of deprecated methods, but i'm neither against including the decorator. 

@slomo should i add the decorator?

docs/phases.md
kleingeist: i don't know. you'd have to check back with @2e2a

adhocracy4/projects/models.py
xi: fixed in 59c6ff2

docs/phases.md
xi: I don't think this is solved. See for example rating in maps #64.

adhocracy4/comments/models.py
```diff
@@ -60,18 +60,20 @@ def save(self, *args, **kwargs):
...
         else:
-            return None
+            return self.module.get_absolute_url()
 
     @property
     def notification_content(self):
         return self.comment
 
     @property
     def project(self):
```
kleingeist: i am not sure. there are so many places where we have kind of mixed objects in the same context where both project and module are required that i think it's okay to have the shortcut (module.project) on many places

adhocracy4/modules/migrations/0003_rm_unique_name.py
slomo: Let me check back first if the timeline might become a core feature.

adhocracy4/projects/models.py
slomo: My comment below is questioning, whether this stuff is even deprecated. If it is, we might just delete those, which makes this breaking but would be fine with me.

adhocracy4/projects/mixins.py
slomo: Or should we make it a valid place for transfering this kind of data, and just add project, phase and module to the request?

adhocracy4/phases/models.py
slomo: Afaik dates are already optional. So either we have already a problem or not.

adhocracy4/comments/models.py
slomo: Mhm ... I am not really convinced. At least we might provide a default implementation in a useful place.

adhocracy4/projects/models.py
slomo: When writing this comment I was thinking about the following feedback:

> This changes our notion from one module can only be active to many modules can be active. But this restriction was actually a deliberate choice in order to simplify the modelling of real world processes/projects and because we don't have any concepts how to handle this ui and ux wise.

So in short my expectation was to fix this the other way round.

PR #154, opened 2017-08-10
remove moment

vellip:
This allows us to dump moment.js. 

This is not breaking, but I still suggest to remove lines located in webpack.config (e.g. https://github.com/liqd/a4-opin/blob/ae114fc8c5bd010dec9b0736f5dd316cd5932bb7/webpack.config.js#L37) or package.json.

### DIFF
--- a/adhocracy4/comments/static/comments/Comment.jsx
+++ b/adhocracy4/comments/static/comments/Comment.jsx
@@ -8,10 +8,13 @@ var CommentManageDropdown = require('./CommentManageDropdown')
 
 var React = require('react')
 var django = require('django')
-var moment = require('moment')
 
 var safeHtml = function (text) {
   return { __html: text }
+}
+
+var localeDate = function (dateStr) {
+  return new Date(dateStr).toLocaleString(document.documentElement.lang)
 }
 
 var Comment = React.createClass({
@@ -102,9 +105,9 @@ var Comment = React.createClass({
     let CommentList = require('./CommentList')
     let lastDate
     if (this.props.modified === null && !this.props.is_deleted) {
-      lastDate = moment(this.props.created).format('lll')
+      lastDate = localeDate(this.props.created)
     } else if (!this.props.is_deleted) {
-      lastDate = django.gettext('Latest edit on') + ' ' + moment(this.props.modified).format('lll')
+      lastDate = django.gettext('Latest edit on') + ' ' + localeDate(this.props.modified)
     }
 
     let moderatorLabel
--- a/package.json
+++ b/package.json
@@ -16,7 +16,6 @@
     "classnames": "^2.2.5",
     "jquery": "2.2.4",
     "js-cookie": "~2.1.2",
-    "moment": "^2.14.1",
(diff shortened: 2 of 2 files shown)


adhocracy4/comments/static/comments/Comment.jsx
```diff
@@ -8,12 +8,15 @@ var CommentManageDropdown = require('./CommentManageDropdown')
...
 var React = require('react')
 var django = require('django')
-var moment = require('moment')
 
 var safeHtml = function (text) {
   return { __html: text }
 }
 
+var localeDate = function (dateStr) {
+  return new Date(dateStr).toLocaleString(document.documentElement.lang)
```
xi: According to spec, the `Date` constructor only supports a very strict date format. Browsers tend to be more lenient, but this can be a source for incompatibilities.

slomo: This values comes from our api, which by [default](http://www.django-rest-framework.org/api-guide/fields/#datefield) uses is8601. According to [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse) this is fine since ecma script 5.

xi: I think MDN is wrong here:

> ECMAScript defines a string interchange format for date-times **based upon a simplification of** the ISO 8601 Extended Format. The format is as follows: YYYY-MM-DDTHH:mm:ss.sssZ
> -- https://www.ecma-international.org/ecma-262/5.1/#sec-15.9.1.15

Our format uses a space instead of `T`. This is valid ISO 8601, but is not supported by ECMAScript.

slomo: Okay in this case we should adapt our API settings to be on the safe side here.

slomo: Basically there are three options here:

  1. Enforce the ECMA script date format on the DateTimeFields of the API
  2. Require all projects to set their `REST_FRAMEWORK = { ' DATETIME_FORMAT': '???' }`
  3. Continue using the current iso format and replace ` ` with `T` before parsing.

slomo:
I marked this PR incomplete because the issue brought up by @xi in https://github.com/liqd/adhocracy4/pull/154#discussion_r132893090.

adhocracy4/comments/static/comments/Comment.jsx
xi: As for differences between ECMA and ISO time formats, also see https://stackoverflow.com/a/45716310

vellip: Actually our format does not use a space. Outputting `this.props.created` gives me `2017-08-24T13:20:14.642234Z`.

slomo: I checked again and the indeed the [DRF documentation](http://www.django-rest-framework.org/api-guide/fields/#datetimefield-format-strings) already states that correct format is used. So we are fine here.

PR #155, opened 2017-08-10
Add ModuleContent/Type registry

kleingeist:
As we decided in the a4workshop

> - we want Moduletypes in Code, maybe they are not required in Database
> - we want one module type per app (to register export views and set module type names)
> - only phases from same app (should be validate)

To precise this means, that
- there will be at most one module type per app
- there will be exactly one module type if an app defines phases

Thus module information is directly coupled to a module and could be stored in the AppConfig. But to prevent name clashes i decided to maintain a separate registry which is analogous to phases. A ModuleContent object can be retrieved by using the app label or the verbose identifier "fancyapp:module".

As it is only useful (and required) to define module types if an app defines phases i decided to use the phases autodiscovery mechanism and require to define the ModuleContent in the `phases.py`

Currently the ModuleContent only has to define the `app` as the apps label. It is expected that a ModuleContent also stores the translated, human readable name and a description. Although it is not clear to me where the description could be displayed. Probably in the Dashboard on creation.
From a ModuleContent object every phase of the app can be retrieved. Those are the only allowed apps for this module type (as we defined).
The ModuleContent type could be used to filter projects based on the contained modules. Currently there is a very ugly hack to get a "project types" filter which is using the blueprint information.

This is a minimal proposal for module content based on the a4 workshop and i'd like to get back some feedback.
In a future extension the ModuleContent would also contain information on howto export and would be used to decide what settings / menuitems to show in the dashboard.

### DIFF
--- a/adhocracy4/modules/__init__.py
+++ b/adhocracy4/modules/__init__.py
@@ -1,0 +1,3 @@
+from .contents import content, ModuleContent
+
+__all__ = ['content', 'ModuleContent']
--- a/adhocracy4/modules/contents.py
+++ b/adhocracy4/modules/contents.py
@@ -1,0 +1,42 @@
+from adhocracy4.phases import content as phases_content
+
+
+class ModuleContent:
+    @property
+    def identifier(self):
+        return '{s.app}:module'.format(s=self)
+
+    def __str__(self):
+        return '{s.__class__.__name__} ({s.identifier})'.format(s=self)
+
+    def allowed_phases(self):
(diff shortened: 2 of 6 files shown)


kleingeist:
This PR will be closed as it was decided to not use explicit module types at this time.
The app of a model may be retrieved by the django apps app.
And the app of a module by the `PhaseContent.app` property. Note, that this won't work if a module has no phase. But that wouldn't be a valid module in the beginning.

PR #159, opened 2017-08-14
Fix has_feature_active in case module is not really a module

xi:
This fixes a regression introduced in #153.

This workaround works for me. @kleingeist can you please try to find a proper fix?


kleingeist:
i was just checking back into the issue, but the problem is actually within our polls rules definition rather then in core. i will close this PR as it fixes the problem on the wrong end

PR #162, opened 2017-08-21
fixing image deletion

mkind:
* in case of no upload image, `imagefields` contains the value `None`. if so, ignore item.

### DIFF
--- a/adhocracy4/images/services.py
+++ b/adhocracy4/images/services.py
@@ -3,6 +3,7 @@ from easy_thumbnails.files import get_thumbnailer
 
 def delete_images(imagefields):
     for imagefield in imagefields:
+        if not imagefield: continue
         thumbnailer = get_thumbnailer(imagefield)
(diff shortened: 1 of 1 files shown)


adhocracy4/images/services.py
```diff
@@ -3,6 +3,7 @@
 
 def delete_images(imagefields):
     for imagefield in imagefields:
+        if not imagefield: continue
```
slomo: How can that happen? Is that the case if the image has not been set?

And why is the continue on the same line as the if? :smile:

kleingeist:
i also noticed an issue with the image signals. since the update to 1.11 the for loop in `images/signals.py` is  executed multiple times and thus the list in _a4images_image_fields contain multiple entries of the same image field. maybe the error is related?

mkind:
Ok, this is fixed in opin by https://github.com/liqd/a4-opin/pull/928. It is no core issue anymore.

PR #164, opened 2017-08-24
ui to explain follow

vellip:
Clicking the button now opens a dropdown where the specific action is explained.

follow dropdown

This is breaking because it requires additional styling. See https://github.com/liqd/a4-meinberlin/pull/728. At least setting
```css
.select-item-desc {
    display: block;
}
```
is recommended, since I couldn't use any block elements due to then invalid html.

### DIFF
--- a/adhocracy4/follows/static/follows/FollowButton.jsx
+++ b/adhocracy4/follows/static/follows/FollowButton.jsx
@@ -11,14 +11,24 @@ class FollowButton extends React.Component {
       follows: 0
     }
   }
-  toggleFollow () {
-    api.follow.change({ enabled: !this.state.followed }, this.props.project)
+  setFollow (enabled) {
+    if (enabled === this.state.followed) {
+      return false
+    }
+
+    api.follow.change({ enabled: enabled }, this.props.project)
       .done((follow) => {
         this.setState({
           followed: follow.enabled,
           follows: follow.follows
         })
       })
+  }
+  enableFollow () {
+    this.setFollow(true)
+  }
+  disableFollow () {
+    this.setFollow(false)
   }
   componentDidMount () {
     api.follow.get(this.props.project)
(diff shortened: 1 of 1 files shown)


adhocracy4/follows/static/follows/FollowButton.jsx
```diff
@@ -39,10 +49,29 @@ class FollowButton extends React.Component {
...
-          &nbsp;{this.state.followed ? django.gettext('Unfollow') : django.gettext('Follow')}
-        </button>
+        <span className="dropdown">
+          <button className="btn btn-secondary dropdown-toggle" type="button" id="follow-dropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+            <i className="fa fa-star" aria-hidden="true" />&nbsp;
+            {this.state.followed ? django.gettext('Unfollow') : django.gettext('Follow')}&nbsp;
+            <i className="fa fa-caret-down" aria-hidden="true" />
+          </button>
+          <span className="dropdown-menu" aria-labelledby="follow-dropdown">
+            <button className="dropdown-item select-item" onClick={this.disableFollow.bind(this)}>
```
slomo: Is the class name `select-item` not a bit over generic?

vellip: I had something like `follow`before which I found too specific. You have any suggestions for something in between?

adhocracy4/follows/static/follows/FollowButton.jsx
```diff
@@ -39,10 +49,29 @@ class FollowButton extends React.Component {
...
+              {!this.state.followed ? <i className="fa fa-check select-item-indicator" aria-hidden="true" /> : null}
+              {django.gettext('Unfollow')}
```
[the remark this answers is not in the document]
vellip: Valid point, thanks!

PR #165, opened 2017-08-24
Add all static and template folders recursively to package

2e2a:
Adding `static` and `templates` of all apps to the package, to avoid forgetting to add templates of new apps.

@slomo: Or was there a good reason to add them manually?

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,9 +1,4 @@
 include LICENSE
 include README
-recursive-include adhocracy4/images/static *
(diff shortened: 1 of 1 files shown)


slomo:
That was because I couldn't get it working with patterns.

2e2a:
> That was because I couldn't get it working with patterns.

Seems to work this way.

slomo:
I tested it on Friday and apparently forgot to merge it.

PR #166, opened 2017-08-25
small comment changes

vellip:
see commits


slomo:
I think it has come up recently as a requirement that the date of a deleted comment should be shown, and I was about to implement this. Maybe you can just solve in this PR?

kleingeist:
args @slomo  we looked at the PR at the same time and i didn't now about the requirement :(

PR #169, opened 2017-08-28
show [creation] date on deleted comments

vellip:
I don't think showing the last modified date makes sense for a deleted comment.


slomo:
Thanks again

PR #172, opened 2017-08-31
fix predicates to return False if argument is None

MagdaN:
This PR enables django permissions work together with our rules system. Without these changes it is not possible to create a group with certain table-based permissions within the django admin, because the names of our rules have the same name pattern as djangos permissions. 

When the django admin is created django checks all permissions available if one of them returns true, but runs in an error when trying to call a predicate in one of our customs rules, as no project or item (whatever the permission_object is) is set. 

Another solution to fix this, would have been to change the name pattern of our rules, but we preferred this solution, as no more adaptions should be necessary now.

### DIFF
--- a/adhocracy4/comments/rules.py
+++ b/adhocracy4/comments/rules.py
@@ -6,10 +6,12 @@ from adhocracy4.phases import predicates as phase_predicates
 
 @rules.predicate
 def content_object_allows_comment(user, comment):
-    content_object = comment.content_object
-    return phase_predicates.has_feature_active(
-        content_object.module, content_object.__class__, 'comment'
-    )
+    if comment:
+        content_object = comment.content_object
+        return phase_predicates.has_feature_active(
+            content_object.module, content_object.__class__, 'comment'
+        )
+    return False
 
 
 rules.add_perm(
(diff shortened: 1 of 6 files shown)


slomo:
Would it be a good idea if we wrote our own decorator for this?

```python
@rules.predicate
@deny_without_object
 def content_object_allows_comment(user, comment):
     ...
```

MagdaN:
Feel free to update my PR or delete this one and make a new one.

2e2a:
> Would it be a good idea if we wrote our own decorator for this?

This would reduce the need to check the object in all rules.
However, if rules are expected to handle `None` object by the default django authentication backend, explicitly having the check in the rules is also fine.

slomo:
I just had another idea on the commute, maybe if we don't want to use permission checking without  objects at all, we could just extend the [rules permission backend](https://github.com/dfunckt/django-rules/blob/master/rules/permissions.py#L23) and provide a fallback for the case that no object was given:

```
A4PermissionBackend(ObjectPermissionBackend):
   def has_perm(self, user, permission, *args, **kwargs):
     if len(args) == 0:
       return None
    else:
      return super().has_perm(user, permission, *args, **kwargs)
```

Then we only need to configure this in our [base.py](https://github.com/liqd/a4-advocate-europe/blob/master/advocate_europe/settings/base.py#L148).

So in theory this would even allow to solve this in AE without any changes to core.

MagdaN:
But would it not be better to have code in core that does not break the django permissions

PR #173, opened 2017-09-01
check if commenting is allowed for check if comments can be rated, commented, changed, or deleted

fuzzylogic2000:
The predicate, which is tested to decide if comments can be rated, commented, changed, or deleted checked if the phase allowed commenting. Instead it should check and does check now if commenting is allowed for that content_type, like in the comment create rule.

From now on, you'll have to make sure, that you test for the phase in the rule in your project. But you will test it there anyway, otherwise commenting would be allowed, even if the phase doesn't.

### DIFF
--- a/adhocracy4/comments/rules.py
+++ b/adhocracy4/comments/rules.py
@@ -1,15 +1,17 @@
 import rules
 
 from adhocracy4.modules import predicates as module_predicates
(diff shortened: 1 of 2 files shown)


fuzzylogic2000:
Added the same logic for rating.

2e2a:
Nice

PR #175, opened 2017-09-06
fallback tiny image for image upload without picture

vellip:
HTML was invalid before. It's easier than  removing the img tag and dynamically inserting one via JS because that would require a bigger rewrite of the js upload code.

The picture is invisible, it's a transparent gif.


slomo:
Is that the right fix here? Shouldn't be the the img tags removed if there is no picture?

xi:
> Shouldn't be the the img tags removed if there is no picture?

In terms of a11y, I am not sure whether this makes any difference. The img has `alt=""`, so it is marked as presentational anyway.

In terms of effort: We set the `src` attribute on the preview image via javascript as soon as an image is selected. This would be much more work if we had to check whether an img element already exists and create it on demand.

In terms of layout/styling this shouldn't make any difference.

PR #177, opened 2017-09-06
Datetime field

2e2a:
Adds a split datetime field.

Required for dashboard form.


adhocracy4/datetimefield/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
```
slomo: This an app because the template should be over writable? I am not saying we should not do this. But does this approach to add an app for each widget really scale?

I would propose to keep this either as an `adhocracy4.forms` package, or if it has to be `adhcoracy4.forms` could also be an app.

adhocracy4/datetimefield/templates/a4datetimefield/datetime_input.html
```diff
@@ -0,0 +1,9 @@
+<div class="l-col-2 widget--datetimeinput__date">
+    {{ date }}
+</div>
+<div class="l-col-2">
```
slomo: In order to be reusable this should probably have a class like `widget--datetime__time`? I have not that much experience with reusing styling, maybe @xi has some advise here?

adhocracy4/datetimefield/widgets.py
```diff
@@ -0,0 +1,78 @@
...
+
+class DateTimeInput(form_widgets.SplitDateTimeWidget):
+    def __init__(self, time_label='', time_default=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.time_label = time_label
+        self.time_default = time_default or datetime.time(hour=0, minute=0)
+
+    class Media:
+        js = (
+            staticfiles_storage.url('datepicker.js'),
```
slomo: Are those file parts of adhocracy4? Should they be?

adhocracy4/datetimefield/templates/a4datetimefield/datetime_input.html
xi: I see three options

1. Use specific, non-reusable selectors like `a4-datetime__time`. CSS needs to be done in the individual projects, possible by using `@extend` with existing selectors.
2. Use roughly the same selectors and styling in all projects.
3. Overwrite the template in each project. The template here is merely an example for documentation purposes.

We are currently doing a bit of all of them (which I think is totally fine btw). I tend to prefer (3), but that doesn't work for react components.

adhocracy4/datetimefield/apps.py
2e2a: That makes sense. Renamed the app.

adhocracy4/datetimefield/widgets.py
xi: This one might be a deal breaker. We searched long and hard for a good datepicker, and found none. In the end vellip ended up packaging one that has been unmaintained since 2014 at https://github.com/vellip/datePicker so it could be used with npm/webpack.

Our setup is to install it from github via npm, and then bundle it together with an initialisation script into datepicker.js via webpack.

Maybe it's best if we just bypass all that here and just add documentation: This widget will create a regular text input with the class `datepicker`. It will also load `datepicker.js` and `datepicker.css`. The individual projects are in charge of creating these files and implementing the actual datepicker.

MagdaN: If this is the final solution, we should use our own fork of that datepicker.

PR #178, opened 2017-09-06
Add copyright fields

2e2a:
I do not like this, but I had to add the copyright just as new fields to the project model.

Trying to add the copyright field via `contribute_to_class` lead in the end to migration problems. And dynamic fields are discouraged by django anyway:
https://code.djangoproject.com/ticket/22555

Even adding the `copyright_field` as a parameter to the image field, similar to `width_field` of `ImageField`, would require a custom image model.

I would see this as a temporary solution and add a custom image model in the long term.


adhocracy4/images/fields.py
```diff
@@ -1,19 +1,11 @@
 from django.db import models
 from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
-from django.utils.functional import lazy
+from django.utils.text import format_lazy
 
 from . import forms, validators
 
 
-# FIXME: backportet from django 1.11. remove this after update
```
slomo: Opin is and will stay on Django 1.8 for the next months. If Django 1.8 support is removed it should also be made clear in the `setup.py`.

adhocracy4/projects/models.py
```diff
@@ -83,6 +89,12 @@ class Project(base.TimeStampedModel):
         ),
         upload_to='projects/tiles',
         blank=True)
+    tile_image_copyright = models.CharField(
```
slomo: Won't at lease a function here be nice, that takes the other field and generates a copy right field from it?

```
def make_copyright_field(imagefield):
    return models.CharField(
      max_length=120,
      verbose_name=imagefield.verbose_name + ' ' + _('copyright')
     ...
   )
```

adhocracy4/images/fields.py
2e2a: Reverted the commit.

adhocracy4/projects/models.py
2e2a: Added a custom copyright field to set these defaults.

adhocracy4/images/fields.py
```diff
@@ -80,3 +80,16 @@ def _help_text(self):
...
         )
         return help_text
+
+
+class ImageCopyrightField(models.CharField):
+
+    def __init__(self, *args, image_name='image', **kwargs):
+        defaults = {
+            'verbose_name': _('{} copyright'.format(image_name).capitalize()),
+            'help_text': _('Copyright shown in the {}'.format(image_name)),
```
xi: Does string extraction work here?

xi: I checked, it works fine.

PR #181, opened 2017-09-12
clearify connection between "follow" and "email"

xi:
see also https://github.com/liqd/a4-meinberlin/pull/787

we got the feedback that the connection between "email", "notification", and
"follow" is not completely clear.

@janaba what do you think of this change?


2e2a:
Makes sense to me.

PR #182, opened 2017-09-18
Fix image copyright translation

kleingeist:
To allow for correct translation format_lazy has to be used for copyright field name and description.

Note that the capitalization method has been removed. This is fine for german translations where nouns are uppercase within words, but it is not 100% perfect for english, as it results in strings like: 
'Copyright shown in the Header image.'

I still think that this is fine, as we otherwise would have to implement a capitalize_lazy method :/


2e2a:
lgtm

PR #183, opened 2017-09-19
update marker when input changes

xi:
This allows to add alternative input methods, e.g. search by address.

Note that you have to trigger the change event manually, e.g.:

    $('id_' + name)
        .val('{"type":"Feature","properties":{},"geometry":{"type":"Point","coordinates":[13,52]}}')
        .trigger('change')

### DIFF
--- a/adhocracy4/maps/static/a4maps/map_choose_point.js
+++ b/adhocracy4/maps/static/a4maps/map_choose_point.js
@@ -128,6 +128,18 @@ var init = function () {
         $('#id_' + name).val(JSON.stringify(shape))
       }
     })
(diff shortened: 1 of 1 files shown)


xi:
If this is accepted, we might consider using the same mechanism to implement polygon presets (#74)

kleingeist:
i think this is an appropriate solution. 
i overcame my initial concerns, that the value/marker is set without checking the boundaries, as this is an internal, not a public interface and we can assume and demand that the project specific code checks the boundaries itself. 
as this is no breaking change i will merge the changes

PR #184, opened 2017-09-19
avoid ES6 in map code

xi:
This function was introduced in 59b6f46.

I missed the fact that this file is not processed by webpack/babel, so we can not use ES6 here.


kleingeist:
is the `for (let line ...)` in line https://github.com/liqd/adhocracy4/pull/184/files#diff-7e649bf207ad127e1db434b206e4bf28L63 valid prior to ES6?

kleingeist:
lgtm

Issue #134, opened 2017-07-18
Refactoring plan


slomo:
fill me with refactoring that are needed

kleingeist:
The project/module may also be derived from the Item itself:
f.ex. in detail the detail view:
* read `module` property of an 'item', which is a foreign key. the 'item' itself is determined via the `slug` param. an 'item' may be f.ex. an `Idea`, a `Poll` or an `OfflineEvent`

slomo:
I edited the proposal to include reading of the module property.

slomo:
Should we split each refactoring into its own issue, and just collect the list of issues here? That would allow to discuss each on it own.

kleingeist:
Altough the deadline is closed for refactoring suggestions, i think we should try to make the api.js extensible without having to commit to the core as our projects are currently leaking into the core that way.

PR #185, opened 2017-09-20
Project Dispatcher

2e2a:
Implements the project/module view helper suggested in #134 

It sets the module and project in different view scenarios:

- kwargs
- url
- get_object

This is a copy of the working solution from a4-meinberlin and could be used in the core dashboard as well.


kleingeist:
closed in favor of #206

PR #188, opened 2017-09-26
Use prop-types package (fixes react deprecation warnings)

kleingeist:
React emitted the appended warning that using `React.PropTypes` is
depreacted in favor of the separate package 'prop-types'. This PR adapts
to these changes.

> Accessing PropTypes via the main React package is deprecated, and will
> be removed in  React v16.0. Use the latest available v15.* prop-types
> package from npm instead. For info on usage, compatibility, migration
> and more, see https://fb.me/prop-types-docs


xi:
LGTM.

PR #189, opened 2017-09-26
workaround string extraction bug in followButton

xi:
fixes https://github.com/liqd/a4-meinberlin/issues/832

I am not really sure why string extraction doesn't work here. I guess that django is confused about some JSX.


kleingeist:
the message extractor stumbles upon the `{!this.state.followed ? <i ... /> : null}`. using `{!this.state.followed ? <i ... ></i> : null}` would work but create an empty element :( adding  parentheses like`(<i.../>)`  or whitespaces doesn't work either. 

as this seems like an upstream bug and is the only place we run into it and this is a simple fix i would merge.

PR #190, opened 2017-09-27
more descriptive help texts on project

xi:
We need to make sure that these help texts are correct for all projects.

Fixes https://github.com/liqd/a4-meinberlin/issues/824 and https://github.com/liqd/a4-meinberlin/issues/823


slomo:
After merging alexejs older PR this has now conflicting migrations.

PR #193, opened 2017-09-28
Add categorie filters

slomo:
The second commit  is a refaftoring for the filterset, which introduces more code than it deletes. If we have no use for the view property in other filters I can remove it. The first one should work on its own.


kleingeist:
i like the class based view aware approach, as it fit's fine into our architecture. but i am not sure if we need it somewhere else.

also i'm not sure about the most "djangonic" way to solve this. it feels weird to set the module and project properties on the request, as they seem clearly view related to me. but on the other hand it feels also weird passing the whole view around. we thought about passing it to forms too, but decided to pass the required args explicitly in the end. unfortunately this is not possible for filters :(

so as said above i prefer passing the view to the filter instead of populating the request object :+1:

kleingeist:
lgtm :)

PR #196, opened 2017-10-11
Prevent image fields from being multiple registered in case of multi table inheritance

kleingeist:
With multi table inheritance the image fields from the parent models
have been added multiple times to the image_fields list.
This limits the fields that may be added to the models concrete fields
and additionally ensures overwritten fields may not be added twice.

### DIFF
--- a/adhocracy4/images/signals.py
+++ b/adhocracy4/images/signals.py
@@ -41,11 +41,12 @@ def delete_images_cascaded(sender, instance, **kwargs):
 
 # Setup signals for all ConfiguredImageFields
 for model in apps.get_models():
-    for field in model._meta.get_fields():
+    for field in model._meta.get_fields(include_parents=False):
         if isinstance(field, ConfiguredImageField):
             image_fields = getattr(model, _IMAGE_FIELDS_ATTR, [])
-            image_fields.append(field.attname)
-            setattr(model, _IMAGE_FIELDS_ATTR, image_fields)
+            if field.attname not in image_fields:
+                image_fields.append(field.attname)
+                setattr(model, _IMAGE_FIELDS_ATTR, image_fields)
 
     if hasattr(model, _IMAGE_FIELDS_ATTR):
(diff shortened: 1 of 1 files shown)


adhocracy4/images/signals.py
```diff
@@ -41,11 +41,12 @@ def delete_images_cascaded(sender, instance, **kwargs):
...
 for model in apps.get_models():
-    for field in model._meta.get_fields():
+    for field in model._meta.get_fields(include_parents=False):
         if isinstance(field, ConfiguredImageField):
             image_fields = getattr(model, _IMAGE_FIELDS_ATTR, [])
-            image_fields.append(field.attname)
-            setattr(model, _IMAGE_FIELDS_ATTR, image_fields)
+            if field.attname not in image_fields:
+                image_fields.append(field.attname)
+                setattr(model, _IMAGE_FIELDS_ATTR, image_fields)
```
xi: Would it make sense to use a set instead of a list here?

kleingeist: yes

fuzzylogic2000: Does that mean this PR is incomplete?

kleingeist: ah sorry @fuzzylogic2000  i forgot to change those lines resp. adding the incomplete label

kleingeist: @xi i looked into this issue again, but at this place a `set` is not usable, as it guarantees no order of it's elements at all.

when models with images are updated, the current images values are stored aside to the list of image fields in an index based manner. both lists are zipped later on. thus the list of fields has to any fixed order.
while it would be possible to use a set to construct the fields "list" first and then convert the sets to lists or tuples with an order i would like to keep the current simple approach. especially as we have currently of maximum 2 images on a single model. and we don't expect to have a great number of images in the future.

xi: Agreed. If the order is relevant we should use a list.

2e2a:
lgtm

PR #198, opened 2017-10-16
add normal user and template tag to discover cta-buttons

fuzzylogic2000:
Adds a "normal user", who has no special permissions, but checks if a logged-in user would be allowed to do whatever the button promises.
The template tag can be used analogous to has_perm from the rules tags. Like here: https://github.com/liqd/a4-opin/pull/962/files
For the comments, the link "Please login to comments" is only displayed, when commenting is possible.

### DIFF
--- a/adhocracy4/comments/templatetags/react_comments.py
+++ b/adhocracy4/comments/templatetags/react_comments.py
@@ -6,6 +6,7 @@ from django.utils.html import format_html
 
 from ..models import Comment
 from ..serializers import ThreadSerializer
+from adhocracy4.rules.discovery import NormalUser
 
 register = template.Library()
 
@@ -28,6 +29,8 @@ def react_comments(context, obj):
     permission = '{ct.app_label}.comment_{ct.model}'.format(ct=contenttype)
     has_comment_permission = user.has_perm(permission, obj)
 
+    would_comment_permission = NormalUser().would_perm(permission, obj)
+
     comments_contenttype = ContentType.objects.get_for_model(Comment)
     pk = obj.pk
 
@@ -44,6 +47,7 @@ def react_comments(context, obj):
         'isModerator': is_moderator,
         'user_name': user_name,
         'isReadOnly': not has_comment_permission,
+        'would_comment_perm': would_comment_permission,
     }
 
     return format_html(
--- a/adhocracy4/comments/templatetags/react_comments.py
+++ b/adhocracy4/comments/templatetags/react_comments.py
@@ -29,7 +29,9 @@ def react_comments(context, obj):
     permission = '{ct.app_label}.comment_{ct.model}'.format(ct=contenttype)
     has_comment_permission = user.has_perm(permission, obj)
 
-    would_comment_permission = NormalUser().would_perm(permission, obj)
+    would_have_comment_permission = NormalUser().would_have_perm(
+        permission, obj
+    )
 
     comments_contenttype = ContentType.objects.get_for_model(Comment)
     pk = obj.pk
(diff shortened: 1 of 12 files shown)


adhocracy4/rules/templatetags/discovery_tags.py
```diff
@@ -0,0 +1,14 @@
+from django import template
+
+from adhocracy4.rules.discovery import NormalUser
+
+register = template.Library()
+
+
+@register.assignment_tag
```
xi: The assignment tag is deprecated since django 1.9. We should use simple tags instead.

fuzzylogic2000: To keep it compatible with 1.8 and OPIN, we need to use the assignment_tag.

adhocracy4/comments/templatetags/react_comments.py
```diff
@@ -44,6 +49,7 @@ def react_comments(context, obj):
         'isModerator': is_moderator,
         'user_name': user_name,
         'isReadOnly': not has_comment_permission,
+        'would_have_perm': would_have_comment_permission,
```
kleingeist: the default javascript style is to use camelCase variables. in meinBerlin we are using a linter which checks it. i'd prefer to use `wouldHavePerm` 

or we could simplify the code overall and just use `isReadOnly` and `isAuthenticated`
`isReadOnly` would be set to `not has_comment_permission and not would_have_comment_permission` and the check within react would be 
```
if (this.context.isAuthenticated && !this.props.isReadOnly) {
// show
} else if (!this.props.isReadOnly) {
// login
} else {
// no party
}
```

i think it would simplify the code as it keeps the generic sounding `would_have_perm` out of the react component. 

but those are just my personal opinions. i would not veto against the current solution. especially as "user_name" is already not camelcase

adhocracy4/ratings/templatetags/react_ratings.py
```diff
@@ -45,6 +50,7 @@ def react_ratings(context, obj):
         'userRatingId': user_rating_id,
         'isReadOnly': not has_rate_permission,
         'style': 'ideas',
+        'would_have_perm': would_have_rate_permission,
```
kleingeist: while the would_have_perm is set it is not used in the react_ratings.jsx. The rating buttons are disabled if isReadOnly is set to true. either the button would have to be enabled if `!isReadOnly || would_have_perm` or the isReadOnly could be adapted in the manner suggested above.

for login redirections maybe the lines [58](https://github.com/liqd/adhocracy4/blob/588e61dccf739f6836458a747fc4dfc73b6432bb/adhocracy4/ratings/static/ratings/react_ratings.jsx#L58) and [79](https://github.com/liqd/adhocracy4/blob/588e61dccf739f6836458a747fc4dfc73b6432bb/adhocracy4/ratings/static/ratings/react_ratings.jsx#L79) of react_ratings.jsx would probably have to be adapted.

adhocracy4/comments/templatetags/react_comments.py
fuzzylogic2000: Good idea, I will use isReadOnly!

adhocracy4/ratings/templatetags/react_ratings.py
fuzzylogic2000: I wasn't quite sure, if this didn't have to wait for the next user story, because there should be a pop-up to login for rating. But with the isReadOnly the user is simply forwarded to login, and the pop-up has to wait until the next user story.

kleingeist:
works nice :) 
will merge and wait for the react cta logic in the next spring :)

PR #199, opened 2017-10-17
Update to DRF 3.7

slomo:


kleingeist:
we checked and there are no problems with meinBerlin. 
thanks for adapting

PR #201, opened 2017-10-18
Fix fakers for project names (titles)

### LINKED ISSUE #200: Faker generates project names with newlines (which breaks notification mails)
Sometimes tests fail because emails don't allow newlines in there subject. This results from fakers behavior for generating "text". If the requested text should have more or equal then 100 chars it adds newlines between sentences. If no char limit is set it defaults to 200.
We are using the text generator with a char limit of 120 for the Organisation and Project name where we expect single line texts. As the project name is part of email notification subjects the tests fail.
In other places we use the "name" faker instead of the the "text" faker but the "name" faker does not allow to limit the number of chars. And it returns relatively short human names which is not what we expect for project names (which are actually titles).

I see 4 fixes for the test problems:
1. Replace "text" with "name" or "sentence" when expecting single lines. 
   As it is not possible to limit the number of chars, the faker may generate data that is not valid in regard to the database.
2. Set the char limit to <100 to prevent newlines, as we have internal knowledge of the faker code. But we may never now 
3. Write a custom provider "title" which allows to limit the char number but does not create newlines.
    Unfortunately the custom provider would have to be added in every projects conftest.py and creating the example provider as documented by faker did not work :(
4. File a bug report on faker to add a char limit to the "name" or "sentence" or a new "title" and hope for the best. In the meantime we could fall back to 1.

kleingeist:
Replace the "text" faker with the "sentence" faker in places where a name/title is expected. 
Using "text" may result in newlines which does not work with email subjects.
This kind of fixes #200 although it might be possible that the database length validation may be violated.

### DIFF
--- a/adhocracy4/test/factories/__init__.py
+++ b/adhocracy4/test/factories/__init__.py
@@ -44,7 +44,7 @@ USER_FACTORY = getattr(settings, 'A4_USER_FACTORY', UserFactory)
 class OrganisationFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = Organisation
-    name = factory.Faker('text', max_nb_chars=120)
+    name = factory.Faker('company')
 
 
 ORGANISATION_FACTORY = getattr(settings,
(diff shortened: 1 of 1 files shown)


2e2a:
lgtm

PR #204, opened 2017-10-23
Make category field translateable

kleingeist:
The category fields had not been translated as the field had been
reinitialized instead of modified. With this PR the auto generated field
(with its localized verbose_name) will be kept and adapted to the
filtered queryset.

Furthermore the required parameter is set to true if there are any
categories for the module.


2e2a:
lgtm

PR #205, opened 2017-10-23
trigger change events on maps

xi:
In meinberlin we have a script that warns users before leaving a changed form. This fixes that script for maps.


slomo:
Maybe the script should also be a core feature, we have a similar script in [ae](https://github.com/liqd/a4-advocate-europe/blob/master/apps/ideas/static/advocate_europe_ideas/form.js).

xi:
> Maybe the script should also be a core feature, we have a similar script in ae.

I think we would need some discussion as the [meinberlin script](https://github.com/liqd/a4-meinberlin/blob/master/meinberlin/apps/contrib/static/js/unload_warning.js) is different in some important aspects.

PR #206, opened 2017-10-24
Project Mixin extension

kleingeist:
Implements the project/module view helper suggested in #134
Replaces  #185 and #115

It sets two properties `project` and `module` on each inheriting view. 
The properties may be derived
1. directly from a `project` or `module` kwarg as set by the Phase- and ModuleDispatcherMixin
2. from a configurable url parameter which defaults to `project_slug` and `module_slug`
3. from the object returned by `get_object` if the view property `get_context_from_object = True`

This PR is currently breaking due to commit 286d7bc4316a5613c9580 which removed the `phase` property from the mixin. It is not clear if the `phase` property is required by opin or if it is replaced by the `ProjectPhaseMixin`.

This is a copy of the working solution from a4-meinberlin and could be used in the core dashboard as well.

### DIFF
--- a/adhocracy4/projects/mixins.py
+++ b/adhocracy4/projects/mixins.py
@@ -1,6 +1,11 @@
+from django.http import Http404
 from django.http.response import HttpResponseRedirect
+from django.shortcuts import get_object_or_404
 from django.utils.functional import cached_property
 from django.views import generic
+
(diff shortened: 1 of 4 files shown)


slomo:
Well if you think its fine, just go ahead.

Issue #86, opened 2017-05-03
Define Content Security Policy


xi:
A [Content Security Policy](https://www.w3.org/TR/CSP/) (CSP) tells browsers to restrict some features. For example, it can be used to disallow inline JavaScript. [django-csp](https://django-csp.readthedocs.io/en/latest/index.html) provides django integration.

It would be nice if we could define a CSP in order to enhance security. The biggest step required for this is to remove all inline JavaScript. In opin, most of the work has already been done in https://github.com/liqd/a4-opin/pull/775. Similar changes should also be implemented here.

xi:
So what CSP do we want? Please add to this list of requirements:

- I would propose to use `default-src 'self'` as a baseline. We could also do `default-src 'none'` to be even more strict, but I don't think this is necessary.
- We often use inline `style="background-image: url(…)"`. There could be ways around this, but maybe it is just ok to allow unsafe-inline styles.
- Maps need to load their tiles, so we need to add the tileserver to `img-src`.
- django-debug-toolbar loads jquery from http://ajax.googleapis.com and has an inline image. This is only relevant for development, but still we need to think about this.
- We use webpack with [`devtool: 'eval'`](https://webpack.js.org/configuration/devtool/). `eval()` is forbidden by default. So we need to either switch to a slower mechanism for sourcemaps or have different webpack config for development and production.

So here is a full CSP (for production) that you can drop into your base template:

```
<meta http-equiv="Content-Security-Policy" content="default-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' https://*.tile.openstreetmap.org/">
```

xi:
A note on development setup: There is also the `Content-Security-Policy-Report-Only` header that reports violations but does not stop them.

mkind:
Thanks for the issue :+1: 

> We often use inline style="background-image: url(…)". There could be ways around this, but maybe it is just ok to allow unsafe-inline styles.

Inline styles might get ugly from a [security point of view](https://scarybeastsecurity.blogspot.de/2009/12/generic-cross-browser-cross-domain.html).

mkind:
AFAIK CSP has an option to prevent clickjacking, too. We might also take this into account.

kleingeist:
not sure where to document meinBerlin specific issues. please move this if it is at the wrong place:

[...]

PR #207, opened 2017-10-30
avoid using inline script

xi:
another step towards a content security policy.

You need to initialise the react components yourself. See https://github.com/liqd/a4-opin/pull/775 for an existing implementation in opin.


xi:
See #86

PR #210, opened 2017-11-13
Ensure time input field of DateTime is always optional.

### LINKED ISSUE #187: Keep html time field optional even if a DateTimeField is set to be required
Time is optional in the backend but the html input field still gets the required attribute if the the DateTimeField is initializes with `required=True`
The time Widget should always be initialized without required.

kleingeist:
As the DateTime Field declares the time part as optional with a fallback
to a default value, the associated input element as declared by the
widget may never be declared as required.
Fixes #187


xi:
Nice!

PR #211, opened 2017-11-14
Strip newlines from email subjects and make SyncEmail a Mixin again

### LINKED ISSUE #174: Subject with new line crashes email sending

kleingeist:
This fixes #174 by replacing newlines from the subject.

Furthermore it makes the SyncEmailMixin a real Mixin again by removing it's inheritance from EmailBase. This may be breaking if
* SyncEmailMixin was used as the sole dependency of a concrete Email implementation
* SyncEmailMixin was defined after the Email or EmailBase dependency

In those cases the application code has to be adapted and the SyncEmailMixin used like:
```
from adhocracy4.emails import Email
from adhocracy4.emails.base import EmailBase
from adhocracy4.emails.mixins import SyncEmailMixin

class SomeEmail(SyncEmailMixin, EmailBase):
    pass

class SomePlatformMixinEmail(SyncEmailMixin, Email):
    pass
```

### DIFF
--- a/adhocracy4/emails/base.py
+++ b/adhocracy4/emails/base.py
@@ -1,3 +1,5 @@
+import re
+
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.contrib.sites import models as site_models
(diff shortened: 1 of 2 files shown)


kleingeist:
@slomo thats not really necessary for the dev branches on meinBerlin and product. i'll merge as soon as core is ready.

PR #212, opened 2017-11-15
Use kwargs with reverse

2e2a:
In a4-product we patch reverse to have support for the partner middleware.  However it works only with kwargs.

Fixed the reverse usage here, to be able to use get_absolute_url of project and module in a4-product.


kleingeist:
lgtm and is not breaking. merged.

Issue #88, opened 2017-05-04
Comments and Answers Redesigned


damusp:
The way comments are structured has been bothering me for quite a long time. In my opinion, it's weird how the "answer" link is positioned in mid air, losing all visual connection.

Here is a screenshot for comparison. 
On the left hand side: status quo on Opin
Right hand side: design for Advocate Europe

<img width="1156" alt="bildschirmfoto 2017-05-04 um 13 52 51" src="https://cloud.githubusercontent.com/assets/22769111/25702601/0b06d528-30d2-11e7-98fd-4d4f1f72c23d.png">

I think keeping all actions on one side will make the whole thing easier to digest and visually more consistent. At the end of the day, more motivating to participate!

xi:
Currently the "answer" button is on the same line as the "Show all 5 answers" button. Both buttons have the exact same result (toggle showing the answers and answer form). Can you expand on how that changes? Maybe add screens of how this would look with answers or the answer form?

damusp:
Hey @xi, here's another design from Advocate Europe:
![answers](https://cloud.githubusercontent.com/assets/22769111/25802238/b996ad3a-33f1-11e7-9086-292125c18679.png)

In this concept the buttons do not result in the same action. "Show all x comments" will show all comments, "Answer" will show the answer form. So each one does what it says, which is quite convenient I think. 

Comments on YouTube actually work almost identical. The main difference is that YouTube also allows answering answers, which we do not (or should we?).

mkind:
as #216 is merged I close this issue

PR #216, opened 2017-11-27
move comment reply button into action bar

xi:
see #88


xi:
This probably needs changes to styling, so I added the "breaking-changes" label.

PR #217, opened 2017-11-28
Add error handling for comments

### LINKED ISSUE #95: Comments react component is missing error handling
Currently the comment react components are missing any error handling when a comment is submitted. 
 If there is a validation or connection error the user will not be informed that her comment has not been saved. 
(see also liqd/a4-meinberlin#362)

xi:
This adds error messages when creating or editing comments fails. I tested it by stopping the server after I had loaded the page.

I am not all too happy with this PR. It does what it is supposed to be, but I found it hard to find my way around the comment code, simply because it is complex already. I hope I was able to refactor my additions to a point where review is possible.

Fixes #95


slomo:
I tested it on advocate-europe, it works nicely. Altough I don't know what to do stylingwise.

PR #223, opened 2017-12-13
Set module descriptions maxlength to 512 chars

kleingeist:
This is due to a requirement in product. And also it is absurd, that the
title may have more chars then the description o.0


kleingeist:
args. what is wrong with travis :(

PR #224, opened 2017-12-14
Update react to version 16

slomo:
I think we can do this now. Mein berlin seemed to be ready long ago I just updated advocate-europe and opin to react16 as well.


kleingeist:
works for me locally. although npm is acting weird as always. not sure why npm link is working only half on my local machine. still meinberlin was built with react 16 now. will test again after this is merged and will fix any problems on meinBerlin then

PR #227, opened 2017-12-20
Remove weight from phase type identifier

slomo:


slomo:
Breaking because needs updates in hard coded phase factories and other places where identifier is used explicitly.

adhocracy4/phases/migrations/0006_remove_weight_from_phase_type.py
```diff
@@ -0,0 +1,33 @@
...
+    for phase in Phase.objects.all():
+        parts = phase.type.split(':')
+        phase.type = ':'.join([parts[0], parts[2]])
+        phase.save()
+
+def add_weight_from_type_string(apps, schema_editor):
+    Phase = apps.get_model('a4phases', 'Phase')
+    for phase in Phase.objects.all():
+        parts = phase.type.split(':')
+        phase.type = ':'.join([parts[0], phase.weight, parts[1]])
```
kleingeist: `phase.weight` is an int and can't be joined. should be `str(phase.weight)`

PR #228, opened 2017-12-20
Add profiling middleware

slomo:
I am not sure if we even should add it, also due to possible license issues. But it was extremely useful to debug slow view.s


kleingeist:
i really like the possibility to profile the app. but i agree, that the licence is not clear :(

Note: there is also a derived middleware with additional info:
https://www.gun.io/blog/fast-as-fuck-django-part-1-using-a-profiler

PR #229, opened 2017-12-21
Improve comment anwser button

slomo:
- add class for better styling
 - make labeling of "show more comments" more intuitive
 - set focus on lciking "anwser" button


kleingeist:
works way better then before. thanks for the changes :)
maybe we should try to re-evaluate the autofocus feature for the documents in meinberlin again. sometime....

PR #230, opened 2018-01-02
read error message from api

### LINKED ISSUE #226: Increase comment limit to 2048 and display better error message
Take error message from REST response and display in existing error message alert. On top it probably makes sense to limit the field as well.

MagdaN:
fixes #226


adhocracy4/comments/static/comments/CommentBox.jsx
```diff
@@ -51,12 +51,16 @@ class CommentBox extends React.Component {
           })
         }
       })
-      .fail(respone => {
+      .fail(response => {
+        var errorMessage = JSON.parse(response.responseText).comment[0]
```
kleingeist: why did you choose `JSON.parse` over the jquery response object `response.responseJSON` [1]. in meinBerlin the later is used [2] but i don't know why

[1] http://api.jquery.com/jquery.ajax/#data-types
[2] f.ex. https://github.com/liqd/a4-meinberlin/blob/310531ffbe8840f280ba7e2daa6b3b7393921192/meinberlin/apps/documents/assets/DocumentManagement.jsx#L272

adhocracy4/comments/static/comments/CommentBox.jsx
```diff
@@ -51,12 +51,16 @@ class CommentBox extends React.Component {
...
         if (typeof parentIndex !== 'undefined') {
-          this.updateStateComment(parentIndex, undefined, {replyError: true})
+          this.updateStateComment(
+            parentIndex,
+            undefined, {replyError: true, errorMessage: errorMessage})
         } else {
           this.setState({
-            error: true
+            error: true,
+            errorMessage: errorMessage
```
kleingeist: the states `errorMessage` is not reset only the `error` flag. i think that's okay, but i am not sure what the most reactish way would be.

adhocracy4/comments/static/comments/CommentBox.jsx
MagdaN: I will check

kleingeist:
lgtm. i am fine with merging. i just have two small notes/questions i added as code comments

adhocracy4/comments/static/comments/CommentBox.jsx
MagdaN: I am not sure, if I understand what you mean. Do you mean I should set the errorMessage also if the submit does not fail? What value would you set it to? Undefined, empty String?

Issue #213, opened 2017-11-15
Extend RichtTextFields to auto clean html


kleingeist:
Currently Uploading|RichTextFields are cleaned from unallowed html by using the models save() method [1].
As this is prone for errors because devs forget to call the clean method manually we should extend Uploading|RichTextFields to auto clean their html. 
This could be achieved by using a custom field that overwrites the pre_save method [2] as described by [3].

[1] https://github.com/liqd/a4-meinberlin/blob/master/meinberlin/apps/ideas/models.py#L44
[2] https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.Field.pre_save
[3] https://docs.djangoproject.com/en/1.11/howto/custom-model-fields/#preprocessing-values-before-saving

PR #234, opened 2018-01-11
Mn 2018 01 fix richtext field export errors

MagdaN:
It seems to me that the RichTextField saves the data in ascii/html and not in utf8, therefore text from those fields is not exported correctly. Using lxml fixed the problem for me with excel. 

But I am not sure if there could be a better solution for this as lxml is quite huge and by adding it here we would add to all our projects.

### DIFF
--- a/adhocracy4/exports/views.py
+++ b/adhocracy4/exports/views.py
@@ -84,6 +84,11 @@ class SimpleItemExportView(AbstractXlsxExportView,
         except TypeError:
             pass
 
+        # Return the get_field_display value for choice fields
+        get_field_display_method = getattr(
+            item, 'get_{}_display'.format(name), None)
+        if get_field_display_method and callable(get_field_display_method):
+            return get_field_display_method()
         # Finally try to get the fields data as a property
         return str(getattr(item, name, ''))
 
--- a/adhocracy4/exports/views.py
+++ b/adhocracy4/exports/views.py
@@ -1,17 +1,17 @@
+import lxml.html
+
 from collections import OrderedDict
 import xlsxwriter
 
 from django.views import generic
 from django.http import HttpResponse
 from django.utils import timezone
-from django.utils.html import strip_tags
 from django.utils.translation import ugettext as _
 
 from .mixins import VirtualFieldMixin
 
 
 class AbstractXlsxExportView(generic.View):
-
 
     def get_filename(self):
         return '%s.xlsx' % (self.get_base_filename())
@@ -40,7 +40,6 @@ class AbstractXlsxExportView(generic.View):
 
         return response
 
-
     def _clean_field(self, field):
         if isinstance(field, str):
             return field.replace('\r', '')
@@ -57,6 +56,10 @@ class SimpleItemExportView(AbstractXlsxExportView,
     def _setup_fields(self):
         raise NotImplementedError
 
+    def _get_field_type(self, item, name):
+        if hasattr(item, name):
+            return item._meta.get_field(name).__class__.__name__
+
     def get_header(self):
         return self._header
 
@@ -68,6 +71,7 @@ class SimpleItemExportView(AbstractXlsxExportView,
                           timezone.now().strftime('%Y%m%dT%H%M%S'))
 
     def get_field_data(self, item, name):
+
         # Use custom getters if they are defined
         get_field_attr_name = 'get_%s_data' % name
         if hasattr(self, get_field_attr_name):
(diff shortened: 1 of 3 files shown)


mkind:
Is there a connection to issue #213?

MagdaN:
@mkind not really, I would say, as this code is only used for exporting the textfield data to excel, and removes all html related characters from text, not just the ones which are not allowed

adhocracy4/exports/views.py
```diff
@@ -84,14 +88,27 @@ def get_field_data(self, item, name):
...
         return str(getattr(item, name, ''))
 
     def get_link_data(self, item):
         return self.request.build_absolute_uri(item.get_absolute_url())
 
-    def get_description_data(self, item):
-        return strip_tags(item.description).strip()
+    def get_rich_text_field_data(self, text):
+        html = lxml.html.fromstring(text)
+        return html.text_content()
```
mkind: is the result the same as calling `clean_html_all` from [transforms.py](https://github.com/liqd/adhocracy4/blob/master/adhocracy4/transforms.py#L6)?

kleingeist:
> Is there a connection to issue #213?
@mkind no issue #213 is to get rid of the need to call the html cleaning manually in the `save` method. this is related to the export.

but i don't quit eget the problem or the proposed solution. if it is an encoding error we could just encode via python stdlib functions.

if it is things like `&gt;` instead of `>` i don't get how lxml fixes the problem.

adhocracy4/exports/views.py
```diff
@@ -84,14 +88,27 @@ def get_field_data(self, item, name):
...
 
+        # Return the get_field_display value for choice fields
+        get_field_display_method = getattr(
+            item, 'get_{}_display'.format(name), None)
+        if get_field_display_method and callable(get_field_display_method):
+            return get_field_display_method()
+
+        fieldtype = self._get_field_type(item, name)
+
+        if fieldtype and fieldtype == 'RichTextField':
```
kleingeist: i don't like the approach to make the `get_field_data` even more complicated. i think it would be better to put logic like the html cleaning to the concrete get_X_data method instead.
maybe we could also remove the call to the `get_X_display` and the logic to handle dicts, too (especially as i realized, that that the call `return item['name']` should be `return item[name]` instead. which we never realized as the the export is never called on lists of dicts)

adhocracy4/exports/views.py
```diff
@@ -84,14 +88,27 @@ def get_field_data(self, item, name):
...
+
         # Finally try to get the fields data as a property
         return str(getattr(item, name, ''))
 
     def get_link_data(self, item):
         return self.request.build_absolute_uri(item.get_absolute_url())
 
-    def get_description_data(self, item):
-        return strip_tags(item.description).strip()
+    def get_rich_text_field_data(self, text):
```
kleingeist: if we keep this general method i would prefer to rename it, as the naming scheme was `get_X_data` where `X` is standing for a (virtual) field which is represented as a column in the export.

MagdaN:
I added this based on: 
https://stackoverflow.com/questions/37486/filter-out-html-tags-and-resolve-entities-in-python

adhocracy4/exports/views.py
MagdaN: I was thinking about this, but I was afraid, that we would then miss this at some point, because we forget or don't know that a field is a RichTextFiled

adhocracy4/exports/views.py
MagdaN: Sure, what would you suggest? We could extend this for more fields

adhocracy4/exports/views.py
kleingeist: but if it is the html encodings then the problem occurs for comments, too

MagdaN: No, because it is not a RichTextField

kleingeist: then, as guesse above, i don't get the problem that is fixed with this pr
if i enter `greater > test` as a comment it gets stored as `greater &gt; test` to the database and exportet as `greater &gt; test` in the comments

MagdaN:
@kleingeist works nicely with html from python, thanks! I removed get_field_data and changed moved the functationality back to get_description_data but was wondering if get_description_data should actually be in A4

adhocracy4/exports/mixins.py
```diff
@@ -87,14 +85,14 @@ def _flat_comments(self, item):
...
-                text=strip_tags(comment.comment).strip()
+                text=self.strip_and_unescape_html(comment.comment)
             )
 
             for reply in comment.child_comments.all():
                 yield self.REPLY_FMT.format(
                     date=reply.created.isoformat(),
                     username=reply.creator.username,
-                    text=strip_tags(reply.comment).strip()
+                    text=self.strip_and_unescape_html(reply.comment).strip()
```
kleingeist: the last `strip()` is not needed

adhocracy4/exports/views.py
```diff
@@ -67,7 +67,12 @@ def get_base_filename(self):
         return '%s_%s' % ('download',
                           timezone.now().strftime('%Y%m%dT%H%M%S'))
 
+    def strip_and_unescape_html(self, text):
+        text = strip_tags(text).strip()
+        return html.unescape(text)
```
kleingeist: sorry for all those comments:
it would make sense to call `html.unescape` prior to`strip()` as it might replace `&nbsp;` with actual whitespaces which then may be removed by strip

MagdaN:
@kleingeist changed it

PR #235, opened 2018-01-15
Collapsible CKEditor fields

kleingeist:
This intorduces a ckeditor extension with support for collapsible elements. 
This is breaking as it is using a new config `collapsible-image-editor` which has to be set in the project specific django configs for `BLEACH_LIST` and for the `CKEDITOR_CONFIGS`
To show the button within the CKEditor `['CollapsibleItem'],` has to be added to `toolbar_Custom`

Currently the collapsible fragment is very simple and does not set any classes. in meinberlin the fragment is overwritten to include the "collapsible" classes. should i move those classes to the a4 core or keep it simple and motivate to overwrite the template?


mkind:
> should i move those classes to the a4 core or keep it simple and motivate to overwrite the template?

keep it simple in my opinion

tests/ckeditor/test_ckeditor_templatetags.py
```diff
@@ -0,0 +1,17 @@
...
+
+from adhocracy4.test.helpers import render_template
+
+
+@pytest.mark.django_db
+def test_transform_collapsibles(project_factory):
+    project = project_factory(information='<div class="collapsible-item">'
+                              '<div class="collapsible-item-title">Title</div>'
+                              '<div class="collapsible-item-body">Body</div>'
+                              '</div')
```
mkind: The tag misses a closing `>`.

mkind:
Including `<div>`-tags works. Why?

kleingeist:
> Including <div>-tags works. Why?

what do you mean by that? to allow for the `<div class="collapsible-item">..</div>` block a new BLEACH config had to be introduced which allows `<div>`s with the `class` attribute. But bleach does not allow to only allow `<div>`s with a certain set of classes. thus if a collapsible field is used, users may enter any number of `<div>` tags

kleingeist:
additional information from the original experimental branch https://github.com/liqd/a4-meinberlin/pull/1131

> IMHO this allows for an easier interface as the collapsibles may be defined at the same place as the other information. And it is more flexible as it allows to add the collapsible at any place within the text.
> 
> But there are also downsides:
> - to prevent from allowing to many elements/attributes in user entered fields the ckeditor part is kept simple and contains only 3 divs with classes
>    - still bleach had to be extended to allow for divs and the class attribute. it is not possible to restrict the class to a specific value. i am not sure if setting classes could be a security concern
>  - for rendering the simple html has to be transformed to the more complex dom with correct collapsible definitions. the code for this is kind of complex:
>    - the text has to be parsed and an element tree has to be built
>    - all the collapsible elements have to be replaced by the correct html. which is involves a render_template call. (advantage: this allows for custom elements/classes as needed by bs3 vs bs4)
>    - the original tree has to be modified and serializes again
> - currently the block are not collapsible within the ckeditor as this would require a lot of additional markup to retain a11y
> - django-ckeditor is broken and won't allow to add additional plugins from the global config, thus a new field type is required
>

PR #236, opened 2018-01-16
 Refactor Export

kleingeist:
# Refactor SimpleItemExport and ItemExportView

The Idea is to have a single View function in the class hierarchy which
is coming from Abstract*ExportView.

The AbstractExportView requires the following additional methods:
- `get_base_filename()`
- `get_header()`
- `export_rows()`

The `AbstractExportView` may be used together with `BaseExport` which
adds support for `VirtualFieldMixin`s and provides the `get_header()`
and `export_rows()` methods. 
`BaseExport` requires the following additional method:
- `get_object_list()`

To define the actual  export `VirtualFieldMixin`s have to be used. Note that the order in which they are defines in the export class view determines the order in the exported xlsx.
For example `class MyExport(ItemExportWithReferenceNumberMixin, ItemExportWithLinkMixin, BaseExport, AbstractXlsxExportView)` ensures that the reference number is the first column and the link the second.

All the functionality from the previous export views (`SimpleItemExportView` and `ItemExportView`) has been moved to mixins. Notable the `ExportModelFieldsMixin`.

The `BaseExport` class is not always used together with
AbstractXlsxExportView as may be seen in meinberlin in
`documents/exports.py`.
It may be discussed if the `get_field_data()` method should fall back to
`object.FOO` as this is actually behaviour from the
`MultipleObjectExport` class.

Finally the `BaseItemExportView` is connecting the AbstractView with the
`MultipleObjectMixin` and provides a `get_base_filename()`. It still
requires to be used together with `VirtualMixins` for the concrete
exports. There is a default implementation for `get_description_data`
which may be used if the concrete export is setting the `fields` to
contain `description`.
But i think this class should be moved back to meinBerlin again, as it
is very specific.


adhocracy4/exports/views.py
```diff
@@ -48,32 +47,33 @@ def _clean_field(self, field):
         return field
 
 
-class SimpleItemExportView(AbstractXlsxExportView,
-                           VirtualFieldMixin):
+class BaseExport(VirtualFieldMixin):
```
kleingeist: after relying on the not implemented `get_object_list()` method this should probably be called `AbstractBaseExport` or `AbstractObjectListExport`

adhocracy4/exports/views.py
```diff
@@ -128,26 +115,34 @@ def _setup_fields(self):
...
-        return header, names
+        return names, header
 
-    def get_base_filename(self):
-        return '%s_%s' % (self.project.slug,
-                          timezone.now().strftime('%Y%m%dT%H%M%S'))
+    def get_object_list(self):
+        return self.get_queryset().all()
+
+    def get_link_data(self, item):
```
kleingeist: i am not sure on which level in the class hierarchy the link should be generated.

adhocracy4/exports/views.py
```diff
@@ -128,26 +115,34 @@ def _setup_fields(self):
...
+        return self.request.build_absolute_uri(item.get_absolute_url())
+
+
+class BaseItemExportView(MultipleObjectExport,
```
kleingeist: i think this should be moved to meinBerlin

PR #239, opened 2018-01-30
Dashboard 2.0

kleingeist:
The dashboard from meinBerlin

Note that all the templates are still using the classes from meinBerlin. We should decide on howto handle css/html in core

missing
- [x] tests 
- [x] solution for pagination and filters
- [ ] generic solution for dynamic formsets (not sure if needed now)
- [x] allow to replace existing component (e.g. AreaSettings)


kleingeist:
yeah i could increase coverage by 6,5%
but there is still much to do until we are back to 90%

kleingeist:
yeah. back to 90%
there is still code that is not covered (for example in mixins.py) but i think that is okay, as it is mostly view related.

kleingeist:
@MagdaN i finished removing the classes from the PR and added a replace function
the only open task is a question about the categories `inline_form.html`, `module_categories_form.html` and `formset.js` which is currently still using meinBerlins class system and which is not as general as it looks, as formset.js includes strings only relevant for the categories.

* should i rename `formset.js` to `categories-formst.js`?
* or should i try to generalize it?
* is it okay to keep the classes?

kleingeist:
oh and one more thing: 
the `form_fields` are using `.form-hint` for the description of a field, while bootsrtrap4 is using `.form-text`.
* do we have to change it?

MagdaN:
@kleingeist, thanks for all the great work! 

Yes, for completion I think I would rename `.form-hint` to `.form-text` (but this is not super important). About the categories am I am not sure. Could we use bootstraps [inline forms](https://getbootstrap.com/docs/4.0/components/forms/#inline-forms) for that? Are we using django formsets with javascript somewhere else also? If yes I would be in favour to try to make the javascript more general, but this can bei done in another PR, I would say.

kleingeist:
@MagdaN i checked back with the bootstrap docs for `.form-text` and i think i understand why in mB `.form-hint` is used. the bs4 docs state:

> Help text **below** inputs can be styled with .form-text. This class includes display: block and adds some top margin for easy spacing from the inputs above.

as we are using the help texts above the input the default bootstrap behaviour would not help us at all.
if it is okay with you i'd propose to keep `.form-hint` for now

PR #240, opened 2018-01-31
Update dependencies and env path

kleingeist:
update the core js and py dependencies to the latest versions
also changes the default env path from `.` to `env`


fuzzylogic2000:
@MagdaN Did you try that? Does it work with your setup? Can we merge?

MagdaN:
@fuzzylogic2000 I have not tested it, I will merge as soon as I have done that

package.json
```diff
@@ -14,25 +14,27 @@
...
-    "jquery": "^3.2.1",
-    "js-cookie": "^2.1.2",
-    "react": "^16.2.0",
-    "react-dom": "^16.2.0",
-    "immutability-helper": "^2.3.0",
+    "immutability-helper": "^2.6.4",
+    "jquery": "^3.3.1",
+    "js-cookie": "^2.2.0",
+    "leaflet": "^1.3.1",
+    "leaflet-draw": "^0.4.14",
```
kleingeist: the merge did not work:
leaflet and leaflet-draw are now in the list twice

PR #241, opened 2018-02-08
export bools as bools

fuzzylogic2000:


kleingeist:
PYTHON!!!!! WHY!
but okay :) will merge

PR #243, opened 2018-02-20
Ka 2018 02 add icon field to category

fuzzylogic2000:
The icons have to be changed to svgs in all projects!

still missing / needing info:

- [x] icon size on map (where is the point?)
- [ ] styling in sub projects
- [ ] default pin stuff in all sub-projects with maps
- [x] aria-label or something like that to explain category icons on category form
- [x] (re)name css-classes


fuzzylogic2000:
The styling in meinberlin is done.
Can we do the styling in and the swap of the default pins, after we do the big dashboard merge?

PR #245, opened 2018-03-01
change map pin handling

fuzzylogic2000:
- the category pins should also be shown on detail pages
- too many pins need to be clustered


kleingeist:
not sure if my changes are meaningful. works in theory but is kind of complicated. django is stupid btw :(

and i managed to generate a circular reference. will probably have to do some renaming and move `*FormFields` to a separate file

PR #248, opened 2018-03-21
Add Offlineevents to core

kleingeist:
missing
- [x] tests
- [x] changelog
- [x] subproject adaptions

Changes required in subprojects:
- Use ConfiguredFileField
- Remove custom file delete signals
- Add FILE_ALIASES to config
  - requires offlineevents alias (may be empty but has to be present)
- Adapt IMAGE_ALIASES: fileformats uses tuple format now
- Adapt css: `.clear-image => .clear-file`

**migration** is done via the `migrate2a4` command in the subprojects

meinBerlin and product don't support document uploads yet (not shown in templates) as a concept/layout... is missing


fuzzylogic2000:
Hmm, still not sure if we need offline events or phases. Will close for now.

PR #261, opened 2018-05-14
cached properties

ihucos:
Ha! It does make a difference:
When viewing 4 projects in meinberlin there are 39 queries without this, with this changes it's "only" 30 queries.
As you can see in the changes for the test there is a slight change in behavior, that is that the value of a property may not change anymore when called multiple times, but I don't think we rely on that in the code.

### DIFF
--- a/adhocracy4/modules/models.py
+++ b/adhocracy4/modules/models.py
@@ -1,6 +1,7 @@
 from autoslug import AutoSlugField
 from django.db import models
 from django.core.urlresolvers import reverse
+from django.utils.functional import cached_property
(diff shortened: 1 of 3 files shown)


ihucos:
Viewing the queries there actually is more room for improvements, we do a lot of queries which only differs in milliseconds in a time that is passed to the query. I'd be happy improve the page load speed more.

ihucos:
autch, I was too optimistically about the query saving on that single page that I did the test. I have to reload two times and compare the result of the second result for some reason, so we are in fact with this on that page I described only saving one query...

PR #267, opened 2018-06-11
Move polls to a4

ihucos:


MagdaN:
@ihucos could you please add some tests here?

PR #269, opened 2018-06-20
update fontawesome - use new classnames

fuzzylogic2000:
There is already a PR (https://github.com/liqd/a4-meinberlin/pull/1243) in meinBerlin using the new fontawesome. But before merging, we need to also update the other projects.

### DIFF
--- a/adhocracy4/dashboard/templates/a4dashboard/project_list.html
+++ b/adhocracy4/dashboard/templates/a4dashboard/project_list.html
@@ -40,7 +40,7 @@
 
                     <a href="{% url 'a4dashboard:project-edit' project_slug=project.slug %}"
                        class="btn btn--light btn--small">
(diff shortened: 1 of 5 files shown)


fuzzylogic2000:
Polls in meinBerlin? They should work.
OPIN? Don't know, you should be able to see it in the settings

phillimorland:
Yeah OPIN polls, what am I looking for in settings? the flashpolls are in base.py but thats its as far as poll stuff goes

fuzzylogic2000:
Yes! When these (https://github.com/liqd/a4-opin/blob/64e6341b99ce75cdf77e2e3b5224464ef72a8ab1/euth_wagtail/settings/base.py#L84) are not yet in there it's because they are in the master, but not yet in that branch

phillimorland:
ok, I didn't find any other issues for opin and mein, still setting up advocate to test

PR #270, opened 2018-06-20
addition of class names and container div for poll styling

phillimorland:
The paragraph tag was used for styling the following element so I decided to just give it a class name rather than adding more divs to both elements.
Added container div around container for positioning purposes.
Also just realised that where I added Answer label will probably need additional stuff which I forgot

### DIFF
--- a/adhocracy4/polls/static/polls/QuestionForm.jsx
+++ b/adhocracy4/polls/static/polls/QuestionForm.jsx
@@ -7,91 +7,96 @@ var ErrorList = require('../../../static/ErrorList')
 const QuestionForm = (props) => {
   return (
     <section className="questionform">
-      <div className="questionform__content questionform__content--border">
-        <div className="form-group">
-          <label
-            htmlFor={'id_questions-' + props.id + '-name'}>
-            {django.gettext('Question')}
-          </label>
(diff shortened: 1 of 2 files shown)


adhocracy4/polls/static/polls/QuestionForm.jsx
```diff
@@ -7,91 +7,96 @@ var ErrorList = require('../../../static/ErrorList')
...
-            {django.gettext('Question')}
-          </label>
-          <textarea
-            id={'id_questions-' + props.id + '-name'}
-            name={'questions-' + props.id + '-name'}
-            value={props.question.label}
-            onChange={(e) => { props.onLabelChange(e.target.value) }} />
-          <ErrorList errors={props.errors} field="label" />
-        </div>
+      <div className="questionform__section">
```
fuzzylogic2000: Instead of adding classes directly to the elemnts you needed to style, you added one to use it as "outer" class? That's fine, but could you also just use the class questionform for that?

phillimorland: I checked and we don't need it, I think it originally didnt work because of broken link with adhocracy but all good with section class now

PR #303, opened 2018-12-27
link to individual comments

treba123:
An experimental solution for https://taiga.liqd.net/project/ggr-spd/us/47

Todo:
 - partly pull in react code from https://github.com/liqd/a4-spd/pull/531 to show a share button and visual highlighting

### DIFF
--- a/adhocracy4/comments/models.py
+++ b/adhocracy4/comments/models.py
@@ -64,9 +64,11 @@ class Comment(base.UserGeneratedContentModel):
 
     def get_absolute_url(self):
         if hasattr(self.content_object, 'get_absolute_url'):
-            return self.content_object.get_absolute_url()
+            return "{}#comment_{}".format(
+                self.content_object.get_absolute_url(), str(self.id))
         else:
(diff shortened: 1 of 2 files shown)


adhocracy4/reports/templates/a4reports/emails/report_moderators.en.email
```diff
@@ -7,6 +7,7 @@
 
 {% block content %}
 {{ report.creator.username }} requested moderation of a {{ report.content_object | verbose_name }} in your project {{ report.project }} on {{ site.name }}. The reason given is "{{ report.description }}".
+<a href={{ site.url }}{{ report.content_object.get_absolute_url }}>Click here</a> to see the {{ report.content_object | verbose_name }}.
```
fuzzylogic2000: The link is already in the cta_url block down below! That should be sufficient (and also work for all report-cases?!).

treba123: Right, didn't see that. Will update

PR #305, opened 2019-01-07
add progress of active phase as property to project

fuzzylogic2000:
As the days left are also a property and we might need the percentage elsewhere, it made sense (to me) to add it here.
But we also need something like https://github.com/liqd/a4-opin/blob/master/euth/projects/templatetags/time_delta_tags.py in meinBerlin. I am unsure where or how to add that, though. we have the days_left in the properties, but these are very coarse (because hours, minutes and seconds of timezone.now are set to zero). In meinBerlin we want days, hours and minutes and need that in the storefront and the participation overview. I don't know how to add the whole thing into the properties, because we might need to do translations with sentences. We could only add seconds_left and add the other stuff in the respective project, but that also doesn't seem too sensible. ???!!

### DIFF
--- a/adhocracy4/projects/models.py
+++ b/adhocracy4/projects/models.py
@@ -266,6 +266,21 @@ class Project(ProjectContactDetailMixin,
             return time_delta.days
         return None
 
+    @property
+    def active_phase_progress(self):
+        """
+        Return the progress of the currently active phase in percent.
+
+        Attention: This method is _deprecated_ as multiple phases may be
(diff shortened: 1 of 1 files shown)


fuzzylogic2000:
@MagdaN Ok, I will just merge this to use it right away. For the time left, we still need to do something else.

fuzzylogic2000:
@MagdaN here! How do we do that? Or where do we put it?

PR #326, opened 2019-03-18
add test organisation and group factory

MagdaN:


adhocracy4/test/factories/__init__.py
```diff
@@ -3,15 +3,22 @@
 
 from dateutil.parser import parse
 from django.conf import settings
-from django.contrib.auth.models import User
+from django.contrib.auth.models import User, Group
 
 from adhocracy4.administrative_districts.models import AdministrativeDistrict
 from adhocracy4.projects.models import Project
-from adhocracy4.organisations.models import Organisation
+from tests.apps.organisations.models import Organisation
```
fuzzylogic2000: Is the linter fine with that? I think, we should check its settings then...

PR #331, opened 2019-04-04
Update to Django 2.2 lts

rmader:
- update django and other packages
 - some trivial changes to make that work
 - remove 'limit_choices_to' from models. This requires an architectural change to our setup and needs to get replaced with some new solution. On the other hand it's not clear to me if the previous solution even worked (https://code.djangoproject.com/ticket/24195)

Note: AE currently uses a previous version of this PR which included migrations (now it doesn't). We need to fix that in the ae databases.

### DIFF
--- a/adhocracy4/api/routers.py
+++ b/adhocracy4/api/routers.py
@@ -16,8 +16,8 @@ class CustomRouterMixin():
 class ContentTypeRouterMixin(CustomRouterMixin):
 
     prefix_regex = (
-        'contenttypes/(?P<content_type>[\d]+)/'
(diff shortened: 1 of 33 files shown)


MagdaN:
@treba123 should we wait to merge this until we have additional PR in all projects?

treba123:
Not sure...if we need an updated A4 for another project we can still make a new branch without this...I'd say lets merge it.

rmader:
This is ready for review/merge. Most of it should be uncontroversial, only the last commit either needs a new solution or uncovered that our old one never worked at all.

PR #332, opened 2019-04-08
add token everywhere where maps are being used

MagdaN:


fuzzylogic2000:
Yes! It works! (In mB without token!) So, I will also test it in opin and bet.in without token? And how do we test it with token?

PR #345, opened 2019-05-08
add is_moderator_marked to comments

fuzzylogic2000:


adhocracy4/comments/models.py
```diff
@@ -45,6 +45,10 @@ class Comment(base.UserGeneratedContentModel):
         null=True,
         editable=False
     )
+    is_recommended = models.BooleanField(
```
treba123: wrong wording ;)

fuzzylogic2000: But that's why WIP!

treba123: ah alright, I thought wenn you added me to reviewers you wanted me to actually look into it :)

adhocracy4/comments/api.py
```diff
@@ -52,3 +53,29 @@ def destroy(self, request, *args, **kwargs):
...
+
+class CommentModerateSet(mixins.CreateModelMixin,
+                         mixins.RetrieveModelMixin,
+                         mixins.UpdateModelMixin,
+                         ContentTypeMixin,
+                         viewsets.GenericViewSet):
+
+    queryset = Comment.objects.all().order_by('-created')
+    serializer_class = CommentModerateSerializer
+    # permission_classes = (ViewSetRulesPermission,)
```
phillimorland: do we not need this?

fuzzylogic2000: Yes! Or not... You found the bit where there is still work to do! I need to make it work with the rules to be more specific. Currently everyone is allowed to edit the is_moderator_marked via the api.

PR #359, opened 2019-07-15
projects: add mixins to cluster overlapping modules

fuzzylogic2000:


adhocracy4/projects/mixins.py
```diff
@@ -129,3 +135,173 @@ def get_context_data(self, **kwargs):
...
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['url_name'] = self.url_name
+        context['extends'] = self.extends
+        if self.url_name == 'module-detail':
+            cluster, idx = self.other_modules
+            next_module = None
+            previous_module = None
+            try:
```
MagdaN: that is a bit hacky, but I will clean everything up in a separate PR with tests

PR #360, opened 2019-07-18
WIP: maps: Assume leaflet webpack entry point

rmader:
Required for https://github.com/liqd/a4-product/pull/798


fuzzylogic2000:
@rmader Is there anything in this PR to be used for the current rework or can we close it?

rmader:
This was actually a super bad idea. Instead we should just get our webpack config into shape so we support slice sharing and tree shaking.

PR #392, opened 2019-09-10
WIP follow button: adding follow btn component

phillimorland:


adhocracy4/follows/static/follow_dropdown/FollowDropdown.jsx
```diff
@@ -0,0 +1,87 @@
...
+          })
+        }
+      })
+  }
+  render () {
+    // FIXME: string extraction seems to be broken with JSX
+    var notFollowingLabel = django.gettext('Not following')
+    var notFollowingHelp = django.gettext('Never be notified.')
+    var followingLabel = django.gettext('Following')
+    var followingHelp = django.gettext('Receive email notifications about this project.')
```
rmader: He this is funny :) this is actually the solution to the translation problems from yesterday

phillimorland: Ahh just adding the strings as variables?

rmader: Yeah...I'm still checking how other projects handle this though :/
The good thing about this solution is, apart from being tiresome, it's quite future proof and doesn't rely on an ever more complex npm setup. So maybe we should start writing our jsx files like this by default.

phillimorland: cunning! and good to know i'll add the button wording like that in the new follow component

PR #398, opened 2019-09-17
Pm 2019 09 trans fix 2

phillimorland:
This seems to fix the last ones


rmader:
Even without fixing translation this looks much better than before

PR #409, opened 2019-10-28
WIP: make modules publishable and fix project.has_finished

fuzzylogic2000:
For the new dashboard for multi modules, the modules have to be publishable independently of the project.
Projects without modules shouldn't have a has_finished property.


fuzzylogic2000:
Ok, with the modules being draft, all the properties should only include published modules. And the current default is True... I think the default for is_draft should be False and wherever the module.is_draft is actually used, it should be made True, when it's created. @MagdaN right?

PR #415, opened 2019-10-30
ci: Use postgresql instead of sqlite

rmader:
So we test on the same db we run on our production systems.


rmader:
We do not only have type handling differences (`smallint out of range`) as expected, but also some real test failures.

```
>           assert action_count == 1
E           assert 0 == 1
```

So this might unveil some real bugs.

PR #457, opened 2020-01-08
Mn 2020 01 comment improvements

MagdaN:


MagdaN:
@phillimorland ups, there was something wrong with my branch, I rebased and removed the first commit

PR #481, opened 2020-02-13
comments_w_cat: rm recommended/moderator_marked from filter option, m…

phillimorland:
…odal, rules, serializer, api and styling and test


fuzzylogic2000:
I fear, we can only remove it from the react as we need the model field and api in spd?! (Sure about the model, unsure about the api...)

phillimorland:
@fuzzylogic2000 balls

PR #504, opened 2020-03-12
making comments with categories like child comments

phillimorland:
@fuzzylogic2000 ok I think it's done, hopefully alongside this it'll all work and once dependencies are updated it'll pass the tests https://github.com/liqd/adhocracy-plus/pull/512


adhocracy4/comments_async/static/comments_with_categories/comment.jsx
```diff
@@ -102,6 +103,10 @@ export default class Comment extends React.Component {
     return this.props.content_type !== this.context.comments_contenttype
   }
 
+  commentCategoryChoices () {
+    return this.props.content_type !== this.context.comments_contenttype
+  }
```
fuzzylogic2000: Why do we need this again? Could we just use displayCategories (which is the same) here or will this return the actual categories?

adhocracy4/comments_async/static/comments_with_categories/comment_list.jsx
```diff
@@ -23,7 +23,6 @@ export default function (props) {
               content_type={comment.content_type}
               object_pk={comment.object_pk}
               is_deleted={comment.is_deleted}
-              is_moderator_marked={comment.is_moderator_marked}
```
fuzzylogic2000: We are still using this in spd - did we also remove it elsewhere in a4? Then this can go, otherwise, I think, this might need to stay. :(

adhocracy4/comments_async/static/comments_with_categories/comment.jsx
```diff
@@ -104,7 +104,9 @@ export default class Comment extends React.Component {
   }
 
   commentCategoryChoices () {
-    return this.props.content_type !== this.context.comments_contenttype
+    if (this.props.moduleType === true) {
+      return this.props.commentCategoryChoices
+    }
```
fuzzylogic2000: Yes, right! Cool!

adhocracy4/comments_async/templatetags/react_comments_async_with_categories.py
```diff
@@ -14,7 +14,7 @@
 
 
 @register.simple_tag(takes_context=True)
-def react_comments_async_with_categories(context, obj):
+def react_comments_async_with_categories(context, obj, debate_app=False):
```
fuzzylogic2000: Yes, that's perfect! You can already rename it to react_comments_async, because where we currently use it, we need to change it to say "True" in the end anyway.

fuzzylogic2000: Actually, it should work and then not break anything, if you rename this and then make a new one like this:
`
@register.simple_tag(takes_context=True)
def react_comments_async_with_categories(context, obj):
    react_comments_async(context, obj, debate_app=True)
`

adhocracy4/comments_async/static/comments_with_categories/comment.jsx
phillimorland: This might be one of the things I need to tidy, there were alot of failed attempts, this might be one of them, I will have a look over it all

fuzzylogic2000: You already fixed it in your next commit. :)

adhocracy4/comments_async/static/comments_with_categories/comment.jsx
fuzzylogic2000: Here!

adhocracy4/comments_async/templatetags/react_comments_async_with_categories.py
```diff
@@ -14,7 +14,7 @@
 
 
 @register.simple_tag(takes_context=True)
-def react_comments_async_with_categories(context, obj):
+def react_comments_async(context, obj, debate_app=True):
     request = context['request']
     user = request.user
```
phillimorland: @fuzzylogic2000 is that what you meant?

fuzzylogic2000: Yeah, but you can also add the other bit where you use react_comments_async in react_comments_async_with_categories, so that that still works on a+ without any further  changes

adhocracy4/comments_async/templatetags/react_comments_async.py
```diff
@@ -27,16 +28,24 @@ def react_comments_async(context, obj):
...
                                )
 
-    comments_contenttype = ContentType.objects.get_for_model(Comment)
-    pk = obj.pk
+    comment_category_choices = getattr(settings, 'A4_COMMENT_CATEGORIES', None)
+    if comment_category_choices:
+        comment_category_choices = dict(
+            (x, str(y)) for x, y in comment_category_choices)
+    else:
+        raise ImproperlyConfigured('set A4_COMMENT_CATEGORIES in settings')
```
fuzzylogic2000: Nice!

adhocracy4/comments_async/templatetags/react_comments_async.py
```diff
@@ -27,16 +28,24 @@ def react_comments_async(context, obj):
...
```
fuzzylogic2000, line 48: Some projects might only use the comments without categories. With the error they also have to set them to do that. Öhm, maybe we can put in some default categories instead? 
@rmader Or do you have a better idea?

fuzzylogic2000: The ones from a+ seem to be good default categories. Should we use them?
We need to be pretty careful if we already use debate comments and change the categories later, though. So, maybe not such a great idea after all.
Oh, we can only ask, set and raise when with_categories is True :woman_facepalming: 
I will put that in!

phillimorland: I'm going to try and mirror to see if I understand :p
So if no comment_category_choices it will trigger the ImproperlyConfigured error. 
You suggest to solve this we have a default setting for categories so that the error is not triggered but this will cause issues when we have changeable categories.
Did I get that right?
Can we not just add another argument to the ifstatment with if categories = true and comment_category_choices: blah blah, or have I really got the wrong end of the stick?

adhocracy4/comments_async/templatetags/react_comments_async.py
```diff
@@ -33,19 +33,20 @@ def react_comments_async(context, obj, debate_app=True):
...
 
+    with_categories = bool(with_categories)
+
     comment_category_choices = getattr(settings, 'A4_COMMENT_CATEGORIES', None)
-    if comment_category_choices:
-        comment_category_choices = dict(
-            (x, str(y)) for x, y in comment_category_choices)
-    else:
-        raise ImproperlyConfigured('set A4_COMMENT_CATEGORIES in settings')
+    if with_categories:
```
fuzzylogic2000, line 44: This solved it! (Very straightforward...)

PR #550, opened 2020-05-20
assets: Register jQuery plugins to window.jQuery

rmader:
The global $ is now usually provided by Webpack - unfortunately that
does not play well with registering jQuery plugins, as the global
jQuery object is not shared between endpoints (or read-only?).

Closes https://github.com/liqd/a4-meinberlin/issues/2935


rmader:
ACKed on mattermost

PR #563, opened 2020-07-15
docs/datamodel: english tweeks, strike throughs and adding small updates

phillimorland:
@fuzzylogic2000 and this is a4 stuff, I have undone most of my changes about weight stuff and just added little sentance about modules

### DIFF
--- a/docs/phases_and_modules.md
+++ b/docs/phases_and_modules.md
@@ -33,7 +33,7 @@ way with the participation module's content.
 ### Current Implementation
 
 -   abstract concept "phase" is split into "phase type" and "phase"
--   phase type is called phase content or phase identifier in one case
+-   phase type is sometimes called phase content or phase identifier
 -   phase type (or content) defines:
     -   permissions
     -   view to use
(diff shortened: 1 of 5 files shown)


docs/phases_and_modules.md
```diff
@@ -45,23 +45,27 @@ way with the participation module's content.
...
+        -   Attention: _deprecated_ as this only works for one module where the
+            phase order is determined by weight.
 -   no checking of start and end dates between phases (if overlap)
--   view of phase displayed from start date to start date of next phase / or indefinitely
+-   view of phase displayed from start date to start date of next phase / or
+    indefinitely?
+-   view of phase displayed from start date to end date
 -   permissions are only set for participants
-    -   moderators and initiators can always do everything
+    -   ~~moderators and~~ initiators can always do everything?
```
fuzzylogic2000, line 56: Yes, on the item level this is right. Only on project level moderators cannot do anything

fuzzylogic2000:
Alright, I started reading and trying to clarify stuff, but my brain can't quite get around the whole thing yet. 
Thoughts:
There are currently two parts in this documentation, and we should work on it split up in these. 
1.
Here and for now we should concentrate on the underlying data models, so work on [datamodel](https://github.com/liqd/adhocracy4/blob/master/docs/datamodel.md) and [phases and modules](https://github.com/liqd/adhocracy4/blob/master/docs/phases_and_modules.md).
- we should also add something describing the dashboard and maybe the poll-app
- should and can we explain a bit better how the rules and the phases work together?
- are there other apps that should be explained?
- can we give an example of a module typically using a4, like idea collection in a+?
2.
- development, releases, guidelines and testing need to be revised (somewhere else/at a later time)
- we use development, but could probably enhance that to explain how to use a4 and give a+ as example
- releases can be either deleted or has to be updated
- guidelines can be deleted
- we could add something like contributing (but maybe explaining stuff better in development is enough)
- testing can be used, thought about, altered?!

fuzzylogic2000:
@rmader Alright! That's it for now! Could you review?

fuzzylogic2000:
@rmader Good that the pre-review reminded me. :) You wanted to keep the stuff from https://github.com/liqd/adhocracy4/pull/563/commits/42ae37db82ac0c08c16ee78d0d2649dafdeda2c9 ? Where should it go? Or we can just leave it in for now, but I am hesitant, because then it will just stay like that forever and we are not doing the stuff like described there.

PR #576, opened 2020-08-20
Ka 2020 08 more docs

fuzzylogic2000:
There is certainly more that should be said, but this is a start...


Rineee (approved):
Cool, thank you for documenting!

PR #585, opened 2020-09-14
apps/dashboard: validate that phases of the same module don't overlap

fuzzylogic2000:
That took a while! But is done now. And of course I found an error when I added the tests... But all should be fine now. You should be able to test with any project.

### DIFF
--- a/adhocracy4/dashboard/forms.py
+++ b/adhocracy4/dashboard/forms.py
@@ -159,10 +159,30 @@ class PhaseForm(forms.ModelForm):
         }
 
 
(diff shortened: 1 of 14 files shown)


phillimorland (commented):
I think it works but my tests are failing anyway

Rineee (changes requested):
Cool, the not opverlapping phase validation works! But now the validation start_date < end_date doesn't work:
`

Exception Type: | KeyError
-- | --
'end_date'

`

Rineee (approved):

Rineee:
Superawesomecool!!

PR #598, opened 2020-09-29
apps/modules/predicates: make sure project group member is also prjec…

fuzzylogic2000:
…t admin

fixes (after update) https://github.com/liqd/a4-meinberlin/issues/3155


Rineee (approved):
Yay!

PR #601, opened 2020-10-06
polls/validators: make poll validation work by comparing question pks…

Rineee:
… as integers, not objects

We were comparing 2 integers with 'is not' which acually tests object identity and only worked bc python keeps an array of integer objects for all integers between -5 and 256 and thus objects are the same for this range. But once question pk > 256, we get a validaton error bc the objects are not the same. 
See https://stackoverflow.com/questions/306313/is-operator-behaves-unexpectedly-with-integers#306353

Best Bug Ever!!! :)

### DIFF
--- a/adhocracy4/polls/validators.py
+++ b/adhocracy4/polls/validators.py
@@ -41,7 +41,7 @@ def single_vote_per_user(user, choice, pk=None):
 
 
 def choice_belongs_to_question(choice, question_pk):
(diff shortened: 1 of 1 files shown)


fuzzylogic2000:
@MagdaN I think you will also appreciate our catch of the day! :)

fuzzylogic2000 (approved):
I love it!

PR #608, opened 2020-10-21
Rm 2020 10 permissions

rmader:


rmader:
This looks very solid to me - @Rineee if you are ok I'd do a small commit cleanup and then this should be ready to get merged, right?

phillimorland:
works perfectly with https://github.com/liqd/adhocracy-plus/pull/862

PR #615, opened 2020-10-30
comments_async: new message when user is not participant of project, …

Rineee:
…but phase active

Oh my god, this really drove me crazy and it looks so simple now... 
@fuzzylogic2000 Where did you want me to comment? In the templatetag?


fuzzylogic2000 (approved):
Very nice! Works like a charm!

fuzzylogic2000:
And yes, in the templatetag would be nice! Should we still merge first and you do that later or do you want to do it in this PR?

Rineee:
> And yes, in the templatetag would be nice! Should we still merge first and you do that later or do you want to do it in this PR?

I'll do it now!

PR #618, opened 2020-11-03
projects/migrations: add backward function to 0029_is_public_to_access

Rineee:


fuzzylogic2000 (approved):
Very nice!

rmader:
Hm I guess I should have written this in the first place, just for good style? :)

PR #619, opened 2020-11-03
Ks 2020 11 alter public to access migrations

Rineee:
Changed order of acces enum values and adjusted migrations.


fuzzylogic2000 (approved):
Totally works! And tests are fine locally!

PR #626, opened 2020-11-17
comments_async: Hide report button for non-authenticated users

### LINKED ISSUE #625: reporting for users not logged in does not work
**URL:** in semi-public project https://aplus-dev.liqd.net/liqd-tests/projects/testing-semi-public-projects/ but also in public project https://aplus-dev.liqd.net/liqd-tests/projects/debatte-test/
**user:** not logged in
**expected behaviour:** I can report critical comments
**behaviour:** I can open the report form but the send button does not react
**important screensize:**
**device & browser:** safari on Mac
**Comment/Question:** I think we should allow unregistered users to report, even though otherwise they can't participate. If they should not be able to, we should not show report-button. 

Screenshot?
Bildschirmfoto 2020-11-04 um 17 19 43

rmader:
Like we do for the old comments.

Closes #625

### DIFF
--- a/adhocracy4/comments_async/static/comments_async/comment.jsx
+++ b/adhocracy4/comments_async/static/comments_async/comment.jsx
@@ -367,7 +367,7 @@ export default class Comment extends React.Component {
                     ><i className="fas fa-share" /> {share}
                     </a>}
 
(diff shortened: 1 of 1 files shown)


fuzzylogic2000 (approved):
:+1:

PR #630, opened 2020-11-17
comments_async: Support markdown in comments

### LINKED ISSUE #629: Include link in comment
**URL:** https://beteiligung.zukunftsgewerkschaft.de/subjects/2020-00683/
**user:** moderator
**expected behaviour:** I can insert a link as a hyperlink in a comment
**behaviour:** I can insert a link, but it is no hyperlink and not clickable

**Comment/Question:** Is this a feature request?

rmader:
This makes us render comments similar to what e.g. github does.

See https://github.com/remarkjs/react-markdown

Closes #629

### DIFF
--- a/adhocracy4/comments_async/static/comments_async/comment.jsx
+++ b/adhocracy4/comments_async/static/comments_async/comment.jsx
@@ -1,5 +1,6 @@
 
 import React from 'react'
+import ReactMarkdown from 'react-markdown'
 import PropTypes from 'prop-types'
 import django from 'django'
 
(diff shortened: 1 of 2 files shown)


rmader:
This probably needs a go from PMs... @fredewe @hmkay @rittermo, opinions on this?

rmader:
Here's how it would look like:
image

hmkay:
I guess this is for @fredewe . Only remark I have is the usability for mobile users. Is it easy or too complicated to have these options?

fuzzylogic2000:
And shoudn't we also change the helptext if we do this?

rmader:
We do not yet have any visible helpers...IMHO most users who want to use such features already know them - or don't need them. The requested feature was being able to paste urls and getting them recognised - this happens automatically now.

fredewe:
It seems to be a nice feature and i like it. So it would be a "go" from my side.

fuzzylogic2000 (approved):
It's just too cool! But I still think, we should maybe add some helptext, so that people know how to use it! (and I merge now, because for people using it, this doesn't make a difference, I would say)

PR #636, opened 2020-11-25
static/api.js : add livequestions POST query

Rineee:
Needed for https://github.com/liqd/a4-meinberlin/pull/3268


rmader:
Hm, while this is nice, I think this doesn't necessarily belong here - see also https://github.com/liqd/a4-meinberlin/blob/master/meinberlin/apps/livequestions/templatetags/react_questions.py#L20 and https://github.com/liqd/adhocracy-plus/blob/master/apps/interactiveevents/templatetags/react_interactiveevents.py#L23

rmader:
In other words, having it in a4 means it would would not work for a+ as the path is slightly different :/

Maybe we can have a similar api.js within the projects or so?

Rineee:
Oh ok, I didnt realize that it will not work for A+ and thought it would be good to have it here... but then I'll just add it to mB! 
I here: https://github.com/liqd/a4-meinberlin/blob/master/meinberlin/apps/livequestions/assets/helpers.js#L3 ?

PR #638, opened 2020-12-03
projects/models: delete uneeded helptext fixes mb issue #3278

phillimorland:
fixes https://github.com/liqd/a4-meinberlin/issues/3278 and makes sense for the other projects

So these changes were already in the async comments but I only had to update the tests when adding to old comments, does this mean we don't test the async comments? I only found an api test but maybe thats enough? I don't quite understand to be honest @rmader @fuzzylogic2000


Rineee:
Oh yes, you are right @philli , async comments are not very well tested, templatetag isnt tested at all, so thats why I didnt get an error when adding it there.. I'll write an issue!

Rineee (approved):
Yeah!

Issue #644, opened 2020-12-14
when there are too many modules and one starts a minute later, the sorting breaks


fuzzylogic2000:
**URL:** https://meinberlin-dev.liqd.net/projekte/test-text-commenting-multimodul/
**user:** 
**expected behaviour:** Every module should be there once, 
**behaviour:** one module is missing and one is there twice
**important screensize:**
**device & browser:** 
**Comment/Question:** 
- happens in here https://github.com/liqd/adhocracy4/blob/ad91f10a04a7630e7553f0dc1d4d8681d29e4c87/adhocracy4/projects/utils.py#L5
- happens when there are multiple modules and one (here Text commenting 2) starts one minute later (00:01 instead of 00:00)

Screenshot?

PR #645, opened 2020-12-14
projects/utils: order projects also by weight to make it distinct

Rineee:
Attempt to fix https://github.com/liqd/adhocracy4/issues/644


fuzzylogic2000 (approved):
Yeah! Let's see if it works!

PR #650, opened 2020-12-22
filters: Add reproducable random case to DistinctOrderingFilter

rmader:
This allows us to have random but distinct ordering. The resulting
order will stay the same as long as the input queryset and the
`random_seed` argument stay the same. The seed AFAIK can have any
format - one example would be `datetime.date.today`.

Required for https://github.com/liqd/a4-civic-europe/pull/275


Rineee (approved):
NIce!!

PR #664, opened 2021-02-04
apps/phases: only check overlapping for existing phase dates, make su…

fuzzylogic2000:
…re no phase date or both are there

fixes https://sentry.liqd.net/sentry/ce-stage/issues/1300
and does not change the behaviour in the dashboard


Rineee (approved):
Very nice!

PR #665, opened 2021-02-05
apps/export: export empty string if value is none

fuzzylogic2000:
fixes that none values are exported as string "None" like on https://meinberlin-dev.liqd.net/dashboard/organisations/liqd/plans/


Rineee (approved):
:+1:

PR #669, opened 2021-02-10
tests/exports/test_export_view: test get_field_data of BaseExport whe…

Rineee:
…n no custom getters are defined


fuzzylogic2000 (approved):
very nice!

PR #670, opened 2021-02-10
index: Add common map functions to exports

rmader:
This way we can finally remove the alias magic in webpack from
our frontends.

Also some more cleanups and ES6 ports.


phillimorland (approved):
Super cool!

PR #671, opened 2021-02-11
maps/assets: Add leaflet attribution layer to vector maps

rmader:
Properly include the passed in attribution - this was overlooked
during the initial implementation.


phillimorland (approved):
very nice!

PR #672, opened 2021-02-11
apps/projects and modules: add time_left property for < 1 second

fuzzylogic2000:
fixes https://sentry.liqd.net/organizations/sentry/issues/1293
I didn't know how to get that error manually, so added the tests to do that.


Rineee (approved):
Really cool!

PR #673, opened 2021-02-11
static: Avoid reinitializing window.jQuery

rmader:
This code runs unconditionally when importing anything from adhocracy4
via node/ES5 `require`. Depending on order, this would make us
overwrite an existing `window.jQuery` instance - which may have been
extented by e.g. `select2`, which then broke.


fuzzylogic2000 (approved):
Very nice! I will merge it even though @phillimorland is meant to review. :)

phillimorland:
super cool!

PR #676, opened 2021-02-18
comments/assets: Use passive scroll listener

rmader:
See https://web.dev/uses-passive-event-listeners/


fuzzylogic2000 (approved):
Uh, nice!

PR #677, opened 2021-02-19
apps/comments: only show comment form for answering when commenting p…

fuzzylogic2000:
…ossible

Will fix https://github.com/liqd/a4-meinberlin/issues/3421

It would have been quite easy displaying the same text by adding the isReadOnly and isContextMember to the CommentForm also for child comments. But that didn't make lot of sense, so this only displays the comment form (or the appropriate text), when the phase allows commenting (or you're logged in as admin or initiator and so on (see isReadOnly)). Tonight my brain is not working well enough anymore to also understand what is and should be happening for semi-public projects. So, we still need to check that. And it will probably also have to be fixed on the async comments.
So, to do:

- [x] check semi-public projects
- [ ] check and (most likely) fix for async comments

### DIFF
--- a/adhocracy4/comments/static/comments/Comment.jsx
+++ b/adhocracy4/comments/static/comments/Comment.jsx
@@ -212,9 +212,9 @@ class Comment extends React.Component {
             </div>
           </div>}
 
-        {this.state.showChildComments
-          ? (
-            <div className="comment-child-list">
+        <div className="comment-child-list">
+          {this.state.showChildComments
+            ? (
               <CommentList
                 comments={this.props.child_comments}
                 parentIndex={this.props.index}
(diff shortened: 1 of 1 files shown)


adhocracy4/comments/static/comments/Comment.jsx
```diff
@@ -232,10 +235,12 @@ class Comment extends React.Component {
                 error={this.props.replyError}
                 errorMessage={this.props.errorMessage}
                 handleErrorClick={() => this.props.handleReplyErrorClick(this.props.index, this.props.parentIndex)}
+                isReadOnly={this.props.isReadOnly}
+                isContextMember={this.props.isContextMember}
```
fuzzylogic2000, line 239: Oh, we don't need these anymore here. Hmm.

fuzzylogic2000:
@Rineee @phillimorland This works (also for semi-public and private projects) and I will merge it, so that it can be tested in mB. We still have to check this out for the async comments.

PR #680, opened 2021-02-22
apps/comments: only show comment form for answering when commenting p…

### LINKED ISSUE #679: unauthenticated user shouldn't see the text to login twice
#677
Screenshot from 2021-02-22 10-29-51

fuzzylogic2000:
…ossible
fixes #679 

I also wanted to add that the whole thing (with #677) to the async comments, but as comments are currently broken on a+, I will do that later!

### DIFF
--- a/adhocracy4/comments/static/comments/Comment.jsx
+++ b/adhocracy4/comments/static/comments/Comment.jsx
@@ -212,9 +212,9 @@ class Comment extends React.Component {
             </div>
           </div>}
 
-        {this.state.showChildComments
-          ? (
-            <div className="comment-child-list">
+        <div className="comment-child-list">
+          {this.state.showChildComments
+            ? (
               <CommentList
                 comments={this.props.child_comments}
                 parentIndex={this.props.index}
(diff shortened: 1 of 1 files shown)


Rineee:
I dont get it, wasnt this merged yesterday in https://github.com/liqd/adhocracy4/pull/677 ? :exploding_head:

fuzzylogic2000:
@Rineee Yes, the one yesterday works when the phase doesn't allow commenting or the project is semi-public. Bt I only noticed later, that in an active public project with an active commenting-phase, it says to login at every comment form fiield, also below the child comments, which looks wired
Right, I didn't add the issue. It's #679

rmader:
async comments should be fixed in https://github.com/liqd/adhocracy4/pull/682

Rineee:
@fuzzylogic2000 But isnt this the very same commit as in https://github.com/liqd/adhocracy4/pull/677/files?

fuzzylogic2000:
@Rineee Ah, yes, when you look at the commit. Thanks! It was meant to be a different change, obviously. :thinking:

Rineee:
But it still confuses me why GitHub shows the changes even though they are already on master .. ?

PR #682, opened 2021-02-23
comments_async: Do not export 'renderComment' as default

rmader:
It broke comment initialization in a+:
`ReactWidget.initialise('a4', 'comment_async', ReactCommentsAsync.renderComment)`


fuzzylogic2000 (approved):
Nice! That's working!
But why does this one break it and the other exports don't?

rmader:
> But why does this one break it and the other exports don't?

Small differences in export/import syntax :/ will probably be the same for all ES6 exports, so we'll have to adapt over time.

PR #688, opened 2021-03-12
apps/projects: make sure predicate is checked against organisation al…

fuzzylogic2000:
…so when project given

fixes https://github.com/liqd/a4-meinberlin/issues/3419

### DIFF
--- a/adhocracy4/organisations/predicates.py
+++ b/adhocracy4/organisations/predicates.py
@@ -26,11 +26,15 @@ def is_org_member(user, subject):
 
 
 @rules.predicate
-def is_org_group_member(user, organisation):
-    if organisation:
-        if hasattr(organisation, 'groups'):
-            org_groups = organisation.groups.all()
-            user_groups = user.groups.all()
-            group = org_groups & user_groups
-            return group.count() > 0
+def is_org_group_member(user, subject):
+    if subject:
+        if hasattr(subject, 'has_initiator') and hasattr(subject, 'groups'):
+            organisation = subject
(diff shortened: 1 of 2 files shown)


adhocracy4/organisations/predicates.py
```diff
@@ -26,9 +26,17 @@ def is_org_member(user, subject):
...
-def is_org_group_member(user, organisation):
-    if organisation:
-        if hasattr(organisation, 'groups'):
+def is_org_group_member(user, subject):
+    if subject:
+        if hasattr(subject, 'has_initiator') and hasattr(subject, 'groups'):
+            organisation = subject
+            org_groups = organisation.groups.all()
+            user_groups = user.groups.all()
+            group = org_groups & user_groups
```
Rineee, line 35: Nitpick :) : What do you think about taking these out of the if statement and combining them with lines 40-42?

Rineee (approved):
Cool, that works perfectly! I made a small comment, but if you think its unnecessary, just merge!

adhocracy4/organisations/predicates.py
fuzzylogic2000: Haha, late brain only thought about the return outside of the ifs. But that can also go in the else! Yes, good point!

fuzzylogic2000: And I didn't even need the else!

Rineee: Perfect!

Rineee (approved):

PR #696, opened 2021-03-29
settings: make isort not check migrations even when passed as argument

fuzzylogic2000:
With this, isort doesn't check the migrations, but flake8 still does, I couldn't fix that. :(

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -17,5 +17,6 @@ indent_style = tab
 # isort configuration
 [*.py]
 skip=migrations,node_modules,venv
+skip_glob = */migrations/*.py
 known_first_party=adhocracy4, tests
 force_single_line = true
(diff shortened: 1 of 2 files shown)


phillimorland:
@fuzzylogic2000 sorry I only just got to this after day of fixing and admin stuff, I think i'm having trouble replicating issue, if I create a migration and commit it, then migration does not cause linting errors despite having long lines ect, am I missing something?

fuzzylogic2000:
@phillimorland Hmm, weird! For this migration https://github.com/liqd/adhocracy4/blob/master/adhocracy4/projects/migrations/0032_increase_image_name_max_length.py it gave me an isort and the line to long error. :( But this PR only fixes the isort one anyway.

PR #705, opened 2021-04-13
Fixing bug in nested comment form

### LINKED ISSUE #678: fix comment answer form for async comments
#677

khamui:
Passing isReadOnly and isContextMember to child component `comment.jsx`.
Adding isAuthenticated flag to condition inside of `comment_form.jsx`.

fixes  #678

### DIFF
--- a/adhocracy4/comments_async/static/comments_async/comment_form.jsx
+++ b/adhocracy4/comments_async/static/comments_async/comment_form.jsx
@@ -76,8 +76,9 @@ export default class CommentForm extends React.Component {
 
   render () {
     const textareaStyle = { height: (this.state.textareaHeight) + 'px' }
+    const { isAuthenticated } = this.context
 
-    if (this.context.isAuthenticated && !this.props.isReadOnly) {
+    if (isAuthenticated && !this.props.isReadOnly) {
       return (
         <div>
           <form id="id-comment-form" className="general-form" onSubmit={this.handleSubmit.bind(this)}>
@@ -111,13 +112,13 @@ export default class CommentForm extends React.Component {
           </form>
         </div>
       )
-    } else if (!this.props.isReadOnly) {
+    } else if (!isAuthenticated && !this.props.isReadOnly) {
       return (
         <div className="a4-comments__login">
           <a href={config.getLoginUrl()}>{django.gettext('Please login to comment')}</a>
         </div>
       )
-    } else if (!this.props.isContextMember) {
+    } else if (isAuthenticated && !this.props.isContextMember) {
       return (
         <div>
           {django.gettext('Only invited users can actively participate.')}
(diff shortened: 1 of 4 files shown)


phillimorland (approved):
very nice and clean! :) and good description.

phillimorland:
@khamui oh one thing, when issue fixing add the issue reference either to commit or to ref, i'll just whack it in now.

adhocracy4/comments_async/static/comments_async/comment_form.jsx
```diff
@@ -111,13 +112,13 @@ export default class CommentForm extends React.Component {
           </form>
         </div>
       )
-    } else if (!this.props.isReadOnly) {
+    } else if (!isAuthenticated && !this.props.isReadOnly) {
```
fuzzylogic2000, line 115: We don't need !isAuthenticated here because that is clear from the if before. But it can stay to be more explicit. :)

adhocracy4/comments_async/static/comments_async/comment_form.jsx
```diff
@@ -111,13 +112,13 @@ export default class CommentForm extends React.Component {
...
       return (
         <div className="a4-comments__login">
           <a href={config.getLoginUrl()}>{django.gettext('Please login to comment')}</a>
         </div>
       )
-    } else if (!this.props.isContextMember) {
+    } else if (isAuthenticated && !this.props.isContextMember) {
```
fuzzylogic2000, line 121: Here we don't want isAuthenticated, because otherwise we would show the next option "The currently active state doesn't allow to comment" also to users that might be allowed to comment once they login. 
If we have a semi-private project for example and the currently active phase does allow invited participants to comment, we want to show this instead of the next option.

fuzzylogic2000 (commented):
Two comments for the comment_form. I will look at the further into it after MMM. :)

adhocracy4/comments_async/static/comments_async/comment_form.jsx
khamui: think i am confused by the `isContextMember` flag, I thought it is depending on the user/member, but it can't be because its set even if no user is logged in.

Based on what information, this flag is set?

khamui:
some side effects for the case described in this issue: #707

fyi @phillimorland @fuzzylogic2000

adhocracy4/comments_async/static/comments_async/comment_form.jsx
Rineee: You can find the definition here: https://github.com/liqd/adhocracy4/blob/fba25b7d8b2341de541592849ad0ba7295c25614/adhocracy4/comments_async/templatetags/react_comments_async.py#L74

The NormalUser is a dummy user defined here: https://github.com/liqd/adhocracy4/blob/master/adhocracy4/rules/discovery.py

So altogether isContextMember is true if the project is public (coming from the NormalUser part, because for public projects any authenticated user is context member) or the user is authenticated and is participant/moderator/org_member.

I hope this helped and wasnt even more confusing :grimacing:

PR #706, opened 2021-04-14
update docs to include virtual env

phillimorland:
This was unclear for people new to the project and environment and meant commiting was made more difficult as our precommit relys on virenv checks so thought would be good to add


phillimorland:
@fuzzylogic2000 this is correct right?

fuzzylogic2000 (approved):
Yes, nice!

PR #712, opened 2021-04-28
apps/questions: fixing poll buttons

khamui:
In Question.jsx a bug around 'showing preliminary results' button is fixed
Additionally following bugs to go back 'to poll' and 'change vote' buttons
are also fixed.

Fixes #2106 (a4-opin repo though)

### DIFF
--- a/adhocracy4/polls/static/polls/Question.jsx
+++ b/adhocracy4/polls/static/polls/Question.jsx
@@ -135,17 +135,21 @@ class Question extends React.Component {
   }
 
   toggleShowResultButton () {
-    let toggleShowResultButtonText
-    if (this.state.showResult) {
-      toggleShowResultButtonText = django.gettext('To poll')
-      if (this.state.selectedChoices.length !== 0) {
-        toggleShowResultButtonText = django.gettext('Change vote')
-      }
-    } else {
-      toggleShowResultButtonText = django.gettext('Show preliminary results')
-    }
+    const { showResult, selectedChoices: { length: isVoted } } = this.state
+    const buttonText = !showResult
+      ? django.gettext('Show preliminary results')
+      : isVoted
+        ? django.gettext('Change vote')
+        : django.gettext('To poll')
+
     return (
-      toggleShowResultButtonText
+      <button
+        type="button"
+        className="btn btn--link"
+        onClick={this.toggleShowResult.bind(this)}
+      >
+        {buttonText}
+      </button>
     )
   }
 


fuzzylogic2000 (changes requested):
Very nice! 
I was only wondering if there isn't more code to remove? We must have at least tried to put that into a link? 
And even though it says "button" in all the variables, it was meant to be a link like in a+: https://aplus-dev.liqd.net/liqd-tests/projects/public-poll/
Also, if you want to, you can clean up the code even more! I mean the other parts of it. (And then when it's all nice and clean, we could even port the code to a+ and mB to have it nice and clean everywhere. Or no, I will think about how to use the a4-code again.)

fuzzylogic2000:
Well, actually it may be best to clean and make it nice in a+ and then use that here as well... Let's talk about that!

adhocracy4/polls/static/polls/Question.jsx
```diff
@@ -135,17 +135,21 @@ class Question extends React.Component {
...
-      }
-    } else {
-      toggleShowResultButtonText = django.gettext('Show preliminary results')
-    }
+    const { showResult, selectedChoices: { length: isVoted } } = this.state
+    const buttonText = !showResult
+      ? django.gettext('Show preliminary results')
+      : isVoted
+        ? django.gettext('Change vote')
+        : django.gettext('To poll')
```
fuzzylogic2000, line 143: Cool! That reads very well!

fuzzylogic2000 (commented):
But it's meant to be a link. :grimacing:

khamui:
> Very nice!
> I was only wondering if there isn't more code to remove? We must have at least tried to put that into a link?
> And even though it says "button" in all the variables, it was meant to be a link like in a+: https://aplus-dev.liqd.net/liqd-tests/projects/public-poll/
> Also, if you want to, you can clean up the code even more! I mean the other parts of it. (And then when it's all nice and clean, we could even port the code to a+ and mB to have it nice and clean everywhere. Or no, I will think about how to use the a4-code again.)

I see. Technically it is a button with the appearance of a link right? Because the click event calls a method, rather than following an href. To be meaningful i would avoid to call a method in an <a> element. Rather we restyle a button.

khamui:
Styles adjustments needed for this is in the following PR (a4-opin). Currently only affecting a4-opin, since its used there only.

https://github.com/liqd/a4-opin/pull/2162

fuzzylogic2000 (approved):
Yeah!

PR #716, opened 2021-05-05
filters: move daily random to new filter class

Rineee:
Sorry, very non-creative with new class name in the end

### DIFF
--- a/adhocracy4/filters/filters.py
+++ b/adhocracy4/filters/filters.py
@@ -1,5 +1,6 @@
 import operator
 import random
+from datetime import date
 from functools import reduce
 
 import django_filters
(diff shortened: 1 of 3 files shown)


Rineee:
Ok, fixing tests first..

adhocracy4/filters/filters.py
```diff
@@ -57,40 +58,41 @@ def __init__(self, data, *args, **kwargs):
...
+
+
+class DistinctOrderingWithDailyRandomFilter(DistinctOrderingFilter):
+    """Note: order reproducability relies on the string representation of the
+    seed
+    """
 
     def filter(self, qs, value):
-        if value == ['?']:
+        if value == ['dailyrandom']:
```
fuzzylogic2000, line 84: We need to change the value in CE then, right? Maybe we keep the '?' for now to be sure not to break anything else.? (Even though I argued against the '?' when we added it in the first place. :smile: )

fuzzylogic2000 (approved):
nice! Only one thing to discuss.

adhocracy4/filters/filters.py
Rineee: I did this on purpose here, because the '?' is a default for random ordering. So without the `if value == [?]` and then all the stuff we do with the date, it would just do a random order.. this is also what it does now if another ordering filter uses '?', like before. I thought its better to distinguish here, because it is not just the default random ordering.. do you know what I mean?

fuzzylogic2000: Yes! Nice!

PR #720, opened 2021-05-11
apps/modules: add countup to starting module as property

fuzzylogic2000:
We need this for the interactive event in a4 (and might need it again elsewhere). Arrrr, and I forgot: tests!! Let's still merge and I will add an issue and tests later!


Rineee (approved):
Cool!

PR #723, opened 2021-05-11
tests/modules: test module_starting_time_left and some more prop…

### LINKED ISSUE #721: add test for module property module_starting_time_left
#720

fuzzylogic2000:
…erties - fixes #721

It also tests the module.get_detail_url now, which was only fixed last week!


Rineee (approved):
Yeah, nice!

PR #727, opened 2021-05-14
apps/dashboard: copy moderators over on project duplicate

fuzzylogic2000:
This avoids having no moderators in a project.
Fixes https://github.com/liqd/adhocracy-plus/issues/1063


Rineee (approved):
Right!

PR #737, opened 2021-06-11
Kl 2021 06 freetext poll models

fuzzylogic2000:
I tried it out in opin and it seems to work fine. At least updating the Db and cleaning the question. Do we want to display any of it in the django-admin? We currently only show the Questions.


Rineee (approved):
Nice! Lets go on from here!

PR #739, opened 2021-06-15
tests/polls: add tests for the new model fields

fuzzylogic2000:
@Rineee There will be more to add after the QS stuff, but let's merge and add that with that?!


Rineee (approved):
Yes! Nice!

PR #744, opened 2021-06-23
polls/models: add validation before save

Rineee:
Tests for new serializers are still missing, but there is a task in taiga and I wanted to wait until we are sure that this is the way to do it.. 

The serialized data can be seen by adding mixinx.RetrieveModelMixin to PollViewSet in polls/api.py . Then /api/polls/pollId shows all questions of the poll (although the vote annotations arent there, as the questions are only annotated in the VoteViewSet and not in the PollViewSet). I tested with Opin, hope it also works in mB :nerd_face:

### DIFF
--- a/adhocracy4/polls/models.py
+++ b/adhocracy4/polls/models.py
@@ -101,9 +101,31 @@ class Question(models.Model):
         answers = self.answers.filter(creator=user)
         if answers.exists():
             # there can only be one answer bc of unique constraint
-            return answers.first().answer
+            return answers.first().id
         else:
             return ''
+
+    def other_choice_answers(self):
+        if self.has_other_option:
+            other_choice = self.choices.filter(is_other_choice=True).first()
+            other_answers = OtherVote.objects.filter(vote__choice=other_choice)
+            return other_answers
+        else:
+            return OtherVote.objects.none()
+
+    def other_choice_user_answer(self, user):
+        if not user.is_authenticated:
+            return ''
+
+        elif self.has_other_option:
+            other_choice = self.choices.filter(is_other_choice=True).first()
+            other_choice_user_answer = OtherVote.objects.filter(
+                vote__creator=user,
+                vote__choice=other_choice)
+            if other_choice_user_answer.exists():
+                # there can only be one other vote as 1:1 relation
+                return other_choice_user_answer.first().vote.id
+        return ''
 
     def get_absolute_url(self):
         return self.poll.get_absolute_url()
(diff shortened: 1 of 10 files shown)


adhocracy4/polls/models.py
```diff
@@ -148,6 +163,23 @@ class Choice(models.Model):
...
+    def clean(self, *args, **kwargs):
+        if self.question.is_open:
+            raise ValidationError({
+                'question': _('Open questions cannot have choices.')
+            })
+        if self.is_other_choice and self.question.choices.count() == 0:
+            raise ValidationError({
+                'is_other_choice': _('"Other" cannot be the only choice. Use '
+                                     'open question or add another choice '
+                                     'first.')
```
fuzzylogic2000, line 175: The text sounds a bit like you are only allowed to check the box if there are questions added first, so maybe without the first?
'"Other" cannot be the only choice. Use '
'open question or add more choices.'

fuzzylogic2000 (commented):
Looks good! (But the tests are failing...) And there are some missing. :)

adhocracy4/polls/models.py
```diff
@@ -71,15 +71,21 @@ def clean(self, *args, **kwargs):
             if self.multiple_choice:
                 raise ValidationError({
                     'is_open': _('Questions with open answers cannot '
-                                 'have multiple choices')
+                                 'have multiple choices.')
                 })
-            elif self.has_other_option:
```
Rineee, line 76: This is now checked when adding a choice

Rineee:
Hm, tests are broken, because apparently I now broke the question api.. will look into this tomorrow!

Rineee:
Ok, tests are fixed, but tests for new serializers are missing, and also tests for both question and vote api should be updated to also test the open stuff.. will add a task and do that once everything is settled.

For voting, the api now expects data of the form:
`    data = {
        'choices': [choice1.pk, choice2.pk],
        'other_choice_answer': 'myOtherChoiceAnswer',
        'open_answer': 'myOpenAnswer'
    }
`

adhocracy4/polls/static/polls/PollManagement.jsx
```diff
@@ -43,6 +43,7 @@ class PollManagement extends React.Component {
   */
 
   getNewQuestion (label = '') {
+    console.log('I am so here')
```
fuzzylogic2000, line 46: This should probably also go before merge?

fuzzylogic2000 (approved):
It looks good and we can merge to get it all together and change the two things and more afterwards!

adhocracy4/polls/static/polls/PollManagement.jsx
Rineee: haha, ups :see_no_evil:

Issue #743, opened 2021-06-22
make sure the next-module thing works for the polls


fuzzylogic2000:
**URL:** https://meinberlin-dev.liqd.net/projekte/this-tests-two-polls-in-one-project/
**user:** 
**expected behaviour:** 
**behaviour:** The poll module from mB uses the get_detail_url (I think) and can thus display the next-module thing, and a back-button. Using the poll app from a4 in mB this breaks.
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?

PR #745, opened 2021-06-23
apps/polls: make sure detail view knows which view to display - fixes…

fuzzylogic2000:
… #743

depending on the timeline and the module clusters, either the project view with or without initial timeline tile or the module view need to be displayed.
This is needed  to work with the timeline/cluster logic made use of in mB.
https://github.com/liqd/a4-meinberlin/commit/a3e568f77c68dfb78845968a044545b212a5ae36
It does not break projects without that logic (opin) as the detail template is overwritten. And there are no multi modules in opin.

### DIFF
--- a/adhocracy4/polls/templates/a4polls/poll_detail.html
+++ b/adhocracy4/polls/templates/a4polls/poll_detail.html
@@ -1,9 +1,9 @@
-{% extends "a4projects/project_detail.html" %}
+{% extends extends %}
 {% load react_polls react_comments %}
 
 
 {% block phase_content %}
-    <div class="module-content">
+    <div class="module-content--light">
         <div class="l-wrapper">
             <div class="l-center-8">
                 {% for question in question_list %}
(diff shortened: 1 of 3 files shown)


adhocracy4/polls/templates/a4polls/poll_detail.html
```diff
@@ -1,9 +1,9 @@
-{% extends "a4projects/project_detail.html" %}
+{% extends extends %}
 {% load react_polls react_comments %}
 
 
 {% block phase_content %}
-    <div class="module-content">
+    <div class="module-content--light">
```
phillimorland, line 6: I don't think we use this class anwhere anymore but I don't think it's important

phillimorland (approved):
Nice

adhocracy4/polls/templates/a4polls/poll_detail.html
fuzzylogic2000: Oh, I only tried with module-content and that looked all wrong. I didn't try without the class. Maybe we can redo it when we add it to a4 with classnames we can use by extend in both projects!

PR #747, opened 2021-06-28
apps/polls: also show poll before project start

fuzzylogic2000:
to make it behave like other modules, we show the content of the poll also prior to project start.

Does not affect opin, because the content of the participation is always only shown after project start, no matter what the rule is. For other projects it does not make sense to show a 403 because the project hasn't started yet, even though it's already published.


phillimorland (approved):
LGTM

PR #748, opened 2021-06-28
polls/api: adding get to api for polls

khamui:


fuzzylogic2000 (approved):
It looks good to me, but @Rineee is the expert!

Rineee:
Looks good!

PR #749, opened 2021-06-28
polls: add js folder of new poll and update path

phillimorland:
Ok i basically replaced the a4 static folder with the mB poll assets folder and will update the mB hash once this is merged


adhocracy4/polls/assets/Question.jsx
```diff
@@ -1,13 +1,10 @@
-var React = require('react')
-var django = require('django')
-var Alert = require('../../../static/Alert')
-var update = require('immutability-helper')
+const React = require('react')
```
khamui, line 1: In my current task I am restructuring this part. Actually I just prepend a Questions (list) component to this.
The action List Item (Question.jsx) i didn't touch yet, so this one here is way more refactored and advanced.

Long story short, how can we get this together?

khamui: @phillimorland

fuzzylogic2000: I think we will merge this and then port your (@khamui) changes over and after that do all the changes in A4 right away!

phillimorland: it should be be fine to just whack it in

khamui:
sorry @phillimorland i just updated that depending branch in mB.

phillimorland:
after merge this branch a4 dep to be updated and also merged to work together https://github.com/liqd/a4-meinberlin/pull/3662

khamui:
really nice! Thanks for transplanting so well and fix all issues around.
I re-read it and looks good to me ✅, but didn't merge because i am not super sure about the dependency branch. Feel free to do it!

adhocracy4/polls/assets/PollQuestion.jsx
```diff
@@ -33,9 +30,6 @@ class Question extends React.Component {
   handleSubmit (event) {
     event.preventDefault()
 
-    const voteCountText = django.gettext('Vote counted')
-    const voteNoCountText = django.gettext('Vote has not been counted due to a server error.')
-
     if (this.state.question.isReadOnly) {
```
phillimorland, line 33: @khamui we have to have translations in react in variables otherwise they are not always picked up by `make messages` can be changed later

PR #751, opened 2021-06-29
polls: using poll.pk instead of module.pk

khamui:


adhocracy4/polls/views.py
```diff
@@ -41,6 +41,7 @@ def get_context_data(self, **kwargs):
         if self.object is not None:
             context['question_list'] = \
                 self.object.questions.annotate_vote_count().all()
```
Rineee, line 43: We could then also remove question_list from the context as it is not used anymore..

phillimorland:
will close as I added you commit to the adding polls branch

PR #752, opened 2021-06-30
polls/serializers: annotate questions in poll serializer

Rineee:
After all, it was much easier than I tried before..


fuzzylogic2000 (approved):
Perfect!

PR #753, opened 2021-06-30
add github actions replacement for travis

goapunk:
Add a django workflow for building via github actions and fancy badge.


fuzzylogic2000 (approved):
:tada:

PR #756, opened 2021-06-30
polls: basic single question component with other text input choice a…

phillimorland:
…nd answers passed to state

### DIFF
--- a/adhocracy4/polls/assets/PollQuestion.jsx
+++ b/adhocracy4/polls/assets/PollQuestion.jsx
@@ -1,124 +1,72 @@
-const React = require('react')
-const django = require('django')
-const update = require('immutability-helper')
+import React from 'react'
+import django from 'django'
 
-export default class PollQuestion extends React.Component {
-  constructor (props) {
-    super(props)
+export const PollQuestion = (props) => {
+  const multiHelpText = props.question.multiple_choice ? <div className="poll__help-text">{django.gettext('Multiple answers are possible.')}</div> : null
 
-    const question = this.props.question
+  return (
+    <div className="poll u-border">
+      <h2>{props.question.label}</h2>
+      {multiHelpText}
+      <div className="poll__rows">
+        {
+          props.question.choices.map((choice, i) => {
+            const checked = props.question.userChoices.indexOf(choice.id) !== -1
 
-    this.state = {
-      question: question,
-      questionHelpText: props.question.help_text,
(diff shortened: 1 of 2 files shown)


phillimorland:
@khamui this is unstyled but we can maybe merge and go on as thought you might need the states? also assuming it'll need some tweeks but not sure yet what will be needed

khamui:
cool! yes styling can happen later. i was thinking about the state and probably its easier to send it onChange back to the parent component (PollQuestions.jsx), i have to provide you some props methods though. Similar to the EditPollQuestions.

So i keep the state updated on the parent and if user hits save we send that entire thing.

syntaxwise is cool. i mean you could convert it to functional components and get rid of the class notation :D

adhocracy4/polls/assets/PollQuestion.jsx
```diff
@@ -94,171 +43,82 @@ class PollQuestion extends React.Component {
...
-                        type="checkbox"
+                        className="input-group__input"
+                        type="text"
                         name="question"
-                        id={'id_choice-' + choice.id + '-multiple'}
-                        value={choice.id}
-                        checked={checked}
-                        onChange={this.handleOnMultiChange.bind(this)}
+                        id={'id_choice-' + choice.id + '-single'}
+                        onChange={this.handleTextChange.bind(this)}
```
khamui, line 112: `{() => props.khaProvidedMethod(question.id, valueOfInput)}`

phillimorland:
@khamui ok, second attempt, hope thats ok? i've decided to leave the commits as is becasue it will help me to see the steps in future when I have to convert other components.

adhocracy4/polls/assets/PollQuestion.jsx
```diff
@@ -1,124 +1,72 @@
...
+              return (
+                <label className="poll-row radio" key={choice.id} htmlFor={'id_choice-' + choice.id + '-multiple'}>
+                  <input
+                    className="poll-row__radio radio__input"
+                    type="checkbox"
+                    name="question"
+                    id={'id_choice-' + choice.id + '-multiple'}
+                    value={choice.id}
+                    checked={checked}
+                    onChange={(e) => { props.OnMultiChange(props.question.id, e.target.value).bind(this) }}
```
khamui, line 51: think you can omit `.bind(this)` and a little typo `OnMulti..` should be `onMulti..` right?

PR #758, opened 2021-07-01
Poll: add open question comp and question helptext

phillimorland:


phillimorland:
@khamui I didn't touch the list in the end as thought more of a pain for you as you're working there and also my solution it turned out wasn't actually working :p anyway thats there for you to grab

khamui:
@phillimorland yes makes sense! let's pair program to get list comp and the single comps together.

PR #759, opened 2021-07-01
apps/polls/userview: polls questions logic etc.

khamui:
this is a wip. But in order to progress, it can be merged


khamui:
okay i was working on updating the state for the voting request. But i didn't manage to use the useState with those immutable-helpers. So at the end i wrote a serializer to do the job. Not sure if this is cool, i will review on monday and test as good as i can.

Current state is still wip.

phillimorland (commented):
looking good! some expected errrors for results but will merge and go on

PR #760, opened 2021-07-02
tests/polls: move poll tests from mB

fuzzylogic2000:


tests/polls/helpers.py
```diff
@@ -0,0 +1,4 @@
+def setup_poll(poll, question, choice, vote):
+    question.poll = poll
+    choice.poll = poll
+    vote.choice = choice
```
Rineee, line 4: Do we use this anywhere? It shouldnt work, because a choice doesnt have a poll, but a question and all the objects would have to be saved after changing and be returned, wouldnt they?

fuzzylogic2000: :smile: And I was happy that it just worked. Good catch! It's removed!

Rineee:
Yess!

PR #762, opened 2021-07-05
apps/polls: add weight to choices and use to save order of choices in…

fuzzylogic2000:
… serializer

Whoop whoop! @Rineee was right! Just using this saves the choices already in their right order!


fuzzylogic2000:
second commit is just a drive-by (or whatever you call that)

khamui:
working fine on local!

PR #771, opened 2021-07-07
Ks 2021 07 error messages poll dashboard

Rineee:


adhocracy4/polls/assets/HelptextForm.jsx
```diff
@@ -16,7 +16,7 @@ export const HelptextForm = (props) => {
           onChange={(e) => { props.onHelptextChange(e.target.value) }}
         />
       </label>
-      <ErrorList errors={props.errors} field="label" />
+      <ErrorList errors={props.errors} field="help_text" />
```
fuzzylogic2000, line 19: Ah! Good catch!

fuzzylogic2000 (approved):
Very nice!

adhocracy4/polls/assets/HelptextForm.jsx
khamui: yeah good catch 🤝

PR #772, opened 2021-07-07
polls/models: move error message to label of choice so that it is see…

Rineee:
…n in dashboard

Sorry, forgot this in the other PR.. so this is to actually show the error message in django admin when trying to add a choice to an open question.. I didnt know how else to do it, because I didnt know how to add it to a field of the questions and just adding it to the question wont show it.. so now I added it to the choice :woman_shrugging:


fuzzylogic2000 (approved):
Cool!

PR #773, opened 2021-07-07
test/factories: move poll factories here

Rineee:


fuzzylogic2000 (commented):
We should also delete the ones that are in the tests folder and then also use these in a4. (Don't mind me if that was the plan anyway, just going through the PRs.)

Rineee:
> We should also delete the ones that are in the tests folder and then also use these in a4. (Don't mind me if that was the plan anyway, just going through the PRs.)

Haha, yes that was the plan anyway, just made this PR for testing it in mB, but its ready now!

Rineee:
But now everything is gone? Dont get it..

fuzzylogic2000:
@Rineee This is weird! But when looking into the code of the branch, the changes are there! So, I guess we merge and deal with it, if that doesn't go as expected?!

Rineee:
> @Rineee This is weird! But when looking into the code of the branch, the changes are there! So, I guess we merge and deal with it, if that doesn't go as expected?!

Yes, lets do that and see what happens!

adhocracy4/test/factories/polls.py
```diff
@@ -1,17 +1,20 @@
...
+
 
 class PollFactory(factory.django.DjangoModelFactory):
 
     class Meta:
         model = models.Poll
 
-    creator = factory.SubFactory(UserFactory)
-    module = factory.SubFactory(factories.ModuleFactory)
+    creator = factory.SubFactory(USER_FACTORY)
```
fuzzylogic2000, line 16: Right! And now the code changes are also back! Just merge whenever it's ready!

PR #774, opened 2021-07-08
polls/wording: adapt strings

fuzzylogic2000:
Alright! This has a ll the wording changes. I hope, this doesn't add conflicts with what you're currently doing @phillimorland and @khamui ?!


Rineee (approved):
Cool, thanks for going through all the strings!

PR #775, opened 2021-07-08
requirements/base: psycopg2 >= 2.9 is incompatible with django 2.2.x

goapunk:
psycopg 2.9.x is incompatible with django 2.2.x so pin to the latest release and make pyup ignore it.


fuzzylogic2000 (approved):
Nice! Thank you!

PR #777, opened 2021-07-08
apps/poll/editview: renaming to BEM convention

khamui:
depending on mein Berlin PR


phillimorland (approved):
hell yeah!

fuzzylogic2000 (commented):
NIce! This will also make it easier for opin!

PR #778, opened 2021-07-08
Pm 2021 07 results improvments

phillimorland:
- [ ] add indication to display users own text answer
- [ ] fix slides being weird (maybe best done in the project)


phillimorland:
@khamui so I was trying to use state for getting the answers but I feel like this is swerving quiet far from best practice so be gentle :p also you may need to look at it with my styling pull request as it might be crap if not

khamui:
nice one! slick one!

PR #780, opened 2021-07-09
polls/api: only save answer and other choice answer if not empty

Rineee:
This still doesnt catch the validation errors, but maybe we could merge this first, so Caro can test properly?


fuzzylogic2000 (approved):
Cool!

PR #781, opened 2021-07-12
apps/budgetingmap/maps: fixing inactive pins (safari & FF)

khamui:
Described in project mB (link to issue):
https://github.com/liqd/a4-meinberlin/issues/3633

A bug in leaflet, which leads to that issue in Safari (and Firefox).
https://github.com/Leaflet/Leaflet/issues/7266


phillimorland (approved):
Looks good, my only concern is this may also affect mobile but i'm happy to merge and check it on dev

PR #789, opened 2021-07-13
apps/poll/userview: refreshing after voting

khamui:


phillimorland (approved):
LGTM, testing locally is hit and miss right now so will merge and check on dev

PR #790, opened 2021-07-15
apps/polls: only show poll initially when phase active

fuzzylogic2000:
This will fix https://github.com/liqd/a4-meinberlin/issues/3736 and https://github.com/liqd/a4-meinberlin/issues/3735 once updated.


Rineee (changes requested):
Looking good! But there are apparently conflicts..

Rineee (approved):
Yeah!!

PR #791, opened 2021-07-15
apps/poll: rm bootstrap classes to ensure easier styling across multi…

phillimorland:
…ple apps and avoid bs update issues in a4


phillimorland:
I will also have to re-tweek mB but this means it's easier to use elsewhere otherwise we had to write over bs classes and when updating it#s annoying to have to check in a4 too

fuzzylogic2000 (approved):
very nice!

PR #792, opened 2021-07-15
comments_async: replace bs class names with updated ones//poll/editPo…

phillimorland:
…llQuestions: rm obsolete class names


phillimorland:
hopefully getting comments working and removing un-needed class in edit poll

fuzzylogic2000 (approved):
Right! Nice!

PR #793, opened 2021-07-19
apps/poll/userview: fix infinite loading spinner and show success alert

khamui:
if no changes/modification on votes

issue reported in OPIN:
https://github.com/liqd/a4-opin/issues/2245


phillimorland (approved):
Very nice!

PR #794, opened 2021-07-19
apps/poll/userview: fix issue with other choice validation

khamui:


khamui:
issue description in meinBerlin repo:
https://github.com/liqd/a4-meinberlin/issues/3737

Rineee:
Cool! But I think we would also need for the multiple choice questions, same error there..

adhocracy4/polls/assets/PollQuestion.jsx
```diff
@@ -63,7 +71,7 @@ export const PollQuestion = (props) => {
                       id={'id_choice-' + choice.id + '-single'}
                       value={choice.id}
                       checked={checked}
-                      onChange={(event) => { handleSingleChange(event) }}
+                      onChange={(event) => { handleSingleChange(event, choice.is_other_choice, checked) }}
```
Rineee, line 74: Do you need the checked there as an argument?

khamui: oops, yes was a leftover of something i tried before. thanks for the hint

adhocracy4/polls/assets/PollQuestion.jsx
```diff
@@ -63,8 +75,8 @@ export const PollQuestion = (props) => {
                       id={'id_choice-' + choice.id + '-single'}
                       value={choice.id}
                       checked={checked}
-                      onChange={(event) => { handleSingleChange(event) }}
-                      disabled={!props.question.authenticated || props.question.isReadOnly}
+                      onChange={(event) => { handleSingleChange(event, choice.is_other_choice) }}
+                      disabled={!props.question.authenticated | props.question.isReadOnly}
```
Rineee, line 79: Ah, this still works with the second | gone? Was this on purpose?

Rineee (approved):
Cool, now it all works! I left one small comment, its rather a question..

khamui:
@Rineee sorry my bad! actually it has nothing to do with XOR (as i said in our call). Just fixed.

PR #798, opened 2021-07-20
organisations/predicates: retrun that user is no org group member if …

Rineee:
…organisation doesnt have groups

fixes https://sentry.liqd.net/organizations/liqd/issues/1454/?project=8&query=is%3Aunresolved


phillimorland (approved):
LGTM

PR #799, opened 2021-07-21
deps: add react-popper for poll

phillimorland:


phillimorland:
works with an updated hash but doesn't work when linked in dev mode, I think becasue it doesn't do a 'full' install?

Rineee (approved):
Yes!

PR #800, opened 2021-07-22
deps: add popperjs core as dependedncy of react-popper

phillimorland:


Rineee (approved):
Yes, lets try this! I will update dep in my PR

PR #805, opened 2021-07-30
apps/projects: test if field event_type exists in offile event

fuzzylogic2000:
We need this to make the poll detail view work with plattforms with and without timeline.
Once updated will fix https://github.com/liqd/a4-opin/issues/2282


fuzzylogic2000:
Ok, it also works in mB and I will merge now, because I wanted to update mB next and then it's already in OPIN and mB before anyone starts to test.

PR #816, opened 2021-08-24
exports/mixins: fixed export time difference and test

sabinammm:
Closes part of #3793, the other one was fixed in a4-meinberlin.


fuzzylogic2000 (approved):
Nice!

PR #817, opened 2021-08-24
polls/assets/PollQuestion: show label of other choice in browser lang…

Rineee:
…uage

Will fix #3787 after translation is added.


fuzzylogic2000 (approved):
Yeah!

PR #818, opened 2021-08-25
projects/models: add help texts to contact mixin

sabinammm:


fuzzylogic2000 (approved):
Looks good! So, let's do the easier version, where I merge first and we can then update a4 together with opening a PR doing everything else we wanted to do!

PR #820, opened 2021-08-26
report/email: getting link from module

khamui:
fixing bug described in mein Berlin issue:
https://github.com/liqd/a4-meinberlin/issues/3692

closes liqd/a4-meinberlin#3692

### DIFF
--- a/adhocracy4/polls/models.py
+++ b/adhocracy4/polls/models.py
@@ -39,7 +39,7 @@ class Poll(module_models.Item):
                                object_id_field='object_pk')
 
     def get_absolute_url(self):
-        return self.project.get_absolute_url()
+        return self.module.get_absolute_url()
 
     def annotated_questions(self):
         return self.questions.annotate_vote_count()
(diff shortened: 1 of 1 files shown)


khamui:
@fuzzylogic2000 i wasn't sure if i should use `get_detail_url` inside of `get_absolute_url` or create a new method. But at the end we wan't to get the projectlink for single module and and the modulelink for multiplee modules in any case right?

and it is someehow a property and not a method, that is why it is without `..url()`

adhocracy4/polls/models.py
```diff
@@ -39,7 +39,7 @@ class Poll(module_models.Item):
                                object_id_field='object_pk')
 
     def get_absolute_url(self):
-        return self.module.get_absolute_url()
+        return self.module.get_detail_url
```
fuzzylogic2000, line 42: Yes, it is a property. Maybe shouldn't be, but anyway. https://github.com/liqd/adhocracy4/blob/33e94010d51b5973de3b89bb3e3fb63aa5e9f03f/adhocracy4/modules/models.py#L98

fuzzylogic2000 (approved):
Yes, nice!

PR #822, opened 2021-08-30
polls/save: disable button if no changes made yet

khamui:
should fix issue in meinberlin:

https://github.com/liqd/a4-meinberlin/issues/3789

closes liqd/a4-meinberlin#3789


fuzzylogic2000 (approved):
Cool!

PR #825, opened 2021-09-03
comments_async: adjusting some classes and rm some, rm extra filter

phillimorland:


phillimorland:
This is to go with: https://github.com/liqd/adhocracy-plus/pull/1415
Opin will also have to be updated and maybe slightly adjusted but filters are already broken there, this should help fix

fuzzylogic2000 (approved):
Whoop whopp! Works!

PR #828, opened 2021-09-06
apps/polls: use new state to determine which poll link to show

fuzzylogic2000:
This fixes the problem with the showing the result not working at all. 
It was introduced with https://github.com/liqd/adhocracy4/pull/822
But I don't think this works as intended (and before also didn't). But I would like to merge, update and then ask @mcastro-lqd and @CarolingerSeilchenspringer how it should behave. There is also the other slightly weird thing where you can click "view result", even if you changed your vote and then the result is shown, but with the uservote checkmark at the wrong option.

### DIFF
--- a/adhocracy4/polls/assets/PollQuestions.jsx
+++ b/adhocracy4/polls/assets/PollQuestions.jsx
@@ -269,7 +269,7 @@ class PollQuestions extends React.Component {
             />
           ))}
           <div className="poll">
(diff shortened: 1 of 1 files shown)


sabinammm (approved):
Works fine when I'm testing it, same as beforre....But yes, I agree that the behavior should be reviewed for intent.

fuzzylogic2000:
I will just merge and update in mB. :)

PR #837, opened 2021-09-13
tests: added project method to a4 for general replacement

sabinammm:
This method serves to replace a project test method and is thus related to PR https://github.com/liqd/adhocracy-plus/pull/1432


fuzzylogic2000 (approved):
Nice! 
Maybe you can check if you can use it somewhere in the tests here as well, so that the coverage is not decreasing. Like here https://github.com/liqd/adhocracy4/blob/ced1aa4db1bf993a9db82a133c1a3ac1b64acbd1/tests/projects/test_project_views.py#L30
Will merge for now anyway, so that you can use it in a+!

PR #838, opened 2021-09-13
tests: replaced 200 response code with new method

sabinammm:
These changes in response to https://github.com/liqd/adhocracy4/pull/837#pullrequestreview-752405254.
Is this what you meant @fuzzylogic2000 or do you think that there are tests that are entirely missing that need to be written with the method? If so, could you point me in the right direction? Now looking into meinBerlin tests!

### DIFF
--- a/tests/dashboard/components/test_views_module_area_settings.py
+++ b/tests/dashboard/components/test_views_module_area_settings.py
@@ -3,6 +3,7 @@ import json
 import pytest
 
 from adhocracy4.dashboard import components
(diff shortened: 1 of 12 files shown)


fuzzylogic2000 (approved):
Yes, that's what I meant. You didn't even need to use it in all tests, just so that it's used. :) 
(And for next time: you can also use it with other error codes, even though we don't do that anywhere yet.)
And yeah, that's +0.6 % for the coverage!

PR #839, opened 2021-09-15
apps/images: make sure validation also works for APIs

fuzzylogic2000:
You can test this directly using the ideas API in a+ and add an image there (/api/modules/<module_pk>/ideas).


khamui:
cool! merging and testing (aplus / ztapp)

PR #840, opened 2021-09-15
poll issue fixing

Rineee:
fixes https://github.com/liqd/a4-meinberlin/issues/3867
fixes https://github.com/liqd/a4-meinberlin/issues/3787

### DIFF
--- a/adhocracy4/polls/assets/PollQuestions.jsx
+++ b/adhocracy4/polls/assets/PollQuestions.jsx
@@ -119,7 +119,7 @@ class PollQuestions extends React.Component {
           onClick={(e) => this.handleSubmit(e)}
           disabled={!this.state.hasVotes}
         >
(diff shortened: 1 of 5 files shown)


adhocracy4/polls/assets/PollQuestions.jsx
```diff
@@ -183,23 +183,13 @@ class PollQuestions extends React.Component {
...
-            }
-          })
+      .then(() => {
+        this.getPollData()
+        this.setState(prevState => {
+          return {
+            loading: false,
+            alert: ALERT_SUCCESS
+          }
         })
```
Rineee, line 193: @khamui Hope nothing gets lost here? I needed the updated hasUserVote, so had the get the whole poll data here...

fuzzylogic2000 (approved):
Nice! Looks good and works!

PR #844, opened 2021-09-22
setup jest testing for adhocracy4

khamui:
Does it make sense in a4?
Maybe more than in aplus (or any other projects).
There is a general dilemma (in my head). We have components in a4 and in projects.
But where should we run the tests?

I didn't manage to run jest snapshots in aplus, but also make it run tests of the node_module dependency in a4.
So for now this commit just runs snapshot tests in a4 only. Not sure if this helps or complicates everything.

What do you guys think?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -10,5 +10,16 @@
     "standard-jsx"],
   "rules": {
     "jsx-quotes": [2, "prefer-double"]
-  }
+  },
+  "overrides": [
+    {
+      "files": [
+        "**/*.jest.js",
+        "**/*.jest.jsx"
+      ],
+      "env": {
+        "jest": true
+      }
+    }
+  ]
 }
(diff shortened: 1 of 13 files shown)


goapunk:
Nice! I'd say it makes sense to have tests locally, so yeah, have them here. I'd personally prefer if we also use the testing-library as we do for the app instead of the rather limited react-test-renderer though.

khamui:
I cleaned up and structured a bit the testing files.
This part can be merged (if all good) to make it available and be used by you.

goapunk:
Thanks you, very nice. Two things which might be nice:
* Add the npm commands to the Makefile (see zt app) and maybe add the other commands as well (testDebug, testNoCov ?)
* Add coverage reports (see zt app)
Both can also be added later though in a new PR

khamui:
> Thanks you, very nice. Two things which might be nice:
> 
> * Add the npm commands to the Makefile (see zt app) and maybe add the other commands as well (testDebug, testNoCov ?)
> * Add coverage reports (see zt app)
>   Both can also be added later though in a new PR

nice thanks for feedback. will have a look into the two things (but next week :) ). so setting this to WIP.

phillimorland (approved):
super cool!

PR #846, opened 2021-09-24
tests: add more tests for async comments

fuzzylogic2000:
The second commit removes an unused file. I checked all our projects, but couldn't find any usage. (The image services were used in opin, so there are services from a4 used somewhere...)


tests/comments/test_async_templatetags.py
```diff
@@ -0,0 +1,185 @@
...
+from django.contrib.auth.models import AnonymousUser
+from django.contrib.contenttypes.models import ContentType
+from django.urls import reverse
+from freezegun import freeze_time
+
+from adhocracy4.test import helpers
+from tests.apps.questions.phases import AskPhase
+
+
+def react_comment_render_for_props(rf, user, question):
```
Rineee, line 15: I was wondering that before: where does this rf fixture come from and what does it stand for actually? :sweat_smile:

Rineee (approved):
So supercool!! And we are way over 90% now, wohooooo :partying_face:

tests/comments/test_async_templatetags.py
fuzzylogic2000: That is a very good question! I think it stands for RequestFactory, but I have no idea how it gets into our tests as rf.
https://docs.djangoproject.com/en/3.2/topics/testing/advanced/#the-request-factory
@Rineee

Rineee: Ah cool, thanks! Now that I knew what to look for, I investigated and it comes with pytest, see here: https://github.com/pytest-dev/pytest-django/blob/6ac1495c17d94db1293659edb01966c278064114/pytest_django/fixtures.py#L378

fuzzylogic2000: Nice! Thanks!

PR #847, opened 2021-09-27
Ks 2021 09 add poll export tests

Rineee:


fuzzylogic2000 (approved):
So cool! :upside_down_face:

PR #848, opened 2021-09-27
comments & comments_async: changed wording of deleted comments

sabinammm:
closes #392 

Deleted comments now show the text suggested in issue https://github.com/liqd/adhocracy-plus/issues/392. BUT

- [x] This doesn't distinguish between comments deleted by creator and moderator,
- [x] The model still has the comment as defined here https://github.com/liqd/adhocracy4/blob/626cfc7779c29456a275ad8403bf70d288cd59c6/adhocracy4/comments/models.py#L64

@phillimorland could you check my code and tell me why this isn't working as it should? https://github.com/liqd/adhocracy4/blob/df9453396e5c9e6ee8c211266823f6f6a4830866/adhocracy4/comments_async/static/comments_async/comment.jsx#L259

@fuzzylogic2000 if I try to change the comment within the model and I makemigrations, it doesn't detect these changes. What am I missing? Do I actually need these changes as initially you said I wouldn't need to change the model?


adhocracy4/comments_async/static/comments_async/comment.jsx
```diff
@@ -254,8 +254,12 @@ export default class Comment extends React.Component {
 
   render () {
     let lastDate
-    if (this.props.modified === null || this.props.is_deleted) {
+    if (this.props.modified === null) {
       lastDate = localeDate(this.props.created)
+    } if (this.props.is_deleted && this.isOwner()) {
```
fuzzylogic2000, line 259: The `this.isOwner()` is about who wrote the comment, I think. If you want to distinguish between the user deleting it and the moderator you should use `is_deleted` for the user and `is_censored` for the moderator.

fuzzylogic2000 (commented):
Hmmm, so maybe we only add the changes into the react and open an issue for the model stuff? 

And I think the serializer you changed is not even used for the comments, but am not supersure about that. :roll_eyes:

sabinammm:
> Hmmm, so maybe we only add the changes into the react and open an issue for the model stuff?
> 
> And I think the serializer you changed is not even used for the comments, but am not supersure about that.

Yes, that makes sense that the serializer has no effect, so it's probably not used...

adhocracy4/comments_async/static/comments_async/comment.jsx
sabinammm: Perfect! Thanks, will try this

sabinammm:
After discussing it with @Rineee : before this can be merged, it's necessary to

- [ ] check which projects make use of the original comments (not async), as these also use the model that has been edited.
- [ ] check failed tests.

sabinammm:
@fuzzylogic2000 this also influences meinBerlin because it uses the "old" comments. I edited them the same way as the async comments. Old comments then would still contain the text that they have been deleted, while new ones then have a label that they have been deleted. I think this is necessary because without the changes, new comments disappear without explanation. I can add this commit then it would look like this:
Comments_in_mB

Does that make sense?

adhocracy4/comments/models.py
```diff
@@ -71,10 +71,10 @@ def save(self, *args, **kwargs):
             self.comment)
 
         if self.is_removed:
-            self.comment = _('deleted by creator')
+            self.comment = ('')
```
fuzzylogic2000, line 74: As we have the same text for deleted and censored now, you can put it into the same if statement with an `or` and save a few lines. :)

fuzzylogic2000 (commented):
Oh, we also need a migration to change the text of the comments that were already deleted to the new empty string!

sabinammm:
@fuzzylogic2000 somewhat irritated: I tested the migration and it works fine. BUT even without applying the migration, the deleted comments now don't show any additional text, as if the new conditional clauses in the comment.jsx is ignored:
image
Did I mess something up during rebasing? Or maybe I have a local issue with my a4 -> project setup?

fuzzylogic2000:
> @fuzzylogic2000 somewhat irritated: I tested the migration and it works fine. BUT even without applying the migration, the deleted comments now don't show any additional text, as if the new conditional clauses in the comment.jsx is ignored. Did I mess something up during rebasing? Or maybe I have a local issue with my a4 -> project setup?

I don't know why this is happening, but it doesn't have anything to do with the new migration as they do not change the flags we use in the jsx. If you want to be sure the coupling works right, you can also use the hash of the last commit.

I will also try it out and review again!

adhocracy4/comments_async/serializers.py
```diff
@@ -148,9 +148,9 @@ class CommentListSerializer(CommentSerializer):
 
     def get_comment(self, obj):
```
fuzzylogic2000, line 149: There is no need to change this method as the model does have the empty text after the migration. You should even be able to remove the whole method as we do not need to overwrite anything anymore.

adhocracy4/comments_async/serializers.py
```diff
@@ -148,9 +148,9 @@ class CommentListSerializer(CommentSerializer):
...
         if obj.is_removed:
-            return _('deleted by creator')
+            return _('')
```
fuzzylogic2000, line 151: And if you would want to change it, there is no need to translate an empty string. :)

tests/comments/test_async_api.py
```diff
@@ -348,7 +348,7 @@ def test_fields(user, apiclient, question_ct, question):
     assert response.data['count'] == 1
 
     commentDict = response.data['results'][0]
-    assert len(commentDict) == 17
+    assert len(commentDict) == 19
```
fuzzylogic2000, line 351: Ah, because we have two new fields in the serialzer!

fuzzylogic2000 (commented):
And I will finish this review when React is working again in a+. :)

fuzzylogic2000 (approved):
Alright! It works fine in a+.
The commits are a bit weird (the first and third seem to do the same thing), and could all be squashed into one commit as they all do one thing. But I am also ok with leaving these four commits.
When you have removed the brackets and the method from the serializer I am fine with merging!

adhocracy4/comments_async/serializers.py
sabinammm: I'll definitely remove it then :). Sorry about the chaotic order - I admit that I was afraid of rebasing yesterday. Won't happen again...

fuzzylogic2000 (approved):
Yeah! Totally works for me now!
Because there is a migration that affects all projects: @Rineee Can you also test it out?

Rineee (approved):
Yes, wonderful, looks and works great!

PR #852, opened 2021-09-28
fix broken test app moderatorfeedback

goapunk:
The moderatorfeedback was missing the `__init__.py` which causes errors on django 3.x and the initial migration (since the migration in `tests/apps/ideas` depends on that (not sure whether we need that dependency?).


tests/apps/moderatorfeedback/apps.py
```diff
@@ -3,4 +3,4 @@
 
 class Config(AppConfig):
     name = 'tests.apps.moderatorfeedback'
-    label = 'a4test_moderatorfeedback'
+    label = 'moderatorfeedback'
```
fuzzylogic2000, line 6: Why did you change this?

fuzzylogic2000 (approved):
Nice!

tests/apps/moderatorfeedback/apps.py
goapunk: because the migrations would fail otherwise on Django 3.x. Weirdly it now fails on django 2.x

goapunk:
Ok, apparently this doesn't work in Django 2.x anymore, so I will just close this as the reason I fixed this was compatibility with Django 3.x

fuzzylogic2000:
Nooo! I have to decide between the tiny inconsistency in the app name and having a migration here, I would go for the migration... :p

goapunk:
Yes, but weirdly this migration is working in Django 2.x even if it shouldn't so I guess it's fine ? :)

PR #855, opened 2021-09-29
dependencies: install assert as fallback for remark-gfm

phillimorland:


phillimorland:
should have been added in https://github.com/liqd/adhocracy4/pull/827

PR #881, opened 2021-11-01
Ks 2021 10 issues

Rineee:
@fuzzylogic2000 Not sure about the deprecation warnings, could not really test as mB needs to updated to django 3 first..  what do you think?


fuzzylogic2000 (approved):
Nice! It all works! 
(You can see that the deprecation stuff is working in the a4 tests.) :)

Rineee:
> (You can see that the deprecation stuff is working in the a4 tests.) :)

Ah yes, cool!

PR #884, opened 2021-11-03
adhocracy4/comments & adhocracy4/comments_async: added is_blocked to …

sabinammm:
…model and api

The is_blocked boolean was added without refactoring/changing `exclude = ()` to `fields=()`. I will do this in a separate branch because it seems to be a bit more complicated.


Rineee:
> …model and api
> 
> Not sure if I over-edited this, BUT @fuzzylogic2000 or @Rineee why can I not see the new field is_blocked in the Comment List when I check the API? image

I think you might have not linked it to your local A4? Because when I do, I get an error ( `'ThreadListSerializer' object has no attribute 'get_is_blocked'` ) and cant access the api .. I will comment below, that should remove the error...

adhocracy4/comments/serializers.py
```diff
@@ -12,6 +12,7 @@ class CommentSerializer(serializers.ModelSerializer):
     is_deleted = serializers.SerializerMethodField()
     ratings = serializers.SerializerMethodField()
     is_moderator = serializers.SerializerMethodField()
+    is_blocked = serializers.SerializerMethodField()
```
Rineee, line 15: SerializerMethodField is only necessary when you dont want to serialize the actual value of the field, but do sth with it before, like for user_name, return 'unknown user' if the comment has been deleted. Since for is_blocked, we only want to serialize the actual value True/False, you only need to add it to the list of fields in Meta Class or, since this is the way it is done here, not add it to the exclude list. So in short: remove this line and the get_is_blocked method and it will work fine (same for async serializer).

Rineee: But in fact, I would not even add it to this serializer as we only use is_blocked in Kosmo so far which uses the async serializer. Do you agree @fuzzylogic2000 ? So: in this serializer please add it to the exclude list and in the async one, what I wrote above is valid :) Sorry, I hope this is not too confusing..

adhocracy4/comments_async/api.py
```diff
@@ -111,3 +111,11 @@ def destroy(self, request, *args, **kwargs):
...
         serializer = self.get_serializer(comment)
         return Response(serializer.data)
+
+    def block(self, request, *args, **kwargs):
+        comment = self.get_object()
+        if self.request.user == comment.creator:
+            comment.is_blocked = True
+        comment.save()
+        serializer = self.get_serializer(comment)
+        return Response(serializer.data)
```
Rineee, line 121: What is this doing?

adhocracy4/comments/serializers.py
sabinammm: No, everything makes perfect sense, thank you!! Just one clarification question: so you mean I should add it **only** to the exclude list for the regular comments/serializer.py and don't exclude it in the comments_async/serializer?

adhocracy4/comments_async/api.py
sabinammm: I added this because I couldn't see is_blocked (out of desperation), but didn't really think it was right. It's gone :)

sabinammm:
Will changed the missing test assertion once all questions are cleared.

adhocracy4/comments/serializers.py
fuzzylogic2000: No, that's exactly what @Rineee meant.
There are two ways to define what is in the serializer. One is naming the fields that should be there and the other one excluding what shouldn't. In the comment serializers we use the latter.
So, to use the is_blocked, you do not exclude it and thus it is shown. But to only show and use it in kosmo and not add it to the "old" comments, you do exclude it in the comment serializer.

adhocracy4/comments/models.py
```diff
@@ -26,6 +26,7 @@ class Comment(base.UserGeneratedContentModel):
     comment = models.TextField(max_length=4000)
     is_removed = models.BooleanField(default=False)
     is_censored = models.BooleanField(default=False)
+    is_blocked = models.BooleanField(default=False, null=True)
```
fuzzylogic2000, line 29: It shouldn't be `null=True`. It's either False or True and that's fine!

adhocracy4/comments/serializers.py
```diff
@@ -55,14 +55,14 @@ def to_internal_value(self, data):
 
     def get_user_name(self, obj):
         """
-        Don't show username if comment is marked removed or censored
+        Don't show username if comment is marked removed or censored or blocked
         """
-        if(obj.is_censored or obj.is_removed):
+        if obj.is_censored or obj.is_removed or obj.is_blocked:
```
fuzzylogic2000, line 60: Hmm, unsure here. If we don't want to care about the new field in the old comments, then we also shouldn't take it into account here? Or do we just leave it in case someone sets it in the django admin?

fuzzylogic2000: Yeah, no, I would remove it. Otherwise it's a bit weird, that the field isn't in the serializer, but still influences the other fields. Do you agree @Rineee and @sabinammm ?

adhocracy4/comments/static/comments/CommentList.jsx
```diff
@@ -22,6 +22,7 @@ const CommentList = (props) => {
               is_deleted={comment.is_deleted}
               is_removed={comment.is_removed}
               is_censored={comment.is_censored}
+              is_blocked={comment.is_blocked}
```
fuzzylogic2000, line 25: This is also for the old comments, so we don't add it, right?

fuzzylogic2000 (commented):
Nice! Nearly there!

adhocracy4/comments/serializers.py
Rineee: Yeah, I was also unsure here.. but you are probably right, even if someone sets it in django-admin, it shouldnt mean anything, so lets delete that here.

sabinammm: Yes, that makes sense because it's in the old comments. I wasn't sure how to deal with them anyway, but would take out everything that has to do with them anyway.

adhocracy4/comments/static/comments/CommentList.jsx
sabinammm: If the old comments will be entirely replaced anyway, then I will take it all out.

fuzzylogic2000 (commented):
I think now you also removed the field from the exclude for the old comments? We still need that to exclude it...

adhocracy4/comments/serializers.py
```diff
@@ -10,16 +10,24 @@ class CommentSerializer(serializers.ModelSerializer):
...
+    child_comments = serializers.ModelSerializer(many=True)
 
     class Meta:
         model = Comment
         read_only_fields = ('modified', 'created', 'id',
                             'user_name', 'ratings', 'content_type',
                             'object_pk', 'last_discussed',
                             'is_moderator_marked')
-        exclude = ('creator',)
+        fields = ('id', 'user_name', 'user_pk', 'user_profile_url',
```
sabinammm, line 24: @Rineee @fuzzylogic2000  as we discussed in stand-up, I want to replace the exclude in the comment serializer with the defined fields. Checking the model for field names didn't work for me, so I checked what the api receives without restrictions by setting ...fields = '__all__'.... This returned to me the list ...{
            "id": 9,
            "user_name": "sabina-follower",
            "user_pk": "5",
            "user_profile_url": "/profile/sabina-follower/",
            "user_image": "/static/images/avatar-01.svg",
            "is_deleted": false,
            "ratings": {
                "positive_ratings": 0,
                "negative_ratings": 0,
                "current_user_rating_value": null,
                "current_user_rating_id": null
            },
            "is_moderator": false,
            "child_comments": [],
            "created": "2021-11-03T16:26:36.983486+01:00",
            "modified": null,
            "object_pk": 1,
            "comment": "afadfdafdafa",
            "is_removed": false,
            "is_censored": false,
            "is_blocked": null,
            "comment_categories": {},
            "last_discussed": null,
            "is_moderator_marked": false,
            "creator": 5,
            "content_type": 112
        },... BUT now tests fail because e.g. Field name `user_pk` is not valid for model `Comment`. This makes sense because it's not in the model. Where does it come from?

adhocracy4/comments/serializers.py
```diff
@@ -10,16 +10,24 @@ class CommentSerializer(serializers.ModelSerializer):
     user_name = serializers.SerializerMethodField()
     user_profile_url = serializers.SerializerMethodField()
     is_deleted = serializers.SerializerMethodField()
-    ratings = serializers.SerializerMethodField()
+    ratings = serializers.ModelSerializer(many=True)
```
sabinammm, line 13: Also not sure if changing the ratings serializer this way is ok, but necessary to make it readable as a nested field in the definition of fields below.

adhocracy4/comments/serializers.py
Rineee: @sabinammm The user_pk, user_name etc come from UserGeneratedContentModel, the model that Comment class inherits from.

I dont get that error if I add the fields you added.. could you check again?
I get another error though for the child_comments, because they are added in the ThreadSerializer (this is the one we are actually using to get the child comments) below, but not explicitly in the fields list. You can solve this by adding 
`
    class Meta(CommentSerializer.Meta):
        model = Comment
        fields = CommentSerializer.Meta.fields + ('child_comments',)
`
to both the ThreadSerializer and ThreadListSerializer, so the child_comments are added to fields list there.

adhocracy4/comments/serializers.py
Rineee: We dont want to serialize the ratings as a nested fields, but only get the number of different ratings which is done in get_ratings() below. So please leave it as it and it should work fine?

adhocracy4/comments_async/serializers.py
```diff
@@ -14,16 +14,24 @@ class CommentSerializer(serializers.ModelSerializer):
...
                             'user_image_fallback', 'ratings',
                             'content_type', 'object_pk')
-        exclude = ('creator',)
+        fields = ('id', 'user_name', 'user_pk', 'user_profile_url',
+                  'user_image', 'is_deleted', 'ratings',
+                  'is_moderator', 'child_comments',
+                  'created', 'modified', 'object_pk', 'comment',
+                  'is_removed', 'is_censored', 'is_blocked',
+                  'comment_categories', 'last_discussed',
+                  'is_moderator_marked', 'content_type')
```
Rineee, line 34: oh and then remove the child_comments here if you add them in Meta of ThreadSerializers

adhocracy4/comments/serializers.py
sabinammm: > @sabinammm The user_pk, user_name etc come from UserGeneratedContentModel, the model that Comment class inherits from.
> 
> I dont get that error if I add the fields you added.. could you check again? 
@Rineee I mean this error: https://github.com/liqd/adhocracy4/runs/4116676609?check_suite_focus=true#step:10:209

I get another error though for the child_comments, because they are added in the ThreadSerializer (this is the one we are actually using to get the child comments) below, but not explicitly in the fields list. You can solve this by adding `
> 
> class Meta(CommentSerializer.Meta):
> 
> ```
> model = Comment
> 
> fields = CommentSerializer.Meta.fields + ('child_comments',)
> ```
> 
> ` to both the ThreadSerializer and ThreadListSerializer, so the child_comments are added to fields list there.

@Rineee yes! will do this, thanks!

Rineee: yes!

sabinammm:
I've changed everything in the **first commit** as reviewed by @fuzzylogic2000 and @Rineee but I'm now getting a **Bad Request** and 
image
when I try to add a child-comment.

The second commit is with regard to not showing a blocked comment and instead the "Blocked by moderator on [datetime]" tag.

fuzzylogic2000 (commented):
The code looks good and is very readable in these small commits! I will check the functionality now.

fuzzylogic2000 (approved):
Yes! It's there!

PR #885, opened 2021-11-08
Sm 2021 11 blocked comments not shown in discussion

sabinammm:
In the 3rd commit:
Changed the mechanism defined in the serializer to block comments from showing in conversation that have been set blocked by moderation. This time it should work correctly.

As of 2nd commit:
Texts of blocked comments are not shown, but marked as blocked instead. Child comments of blocked comments are still displayed:
image

As discussed, this is only implemented for async comments.

### DIFF
--- a/adhocracy4/comments_async/serializers.py
+++ b/adhocracy4/comments_async/serializers.py
@@ -27,10 +27,12 @@ class CommentSerializer(serializers.ModelSerializer):
 
     def to_representation(self, instance):
         """
-        Create a dictionary form categories.
+        Create a dictionary form categories and don't show blocked comments.
 
         Gets the categories and adds them along with their values
         to a dictionary.
+        Also gets the comments and blocks their content
+        from being shown if they are set to blocked.
         """
         ret = super().to_representation(instance)
         categories = {}
@@ -47,6 +49,9 @@ class CommentSerializer(serializers.ModelSerializer):
                 else:
                     categories[category] = category
         ret['comment_categories'] = categories
+        is_blocked = ret.get('is_blocked')
+        if is_blocked:
+            ret['comment'] = ''
         return ret
 
     def to_internal_value(self, data):
--- a/adhocracy4/comments_async/static/comments_async/comment.jsx
+++ b/adhocracy4/comments_async/static/comments_async/comment.jsx
@@ -260,6 +260,8 @@ export default class Comment extends React.Component {
       lastDate = django.gettext('Deleted by creator on') + ' ' + localeDate(this.props.modified)
     } else if (this.props.is_censored) {
       lastDate = django.gettext('Deleted by moderator on') + ' ' + localeDate(this.props.modified)
+    } else if (this.props.is_blocked) {
+      lastDate = django.gettext('Blocked by moderator on') + ' ' + localeDate(this.props.modified)
     } else {
       lastDate = django.gettext('Latest edit on') + ' ' + localeDate(this.props.modified)
     }
--- a/adhocracy4/comments_async/static/comments_async/comment_list.jsx
+++ b/adhocracy4/comments_async/static/comments_async/comment_list.jsx
@@ -25,6 +25,7 @@ export default function (props) {
               is_deleted={comment.is_deleted}
               is_removed={comment.is_removed}
               is_censored={comment.is_censored}
+              is_blocked={comment.is_blocked}
               is_moderator_marked={comment.is_moderator_marked}
               index={index}
               parentIndex={props.parentIndex}


adhocracy4/comments_async/static/comments_async/comment.jsx
```diff
@@ -198,16 +198,18 @@ export default class Comment extends React.Component {
...
-        <div className={'a4-comments__text' + (this.state.anchored ? ' a4-comments__text--highlighted' : '')}>
-          <ReactMarkdown
-            disallowedTypes={['heading']}
-            plugins={[[gfm, { singleTilde: false }]]}
-          >
-            {content}
-          </ReactMarkdown>
-        </div>
-      )
+      if (!this.props.is_blocked) {
```
fuzzylogic2000, line 201: I am a bit confused about all the ifs and elses here. 
But anyway, what is this for? Why do we need that if? Where is the difference to is_removed and is_censored?

sabinammm: Quite frankly, I don't know what the point of having is_censored and is_blocked is since I agree that these seem to mean the same thing. But since they are seperate properties of comments, I presume there is or was an intent of having both. is_removed is different since the writer of the comment can remove it, and this the shows a different label than if a moderator removes a comment:
image

fuzzylogic2000: Yes, I know that they are nearly the same. (We discussed why we added the is_blocked instead of using is_censored.)
My question was why you need the if-statement here for is_blocked that you do not need for is_censored. So, this if statement is only for is_blocked. Why did you add it?

sabinammm: Oh, this is because while is_censored deletes the text of the comment, is_blocked doesn't delete the text, it only stops it from being shown. Does that make sense?

fuzzylogic2000: Yes! But we also do not show the text for the deleted comments (you changed that in your PR fixing how the deleted comments are shown, right?). So, do we change the text in the serializer? Or how do we manage to not show the text for censored or removed comments?
Right, or did you change that in the method? Yes, that's what we did! Hmm, then maybe instead of this if-statement we do change it in the serializer? Or make the whole if-bit more readable. 
I would prefer changing the serializer. What do you think @Rineee and @phillimorland ?

phillimorland: as we already use the serializer to remove the user image and name I think we should change it in the serializer too, it's more consistent with our other code

sabinammm: Ok, then:
Do I overwrite a get-method for the comment text itself in the comments_async/serializer to return the comment(-text) only if it is not is_blocked? 
Then I could get rid of this if-statement, but just to make sure I'm getting the method right...

fuzzylogic2000: Yes, that's how you do it! You add it as a SerializerMethodField and tell it in the `get_comment` what to do.

sabinammm: Ok, I've figured out how to change it and pushed this. But 4 tests are failing due to it having a negative impact on the comments (i.e. child comments made after the blocking are also blocked automatically). So I still have to solve this.

adhocracy4/comments_async/serializers.py
```diff
@@ -31,6 +31,9 @@ def to_representation(self, instance):
 
         Gets the categories and adds them along with their values
         to a dictionary.
+
+        Also gets the comments and blocks their content
+        from being shown if they are set to blocked.
```
fuzzylogic2000, line 36: The docstring should always start with one sentence saying what the whole method does, and then if needed a longer explanation. So, your addition should also go into the first sentence. And remove the newline in the longer explanation, please.
https://www.python.org/dev/peps/pep-0257/

fuzzylogic2000: But only, if we don't put this elsewhere anyway...

adhocracy4/comments_async/serializers.py
```diff
@@ -47,6 +50,9 @@ def to_representation(self, instance):
                 else:
                     categories[category] = category
         ret['comment_categories'] = categories
+        is_blocked = ret.get('is_blocked')
```
fuzzylogic2000, line 53: While in the docs of the model serializer, it says to use this to add fields https://www.django-rest-framework.org/api-guide/serializers/#advanced-serializer-usage
There is also another way of doing that: https://www.django-rest-framework.org/api-guide/relations/#custom-relational-fields
Should we change that or keep it? What do you think @Rineee ?

adhocracy4/comments_async/serializers.py
```diff
@@ -47,6 +50,9 @@ def to_representation(self, instance):
...
+        if is_blocked:
+            ret.__setitem__('comment', '')
```
fuzzylogic2000, line 55: Why not `ret['comment'] = ''`?

fuzzylogic2000 (changes requested):
Nice! It works!
Can you please squash the commit, though. While we like small commits that add only one thing, commits in one PR shouldn't change things from the same PR. See https://wiki.liqd.net/git 
> Commits within the same PR should not correct each other.

adhocracy4/comments_async/serializers.py
sabinammm: I changed this in reference to your first comment, but don't understand what you mean with the second comment. I added this as a refactor commit. If that's too much, I can squash them again, though.

adhocracy4/comments_async/serializers.py
sabinammm: Will wait for @Rineee to feedback on this

adhocracy4/comments_async/serializers.py
sabinammm: I refactored this in the last commit. Thx!

adhocracy4/comments_async/serializers.py
Rineee: So you probably thought about that, but what was the problem with just adding `comment = serializers.SerializerMethodField()` and then return an empty string in `get_comment()` if the comment is_blocked?

If that doesnt work, I think we need to leave it here in to_representation as it needs the is_blocked to decide what is serialized And that doesnt work with the above approach as this only gets the comment value from database and then operates on that. So it doesnt have the whole instance (with is_blocked), but just the field itself. I think?

BUT: I would definitely refactor the category serialization above with custom relational fields. I also did that for labels in the idea serializer in Aplus here: https://github.com/liqd/adhocracy-plus/blob/a4d62b89cbe297202a3d18dbb2b9f86ad75eaf5d/apps/ideas/serializers.py#L12.

adhocracy4/comments_async/serializers.py
Rineee: I think what @fuzzylogic2000 meant in the second comment was, that you dont need to change the docstrings in case we dont leave the serialization of the comment in this method, but do it another, see discussed below... 
Concerning the different commits: I also think it would be good to squash them as @fuzzylogic2000 explained here https://github.com/liqd/adhocracy4/pull/885#pullrequestreview-803832660.

adhocracy4/comments_async/serializers.py
sabinammm: @Rineee adding the SerializerMethodField and get_comment() caused the errors that I initially mentioned in https://github.com/liqd/adhocracy4/pull/885#issue-1047277641, which is why I then edited to_representation to include changing the way the comment text was **represented** instead of overwriting the content of the text permanently (as well as the child-comments) and causing those errors. This approach was visible before I squashed the commits.

Just to make sure I understand your last suggestion correctly: The approach in https://github.com/liqd/adhocracy-plus/blob/a4d62b89cbe297202a3d18dbb2b9f86ad75eaf5d/apps/ideas/serializers.py#L12 to refactor the category serialization would then mean that I 

- take this out of to_representation in the CommentSerializer and move it to a separate class CategorySerializer(serializers.StringRelatedField)?
- Why not model it on the CategoryField(serializers.Field)?

Thanks for your feedback:pray:

sabinammm: @Rineee Just to make sure I understand your last suggestion correctly: The approach in https://github.com/liqd/adhocracy-plus/blob/a4d62b89cbe297202a3d18dbb2b9f86ad75eaf5d/apps/ideas/serializers.py#L12 to refactor the category serialization would then mean that I 

- take this out of to_representation in the CommentSerializer and move it to a separate class CategorySerializer(serializers.StringRelatedField)?
- Why not model it on the CategoryField(serializers.Field)?

Thanks for your feedback:pray:

fuzzylogic2000 (approved):
Nice!

PR #894, opened 2021-11-15
tests/comments: test async comment serializer

sabinammm:
This test is written based on https://github.com/liqd/a4-meinberlin/blob/main/tests/plans/test_serializer.py

@fuzzylogic2000 it produces the following error message:
E               TypeError: Comment() got an unexpected keyword argument 'comments'

I presume I'm not understanding how to translate the example to the comment, comment list and serializer.

### DIFF
--- a/tests/comments/test_async_serializer.py
+++ b/tests/comments/test_async_serializer.py
@@ -1,0 +1,193 @@
+import pytest
+from django.contrib.contenttypes.models import ContentType
+from django.urls import reverse
+
+from adhocracy4.comments.models import Comment
+from tests.apps.questions.models import Question
+
+
+@pytest.mark.django_db
+def test_serializer(apiclient, question, comment_factory, user):
+    question_ct = ContentType.objects.get_for_model(Question)
+    comment_ct = ContentType.objects.get_for_model(Comment)
+    moderator = question.project.moderators.first()
+
+    comment_factory(pk=1, content_object=question, creator=user)
+    comment_factory(pk=2, content_object=question, creator=user,
+                    comment_categories='QUE')
+    comment_factory(pk=3, content_object=question, creator=user,
+                    is_blocked=True)
+    comment_factory(pk=4, content_object=question, creator=user,
+                    is_censored=True)
+    comment_factory(pk=5, content_object=question, creator=user,
+                    is_removed=True)
+    comment_factory(pk=6, content_object=question, creator=user,
+                    is_moderator_marked=True)
+    comment = comment_factory(pk=7, content_object=question, creator=moderator)
+    comment_factory(pk=8, content_object=comment, creator=user)
+
+    apiclient.force_authenticate(user=user)
+    url = reverse(
+        'comments_async-list',
+        kwargs={'content_type': question_ct.pk,
+                'object_pk': question.pk})
+    response = apiclient.get(url)
+
+    comment_data = response.data['results']
+    assert len(comment_data) == 7
+
+    # Testing a plain comment
+    comment_1 = comment_data[6]
+    assert comment_1['id'] == 1
+    assert comment_1['user_name'] == ''  # should be user.username ?!!
+    # In the react_comments_async template tags,
+    # this is set as user_name = str(user.id)
+    # And in the react_comments template tags,
+    # this is set to user.username AFTER the serializer
+    assert comment_1['user_pk'] == str(user.pk)
+    # django auth user model does not define get_absolute_url()
+    assert comment_1['user_profile_url'] == ''
+    assert comment_1['is_deleted'] is False
+    assert comment_1['is_moderator'] is False
+    assert len(comment_1['child_comments']) == 0
+    assert comment_1['comment'] == Comment.objects.get(pk=1).comment
+    assert comment_1['object_pk'] == question.pk
+    assert comment_1['is_removed'] is False
+    assert comment_1['is_censored'] is False
+    assert comment_1['is_blocked'] is False
+    assert len(comment_1['comment_categories']) == 0
+    assert comment_1['last_discussed'] is None
+    assert comment_1['is_moderator_marked'] is False
+    assert comment_1['content_type'] == question_ct.pk
+
+    # Testing comment with a defined category:
+    comment_2 = comment_data[5]
+    assert comment_2['id'] == 2
+    assert comment_2['user_name'] == ''
+    assert comment_2['user_pk'] == str(user.pk)
+    assert comment_2['user_profile_url'] == ''
+    assert comment_2['is_deleted'] is False
+    assert comment_2['is_moderator'] is False
+    assert len(comment_2['child_comments']) == 0
+    assert comment_2['comment'] == Comment.objects.get(pk=2).comment
+    assert comment_2['object_pk'] == question.pk
+    assert comment_2['is_removed'] is False
+    assert comment_2['is_censored'] is False
+    assert comment_2['is_blocked'] is False
+    assert len(comment_2['comment_categories']) == 1
+    assert comment_2['comment_categories'] == {'QUE': 'Question'}
+    assert comment_2['last_discussed'] is None
+    assert comment_2['is_moderator_marked'] is False
+    assert comment_2['content_type'] == question_ct.pk
+
+    # Testing a comment that is set to blocked by a moderator:
+    comment_3 = comment_data[4]
+    assert comment_3['id'] == 3
+    assert comment_3['user_name'] == ''
+    assert comment_3['user_pk'] == str(user.pk)
+    assert comment_3['user_profile_url'] == ''
+    assert comment_3['is_deleted'] is False
+    assert comment_3['is_moderator'] is False
+    assert len(comment_3['child_comments']) == 0
+    assert comment_3['comment'] != Comment.objects.get(pk=3).comment
+    assert comment_3['comment'] == ''
+    assert comment_3['object_pk'] == question.pk
+    assert comment_3['is_removed'] is False
+    assert comment_3['is_censored'] is False
+    assert comment_3['is_blocked'] is True
+    assert len(comment_3['comment_categories']) == 0
+    assert comment_3['last_discussed'] is None
+    assert comment_3['is_moderator_marked'] is False
+    assert comment_3['content_type'] == question_ct.pk
+
+    # Testing a comment that has been censored (deleted) by a moderator:
+    comment_4 = comment_data[3]
+    assert comment_4['id'] == 4
+    assert comment_4['user_name'] == 'unknown user'
+    assert comment_4['user_pk'] == -1
+    assert comment_4['user_profile_url'] == ''
+    assert comment_4['is_deleted'] is True
+    assert comment_4['is_moderator'] is False
+    assert len(comment_4['child_comments']) == 0
+    assert comment_4['comment'] == Comment.objects.get(pk=4).comment
+    assert comment_4['comment'] == ''
+    assert comment_4['object_pk'] == question.pk
+    assert comment_4['is_removed'] is False
+    assert comment_4['is_censored'] is True
+    assert comment_4['is_blocked'] is False
+    assert len(comment_4['comment_categories']) == 0
+    assert comment_4['last_discussed'] is None
+    assert comment_4['is_moderator_marked'] is False
+    assert comment_4['content_type'] == question_ct.pk
+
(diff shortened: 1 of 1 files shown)


tests/comments/test_async_serializer.py
```diff
@@ -0,0 +1,20 @@
...
+    comment_factory.create(pk=1)
+    comment_factory.create(pk=2, is_blocked=True)
+    comment_factory.create(pk=3, is_censored=True)
+    comment_factory.create(pk=4, is_removed=True)
+    comment_factory.create(pk=5, is_moderator_marked=True)
+
+    comment_list = Comment.objects.all().order_by('pk')
+
+    comment_serializer = CommentSerializer(comment_list, many=True)
+    comment_data = comment_serializer.data
```
sabinammm, line 19: This line fails, as trace shows 
*** KeyError: 'request'
with a reference to [line 123 in the serializer](https://github.com/liqd/adhocracy4/blob/3a30aa9cd701416516f41a66e731151f70b51aea/adhocracy4/comments_async/serializers.py#L123):
        user = self.context['request'].user 
after the test has run.

tests/comments/test_async_serializer.py
```diff
@@ -0,0 +1,23 @@
...
+    comment_factory.create(pk=2, is_blocked=True)
+    comment_factory.create(pk=3, is_censored=True)
+    comment_factory.create(pk=4, is_removed=True)
+    comment_factory.create(pk=5, is_moderator_marked=True)
+
+    comment_list = Comment.objects.all().order_by('pk')
+
+    comment_serializer = CommentSerializer(comment_list, many=True)
+
+    # comment_data = comment_serializer.data
```
sabinammm, line 20: This line fails, as trace shows 
*** KeyError: 'request'
with a reference to [line 123 in the serializer](https://github.com/liqd/adhocracy4/blob/3a30aa9cd701416516f41a66e731151f70b51aea/adhocracy4/comments_async/serializers.py#L123):
        user = self.context['request'].user 
after the test has run.

@fuzzylogic2000 I can't reference the data for other assert statements.

Would I have to add someting like
`apiclient.force_authenticate(user=user)`
?

sabinammm: I also tried to add 
`request = rf.get('/')`
`comment_serializer = CommentSerializer(comment_list, many=True, context={'request': request})`

as it is done in tests/categories/test_categories_filters.py

This returns:

comment_serializer.data
*** AttributeError: 'WSGIRequest' object has no attribute 'user'

fuzzylogic2000:
I added the changes as a second commit, so you can see what I've changed. When the tests are done, please put them all in one commit.
As you also saw, the serializer needed the user from the request to work. To get that, I used the apiclient with an authenticated user. And what I did then, was adding a few different things and checking all the fields. The comments you added with the fields you actually wanted to test are left for you. So, check in the test what these differerent fields do to the other fields. And check the serializer for more stuff, we also might need to or want to check. (Like last_discussed, even though, I think we have tests checking that field somewhere elready?)
Right, but with using the apiclient, we get a different order of the comments (by created date). To easily spot when the order of the comments is a problem in these tests, I added the assert for the pk.

tests/comments/test_async_serializer.py
```diff
@@ -0,0 +1,181 @@
...
+        kwargs={'content_type': question_ct.pk,
+                'object_pk': question.pk})
+    response = apiclient.get(url)
+
+    comment_data = response.data['results']
+    assert len(comment_data) == 7
+
+    comment_1 = comment_data[6]
+    assert comment_1['id'] == 1
+    assert comment_1['user_name'] == ''  # should be user.username ?!!
```
sabinammm, line 41: Regarding the question: Is this part of react, i.e. comment_list.jsx tested in test_templatetags.py and thus not mocked here?

sabinammm: These are my thoughts on the changes I implemented:

- `is_removed` and `is_censored` are exactly the same in the context of the other serializer fields, except for those two fields. The difference is who can perform the action, which is already tested in **test_api.py**.
- `last_discussed` is tested in **test_signals.py**
- I added one comment for testing the comment categories because they are set in **settings.py** and might be needed somewhere, but are never tested anywhere thus far.

Thanks for this!! It visualizes how the fields affect each other very well. 
Is there anything else that I missed?

tests/comments/test_async_serializer.py
```diff
@@ -0,0 +1,181 @@
...
+    assert comment_2['is_deleted'] is False
+    assert comment_2['is_moderator'] is False
+    assert len(comment_2['child_comments']) == 0
+    assert comment_2['comment'] == Comment.objects.get(pk=2).comment
+    assert comment_2['object_pk'] == question.pk
+    assert comment_2['is_removed'] is False
+    assert comment_2['is_censored'] is False
+    assert comment_2['is_blocked'] is False
+    assert len(comment_2['comment_categories']) == 1
+    comment_1['comment_categories'] == 'QUE'
```
fuzzylogic2000, line 72: I guess, you did forget the assert here?

tests/comments/test_async_serializer.py
fuzzylogic2000: Cool, that you tested all the fields!
But I thought you could also find weird things and inconsistencies in the serializer. For example finding out why the user does not have a name here. And what we use is_deleted for, when we also have is_removed and so on. Also, should the user pk and name for blocked comments also change?

fuzzylogic2000: Maybe put a comment above the assert blocks what it is testing. Like "is_blocked, one category", "child comment", "comment added by moderator" and so on? Then it might be easier to spot the weird things.

fuzzylogic2000 (commented):
Nice! Just little things.

sabinammm:
@fuzzylogic2000 

- I created three new issues with regards to potentially redundant attributes and their effects.
- I think it would be important to test the effect of the user_image of blocking by moderatores etc. But I couldn't figure out how to mock a user_image that is not None.  The same as here is the case in the [api test](https://github.com/liqd/adhocracy4/blob/6e50fc38aea91b851d77549f32cc02bfa173e766/tests/comments/test_async_api.py#L365).
- I attempted to figure out why the user_name is not set to user.username in the test. The same as here is the case in the [api test](https://github.com/liqd/adhocracy4/blob/6e50fc38aea91b851d77549f32cc02bfa173e766/tests/comments/test_async_api.py#L368).

tests/comments/test_async_serializer.py
```diff
@@ -0,0 +1,229 @@
...
+    assert len(comment_data) == 7
+
+    # Testing a plain comment
+    comment_1 = comment_data[6]
+    assert comment_1['id'] == 1
+    assert comment_1['user_name'] == ''  # should be user.username ?!!
+    # In the react_comments_async template tags,
+    # this is set as user_name = str(user.id)
+    # And in the react_comments template tags,
+    # this is set to user.username AFTER the serializer
```
fuzzylogic2000, line 46: I think that is all a bit weird and should be refactored. I mean the code doing that not the test. :)

tests/comments/test_async_serializer.py
```diff
@@ -0,0 +1,229 @@
...
+    assert comment_3['comment'] != Comment.objects.get(pk=3).comment
+    assert comment_3['comment'] == ''
+    assert comment_3['object_pk'] == question.pk
+    assert comment_3['is_removed'] is False
+    assert comment_3['is_censored'] is False
+    assert comment_3['is_blocked'] is True
+    assert len(comment_3['comment_categories']) == 0
+    assert comment_3['last_discussed'] is None
+    assert comment_3['is_moderator_marked'] is False
+    # Shouldn't a blocked comment be moderator marked?
```
fuzzylogic2000, line 101: No, the moderator mark is the opposite. It is used for comments that the moderator wants to everyone to see and is used to put the comment at the top at the list with the ordering filter in one project.

tests/comments/test_async_serializer.py
```diff
@@ -0,0 +1,229 @@
...
+    assert comment_3['is_deleted'] is False
+
+    # A censored comment is actually is_deleted.
+    assert comment_4['is_censored'] is True
+    assert comment_4['is_deleted'] is True
+    # But does this make sense or should the content remain accessible?
+
+    # A removed comment should be is_deleted:
+    assert comment_5['is_removed'] is True
+    assert comment_5['is_deleted'] is True
```
fuzzylogic2000, line 157: But what do we use the is_deleted for? Does it make sense to put it into the serializer like that? I guess it is only in the serializer, or is it also a property of the model?
I mean, it does make sense that it is `True` for is_censored and is_removed and `False` for is_blocked, because for the first two we delete the comment text, but we do not do that for the latter. But do we need this field in the serializer? What do we use it for in the React code? And if you know what we use it for, does it make sense, that is_blocked is not is_deleted?

tests/comments/test_async_serializer.py
```diff
@@ -0,0 +1,229 @@
...
+    # The difference between is_removed and is_censored,
+    # is that is_removed is carried out by the creator,
+    # and is_censored can only be done by a moderator:
+    assert comment_5['is_moderator'] is False
+    # But this differentiation does not occur in the serializer:
+    assert comment_4['is_moderator'] is False
+    # Because the serializer does not differentiate between
+    # moderator and creator for deleting the comment,
+    # does it actually make sense to show this
+    # differentiation in the serializer?
```
fuzzylogic2000, line 168: The is_moderator field is True for comments added by a moderator. (As you can see in the comment added by a moderator in the test below). We use it to display a "moderator" badge with theses comments, where the creator was also moderator of the project.

tests/comments/test_async_serializer.py
```diff
@@ -0,0 +1,229 @@
...
+
+    # Why is there an is_removed property
+    # if there is also an is_deleted property?
```
fuzzylogic2000, line 171: is_removed is a model field and is_deleted is either only a serializer field or a serializer field and a property of the model.

tests/comments/test_async_serializer.py
```diff
@@ -0,0 +1,229 @@
...
+    assert comment_6['object_pk'] == question.pk
+    assert comment_6['is_removed'] is False
+    assert comment_6['is_censored'] is False
+    assert comment_6['is_blocked'] is False
+    assert len(comment_6['comment_categories']) == 0
+    assert comment_6['last_discussed'] is None
+    assert comment_6['is_moderator_marked'] is True
+    assert comment_6['content_type'] == question_ct.pk
+    # Moderator-marked comments no longer show the user name
+    # nor the link to the user profile. Does this make sense?
```
fuzzylogic2000, line 192: That's not true, is it? The moderator marked does not change anything apart from the boolean field is_moderator_marked, which is how it should be. The username and the link do not work for all comments in these tests. The link because Django's user model does not include the get_absolute_url and the name I commented above.

tests/comments/test_async_serializer.py
```diff
@@ -0,0 +1,229 @@
...
+
+    # Testing a comment with child-comments:
```
fuzzylogic2000, line 194: and added by a moderator

tests/comments/test_async_serializer.py
```diff
@@ -0,0 +1,229 @@
...
+    assert comment_8['is_removed'] is False
+    assert comment_8['is_censored'] is False
+    assert comment_8['is_blocked'] is False
+    assert len(comment_8['comment_categories']) == 0
+    assert comment_8['last_discussed'] is None
+    assert comment_8['is_moderator_marked'] is False
+    assert comment_8['content_type'] == comment_ct.pk
+
+    # Notes:
+    # The property 'last_discussed' is tested in test_signals.py.
```
fuzzylogic2000, line 228: The property last_discussed should change to a date when a child comment is added. But obviously that is not working when the child comment is added via the comment_factory.

tests/comments/test_async_serializer.py
```diff
@@ -0,0 +1,229 @@
...
+    # Is the property 'is_moderator_marked' ever actually used?
```
fuzzylogic2000, line 229: yes, for an ordering filter in one project. We can just keep it (and either remove it at some point when we do not need it anymore or also use it in another project.

fuzzylogic2000 (commented):
Alright! I added some comments. I think, I will now remove your comments from the tests (after checking that you can still see them in the review) and merge.

PR #904, opened 2021-11-25
adhocracy4/polls: show disabled arrows for one open question answer

sabinammm:
Fixes https://github.com/liqd/a4-meinberlin/issues/3788


fuzzylogic2000 (changes requested):
But this does the opposite from what you say it does.
Screenshot from 2021-11-26 09-54-23

phillimorland:
@sabinammm @fuzzylogic2000 this was actually more complicated then it should have been, we were overwriting the arrows in the component but actually that wasn't working correctly so I ended up fixing issue in another pull request:

PR #905, opened 2021-11-26
Sm 2021 11 issue blocked comments cannot be rated

### LINKED ISSUE #899: async serializer: also set is_deleted to True for blocked comments
**URL:** 
**user:** 
**expected behaviour:** readable code, comments should have distinctive properties with distinct names, meanings and behavior
**behaviour:** **Async comments** have the properties `is_removed` and `is_deleted`. Comments that are either removed (by a moderator) or deleted (by a creator) are both set to `is_deleted` and their content is permanently deleted.
**important screensize:**
**device & browser:** 
**Comment/Question:** 
Because the serializer does not differentiate between moderator and creator for deleting the comment, does it actually make sense to show this differentiation in the serializer?
This is related to #902 and #903
Screenshot?

sabinammm:
Fixes #899

### DIFF
--- a/adhocracy4/comments_async/serializers.py
+++ b/adhocracy4/comments_async/serializers.py
@@ -112,7 +112,7 @@ class CommentSerializer(serializers.ModelSerializer):
 
     def get_is_deleted(self, obj):
         """Return true if one of the flags is set."""
-        return (obj.is_censored or obj.is_removed)
+        return obj.is_censored or obj.is_removed or obj.is_blocked
 
     def get_ratings(self, comment):
         """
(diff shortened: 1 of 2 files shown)


fuzzylogic2000 (commented):
Cool, it disables the voting.
But now other things are behaving a bit weird (we do not show the user image, but the user name and more). I will open an issue in kosmo discussing what exactly the blocked comment should look like. And then when we know, we can change it to that.
Current status:
Screenshot from 2021-11-26 15-43-13

With changes from this PR:
Screenshot from 2021-11-26 15-41-23

fuzzylogic2000:
https://github.com/liqd/a4-kosmo/issues/166

PR #906, opened 2021-11-29
WIP: Kl 2021 11 add module to request

fuzzylogic2000:
By removing the ClassBasedViewFilterSet, it obviously breaks the tests. But before looking into that, I would like to know if this can be the way to go.


fuzzylogic2000:
As annotating in the filter does not produce more SQL-queries (additionally to the supermany ones that are there...), I will close this and use the other approach. like https://github.com/liqd/a4-meinberlin/pull/3996

PR #912, opened 2021-12-01
apps/filters: make sure ordering also works on annotated fields

fuzzylogic2000:


fuzzylogic2000:
Same here! Done! I will try to add tests, though... @Rineee @goapunk

adhocracy4/filters/widgets.py
```diff
@@ -61,6 +61,12 @@ def render(self, name, value, attrs=None, choices=(), renderer=None):
         })
 
 
+class OrderingWidget(DropdownLinkWidget):
+    """Used in .filters.OrderingFilter"""
```
Rineee, line 65: This should be DynamicChoicesOrderingFilter, right?

adhocracy4/filters/filters.py
```diff
@@ -75,6 +78,27 @@ def filter(self, qs, value):
...
+
+    def filter(self, qs, value):
+        qs = qs.annotate_positive_rating_count() \
+            .annotate_negative_rating_count() \
+            .annotate_comment_count()
+        if value == ['-comment_count']:
+            return qs.order_by('-comment_count')
+        elif value == ['-positive_rating_count']:
+            return qs.order_by('-positive_rating_count')
+
```
Rineee, line 98: Not sure, but maybe do we want a short comment, why we are annotating here? Just so we remember in the future..

Rineee (approved):
Yes, supercool! :partying_face:  Just two tiny things about comments..

adhocracy4/filters/widgets.py
fuzzylogic2000: Right, good catch! I renamed the filter while moving and added the new comment at the same time and then did not update that...

adhocracy4/filters/filters.py
fuzzylogic2000: Yes! Good idea!

PR #923, opened 2021-12-14
polls/PollReszults: rm overwrite that was braking behaviour and add i…

phillimorland:
…n classname for additional styling when less then 1 answer

- removed the overwrite and reimplemented the default behaviour for slick slider with additional styling in mB pull request

### DIFF
--- a/adhocracy4/polls/assets/PollResults.jsx
+++ b/adhocracy4/polls/assets/PollResults.jsx
@@ -2,26 +2,6 @@ import React from 'react'
 import Slider from 'react-slick'
 import django from 'django'
 
-function NextArrow (props) {
-  const { onClick } = props
-  return (
-    <button className="slick-next" onClick={onClick} aria-label={django.ngettext('Next answer')}><i className="fa fa-chevron-right" /></button>
-  )
-}
-
-function PrevArrow (props) {
-  const { onClick } = props
-  return (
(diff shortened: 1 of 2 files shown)


phillimorland:
will also fix spinner

sabinammm (approved):
Wow - looks really nice now!

adhocracy4/polls/assets/PollResults.jsx
```diff
@@ -156,11 +134,9 @@ export default class PollResult extends React.Component {
                               <div className="poll-slider__answer">
                                 {this.isUserAnswer(slide) && <i className="fas fa-check-circle" />} {slide.answer}
                               </div>
-                              <div className="poll-slider__count">
+                              <div className={this.props.question.other_choice_answers.length > 1 ? 'poll-slider__count--spaced' : 'poll-slider__count'}>
```
fuzzylogic2000, line 137: Ah, I needed a while to get that. But this only styles the number of the answer, right? So the `3/127`. And not showing the numbers when there is only one thing in the slider comes from the slider and that's the overwrite you removed? Good catch!

fuzzylogic2000 (approved):
Nice! The code looks good to me and if @sabinammm says the page also looks good, then I will merge!

adhocracy4/polls/assets/PollResults.jsx
phillimorland: this actually just moves the number to the corner when the slider automatically removes the arrows when there's only 1 slide

PR #925, opened 2021-12-15
deps/python: update psycopg2-binary

fuzzylogic2000:
I guess we should also update it here?


goapunk:
@fuzzylogic2000 I do wonder why it was set to < 2.9, but as it seems to work on our dev server so far I guess it's fine

goapunk (approved):

PR #926, opened 2021-12-17
modules/models: show module name in string representation of module a…

Rineee:
…nd remove linting errors

Super annoying with the docstring linting.. it forces you to have one summary sentence that ends with a . , so I had to shorten some descriptions :woman_shrugging:


fuzzylogic2000 (approved):
Nice!

PR #931, opened 2021-12-21
apps/modules/predicates: refactor predicates

### LINKED ISSUE #927: refactor project and module predicates
**URL:** 
**user:** 
**expected behaviour:** 
**behaviour:** We currently have two places to check predicated depending on items. Most are separately added in the modules predicates, using the projects predicate, but with the project from the item:
for initiators: https://github.com/liqd/adhocracy4/blob/4dbbd6f7c2d882295c92f2c51a08650f78f131fd/adhocracy4/modules/predicates.py#L16
for moderators: https://github.com/liqd/adhocracy4/blob/4dbbd6f7c2d882295c92f2c51a08650f78f131fd/adhocracy4/modules/predicates.py#L23
and also if the project (and thus the item) is live: https://github.com/liqd/adhocracy4/blob/4dbbd6f7c2d882295c92f2c51a08650f78f131fd/adhocracy4/modules/predicates.py#L53

Only for project group members, we check in the project predicates for both cases, when we get the project or the item:
https://github.com/liqd/adhocracy4/blob/4dbbd6f7c2d882295c92f2c51a08650f78f131fd/adhocracy4/projects/predicates.py#L7

**important screensize:**
**device & browser:** 
**Comment/Question:** While I find both methods make sense, I would like to only do it ONE way.
Either check both in the project predicates (checking if we already have the project and otherwise getting it) 
or check the predicates relying on projects in the project predicates and then add them again in the module predicates and use them to check if the project of the item has the predicate.

I think, I slightly prefer the latter as it is more explicit (not checking `hasattr(subject, 'project')` to find out, if we already have the project or sth. else) and it would only mean refactoring one predicate. :)
@Rineee @sabinammm Which one do you prefer?

fuzzylogic2000:
- improves naming of one predicate
- fixes #927
- adds project group members to the tests

There is more to do, but this is a start. :) And with that, we can also work on the rules in mB and make all the things not allowed for moderators, that we need to. So, can we already merge and later when the story is done, we can add more tests?

### DIFF
--- a/adhocracy4/modules/predicates.py
+++ b/adhocracy4/modules/predicates.py
@@ -1,3 +1,5 @@
+import warnings
+
 import rules
 from rules import predicates as rules_predicates
 
@@ -12,6 +14,7 @@ from adhocracy4.projects.predicates import is_public
 from adhocracy4.projects.predicates import is_semipublic
 
 
+# Predicates testing roles
 @rules.predicate
 def is_context_initiator(user, item):
     if item:
@@ -27,7 +30,20 @@ def is_context_moderator(user, item):
 
 
 @rules.predicate
+def is_context_group_member(user, item):
+    if item:
+        return is_prj_group_member(user, item.project)
+
+
+@rules.predicate
 def is_context_member(user, item):
+    """ Return if normal user is project participant or org member.
+
+
+    In public projects every registered user is a participant.
+    In private or semi-public projects only invited participants are
+    participants.
+    """
     if item:
         return (is_project_member(user, item.project) |
                 is_org_member(user, item.project.organisation))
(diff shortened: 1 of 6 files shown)


adhocracy4/modules/predicates.py
```diff
@@ -26,8 +29,21 @@ def is_context_moderator(user, item):
     return False
 
 
+@rules.predicate
+def is_context_group_member(user, item):
+    if item:
+        return is_prj_group_member(user, item.project)
```
sabinammm, line 35: Yes, this makes a lot more sense :)

sabinammm (approved):
Awesome!!

fuzzylogic2000:
@Rineee Could you have a look and maybe merge if you don't hate it? :grimacing:

Rineee (approved):
Cool!

PR #932, opened 2021-12-21
Ks 2021 12 voting phase predicates

Rineee:
Not sure if these comments make sense?


fuzzylogic2000 (approved):
Yes, I think, it does make sense! Hopefully we still think so in three months. :upside_down_face:

PR #933, opened 2021-12-22
display maps: leaflet.markerClusterGroup import earlier

khamui:
to fix issue in meinBerlin 

closes liqd/a4-meinberlin#4100


phillimorland (commented):
nice solution to pass linter but won't install with mB due to updates to proposal item so can't test will merge and try testing with the commit

PR #938, opened 2021-12-28
[#5051(part 1)]Comment async: add classes for mB styling and small changes

phillimorland:
part 1 of adding async comments to mB, expect to have to make adjustments in Opin and A+ as did small changes for sematic purposes and rm some classnames that were difficult to overwrite in mB.
This can be merged before mB release as will not effect current main


khamui:
looking good from what I see in code!

PR #943, opened 2022-01-04
apps/filters: only annotate rating when item has rates

fuzzylogic2000:
Turns out: not all items can be rated...


Rineee (approved):
Right!

PR #944, opened 2022-01-07
apps/dashboard: wording fix

fuzzylogic2000:
fixes https://github.com/liqd/adhocracy-plus/issues/293 (but we will have to look out for that string in mB and OPIN as well, then).


Rineee (approved):
oh, yes!

PR #952, opened 2022-01-17
apps/emails: add organsiation to get_site and report emails

fuzzylogic2000:
This even works locally. Because of hasattr, it also works with projects without a site added to the organisation.
I thought about also adding the organisation stuff to the attachments to display another logo for organisation emails like in a+, but as we do not have that in mB, but do have organisation logos there, I decided against that.

### DIFF
--- a/adhocracy4/emails/base.py
+++ b/adhocracy4/emails/base.py
@@ -21,7 +21,15 @@ class EmailBase:
     # then pass the exception with all mails to `handle_report`
     enable_reporting = False
 
+    def get_organisation(self):
+        return
+
     def get_site(self):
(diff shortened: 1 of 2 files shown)


Rineee:
@fuzzylogic2000 Can you give me some instructions how to test this? How do I properly set the site of an organisation?

fuzzylogic2000:
@Rineee Alright, this is not draft anymore. As we do not do the languages for now, this is ready for review!
Please test with the explanation above and this PR: https://github.com/liqd/adhocracy-plus/pull/1653

Rineee (approved):
Yes, that works perfectly now!

PR #957, opened 2022-01-19
comments/serializers: add comment_content_type

Rineee:
I didnt replace the now used comments_contenttype that is passed through template tags so far as I thought we can do that when we start working on the comments anyway? So for now, this is just extra..


fuzzylogic2000 (approved):
Yeah!

PR #963, opened 2022-01-31
comments_async/serializers: add changing and deleting permission

Rineee:
Needed in the app..


sabinammm (approved):
Looks great! Tested this in a+ with different users.

fuzzylogic2000 (approved):
Yeah!

PR #964, opened 2022-01-31
comments_async/serializer & tests/comments: serializer treats blocked comments s…

### LINKED ISSUE #899: async serializer: also set is_deleted to True for blocked comments

sabinammm:
…ame as deleted comments

fixes #899
also fixes [#166 in kosmo](https://github.com/liqd/a4-kosmo/issues/166)


fuzzylogic2000 (commented):
Yes! That's nearly it.
The only thing left is the categories. In a debate, that is still shown for the blocked comment, but not the other two. Let me know if you have a hard time finding that!
Screenshot from 2022-02-01 11-35-08

adhocracy4/comments_async/serializers.py
```diff
@@ -40,7 +40,10 @@ def to_representation(self, instance):
         """
         ret = super().to_representation(instance)
         categories = {}
-        if ret['comment_categories']:
+        if ret['is_blocked']:
+            ret['comment_categories'] = categories
```
fuzzylogic2000, line 44: I don't really get, why you put it here and moved the if and made the other one to an `elif`.

fuzzylogic2000 (changes requested):
Cool, that you found where it happens. I would solve it a bit differently, though. Let's have a look together tomorrow.

fuzzylogic2000:
@sabinammm I removed the bit with the categories again and added another commit that changes it to returning the empty dict when the comment is blocked.

sabinammm:
> @sabinammm I removed the bit with the categories again and added another commit that changes it to returning the empty dict when the comment is blocked.

Ok, I see, thank you!!

PR #970, opened 2022-02-09
DEPRECATED (see PR #987):  comments/api-calls: removing axios ...

khamui:
…ls to api.js

### DIFF
--- a/adhocracy4/static/api.js
+++ b/adhocracy4/static/api.js
@@ -12,6 +12,18 @@ document.addEventListener('a4.embed.ready', init, false)
 
 const baseURL = '/api/'
 
+// This converts params = { ordering: "new" }
+// To ....endpoint/comments/?ordering=new
+const encodeQueryString = paramsObject => {
+  const params = Object.keys(paramsObject)
+  return params.reduce((acc, curr) => {
+    const symbol = acc === '' ? '?' : '&'
+    return paramsObject[curr]
+      ? `${acc}${symbol}${curr}=${paramsObject[curr]}`
+      : acc
+  }, '')
+}
+
 const api = (function () {
   const urls = {
     report: baseURL + 'reports/',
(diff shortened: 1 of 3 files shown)


adhocracy4/static/api.js
```diff
@@ -12,6 +12,18 @@ document.addEventListener('a4.embed.ready', init, false)
...
+// To ....endpoint/comments/?ordering=new
+const encodeQueryString = paramsObject => {
+  const params = Object.keys(paramsObject)
+  return params.reduce((acc, curr) => {
+    const symbol = acc === '' ? '?' : '&'
+    return paramsObject[curr]
+      ? `${acc}${symbol}${curr}=${paramsObject[curr]}`
+      : acc
+  }, '')
+}
```
phillimorland, line 25: yo sorry for dumb question, why do we want to do this? im toatlly crap with api stuff

phillimorland (commented):
looks good and works, I am still bad with API stuff so maybe @fuzzylogic2000 or @Rineee wanst to take a look

adhocracy4/static/api.js
khamui: ah yes, good point! this creates the querystring to sort and add filters to the api call.
Axios provides it builltin, fetch (which will be used in future) does not.

There are a couple of packages doing it, but i thought getting rid of axios and introducing something else does not help. So I have this little method to make it work. This is something we will need if we refactor this file to use fetch instead of jQuery.ajax().

adhocracy4/static/api.js
```diff
@@ -73,6 +85,30 @@ const api = (function () {
...
+      },
+      complete: function () {
+        $body.removeClass('loading')
+      }
+    }
+
+    $body.addClass('loading')
+    return $.ajax(defaultParams)
+  }
+
```
Rineee, line 111: Would there also be a way to use the other _sendRequest? I find it a bit confusing that we now here introduce another function, whereas we use _sendRequest for posting/changing comments.. But probably I dont get it? Happy to discuss!

khamui: as we would refactor this file at some point to get rid of jQuery's ajax() I thought it would be easier to "duplicate" this method for now.

Actually I was thinking either to pass the same request infos (e.g. baseUrl + contenttypeid + comment id) to use the _sendRequest. But that seems complicated.

The other way would be to make sure _sendRequest can work with api-urls directly but that would also complicate the _sendRequest. So I thought making its own method is the easiest way, no? @Rineee

Rineee: Yeah, no I think make _sendRequest work with api-urls directly does not really make sense.

I see your point that it is easier this ways, but tbh I would prefer using it the same way as with all other requests, because then we could also get rid of the commentApiUrl that is passed through template tag. And also for refactoring isnt it easier if we just have one way of sending requests and passing infos?

I can also look into it if you want!

khamui:
closing in favor of #987

PR #972, opened 2022-02-09
comments/jsx-translations: storing in variable

khamui:


adhocracy4/comments_async/static/comments_async/comment_box.jsx
```diff
@@ -18,6 +18,18 @@ const sorts = {
   dis: django.gettext('Last discussed')
```
khamui, line 18: the idea to store it in a dictionary I stole from here!
I like it because it wraps and gives all translations a namespace. What do you think?

phillimorland (approved):
super nice :)

PR #977, opened 2022-02-11
WIP: Kl 2022 refactor serializer

fuzzylogic2000:
still missing: the commenting permissions on the object, but also the comment


fuzzylogic2000:
@Rineee The last commit is obviously not yet done. The list method either needs to get in there in the right MRO, or better the Pagination thing needs to also get a super(). But this morning, my brain is not good and I will not find out now, where to get the other info needed. The current list method from the pagination thing redoes the original list method from the ListMixin: https://www.cdrf.co/3.12/rest_framework.mixins/ListModelMixin.html

fuzzylogic2000:
@Rineee @khamui It would be great if both of you could review. Probably easiest by going through the commits.

adhocracy4/comments_async/serializers.py
```diff
@@ -25,8 +25,7 @@ class CommentSerializer(serializers.ModelSerializer):
     class Meta:
         model = Comment
         read_only_fields = ('modified', 'created', 'id',
-                            'user_name', 'user_pk', 'user_image',
-                            'user_image_fallback', 'ratings',
+                            'user_name', 'user_pk', 'user_image', 'ratings',
```
Rineee, line 25: Could we also remove user_pk from here, since its not needed for isOwner() anymore?

fuzzylogic2000: Right! Yes, we don't use it for anything else? No, what should that be. :) So yes, please remove it!

adhocracy4/comments_async/serializers.py
```diff
@@ -58,6 +52,41 @@ def to_representation(self, instance):
...
+            ret['has_deleting_permission'] = user.has_perm(
+                'a4comments.delete_comment', instance
+            )
+            ret['has_moderating_permission'] = user.has_perm(
+                'a4comments.moderate_comment', instance
+            )
+            ret['has_comment_commenting_permission'] = user.has_perm(
+                'a4comments.comment_comment', instance
+            )
+
```
Rineee, line 89: It feels like we talked about this before, but can you explain to me again why we add all of these here instead of using own fields?

adhocracy4/comments_async/templatetags/react_comments_async.py
```diff
@@ -29,9 +28,6 @@ def react_comments_async(context, obj, with_categories=False):
     would_have_comment_permission = NormalUser().would_have_perm(
         permission, obj)
 
-    comments_contenttype = ContentType.objects.get_for_model(Comment)
-    pk = obj.pk
-
```
Rineee, line 35: Nice!

adhocracy4/comments_async/static/comments_async/comment.jsx
```diff
@@ -230,7 +230,7 @@ export default class Comment extends React.Component {
   }
 
   renderDeleteModal () {
-    if (this.props.is_users_own_comment || this.context.isModerator) {
+    if (this.props.has_deleting_permission) {
```
Rineee, line 233: So much better!

adhocracy4/comments_async/static/comments_async/comment_manage_dropdown.jsx
```diff
@@ -20,9 +20,11 @@ const CommentManageDropdown = (props) => {
         <i className="fas fa-ellipsis-h" aria-hidden="true" />
       </button>
       <div className="dropdown-menu dropdown-menu-end">
-        {(props.renderOwnerOptions || props.renderModeratorOptions) && [
+        {props.has_changing_permission && [
           <button key="1" className="dropdown-item" type="button" onClick={props.handleToggleEdit}>{translated.edit}</button>,
-          <div className="divider" key="2" />,
+          <div className="divider" key="2" />
+        ]}
+        {props.has_deleting_permission && [
```
Rineee, line 27: Also so much clearer, ha!

adhocracy4/comments_async/serializers.py
fuzzylogic2000: @Rineee Because we only need them in the representation, I thought it would be better this way. Instead of getting the user from the request in sooo many own fields, we do that only once, saving many lines of basically the same code.

adhocracy4/comments_async/templatetags/react_comments_async.py
```diff
@@ -39,17 +33,12 @@ def react_comments_async(context, obj, with_categories=False):
...
-                 and user is not participant or project is draft
-                 (negation of modules.predicates.is_allowed_comment_item)
     isContextMember - true if project is public or user is
                       participant/moderator/org_member
     '''
     attributes = {
         'subjectType': contenttype.pk,
         'subjectId': obj.pk,
-        'isReadOnly': (not has_comment_permission
-                       and not would_have_comment_permission),
```
Rineee, line 63: Love it, bye bye!

Rineee (approved):
Super awesome great refactor!! :tada: Especially with the permissions, everything is so much clearer!

adhocracy4/comments_async/serializers.py
Rineee: An idea could be to add them all in one field permissions?

fuzzylogic2000: Yeah, I also thought about that. That we don't use the to_representation if we do not really really need to. 
But I still think this is better than adding all the fields to the top and then again at the bottom as get method.

adhocracy4/comments_async/serializers.py
fuzzylogic2000: Ok, I will do that. :smile:

khamui:
Nice refactor (even if i don't follow 100 %) and sorry for reviewing it after it got merged :)
Just two things kind of related, worth mentioning/asking @fuzzylogic2000 :

1. While reading through your commits I saw `this.context` and `getChildContext`, which you removed here. But reading up on it I saw this is a legacy feature that will be removed at some point in future (https://reactjs.org/docs/legacy-context.html). Should we keep it in mind and add a task to scan our codebases to replace it? Tagging @phillimorland 

2. Out of curiosity: What does `wouldHaveCommentingPermission` mean? and how does it differ from `hasCommentingPermission`?

fuzzylogic2000:
> Nice refactor (even if i don't follow 100 %) and sorry for reviewing it after it got merged :) Just two things kind of related, worth mentioning/asking @fuzzylogic2000 :
> 
>     1. While reading through your commits I saw `this.context` and `getChildContext`, which you removed here. But reading up on it I saw this is a legacy feature that will be removed at some point in future (https://reactjs.org/docs/legacy-context.html). Should we keep it in mind and add a task to scan our codebases to replace it? Tagging @phillimorland

Yes! Let's add issues to the projects, we use that in? Then we see and do them when we have time.

> 
>     2. Out of curiosity: What does `wouldHaveCommentingPermission` mean? and how does it differ from `hasCommentingPermission`?

The `hasCommentingPersmission` checks on the authenticated user if she is allowed to do that. The 'wouldHaveCommentingPermission` takes the `NormalUser` (a hypothetical user that is only registered and does not have any other rights, roles or permissions) and checks if he has that permission. With that we can tell if someone is allowed to comment if logged in. 
We also use a `wouldHave...Permission` to show the idea create button if the user is not authenticated. https://github.com/liqd/a4-meinberlin/blob/4bdcae9999015c836321135da07c129baf080891/meinberlin/apps/ideas/templates/meinberlin_ideas/idea_list.html#L5

khamui:
ah thanks! i received it and even opened it but was distracted by another task

1.) okay sounds good! will go through the projects and add if i find it there.
2.) ah okay, not sure if I follow fully. So the NormalUser case is for instance for unregistered/public users?)

fuzzylogic2000:
>  2.) ah okay, not sure if I follow fully. So the NormalUser case is for instance for unregistered/public users?)

Yeah, with the would_have_permissions we check if a user would have that permission when she is only authenticated and nothing else. So, that we can display "add an idea" in that case and when that is clicked redirect to the login and after login the "add an idea" is still there and possible. To prevent that we display it and then when the user is logged in, the phase is over or only invited participants are allowed to take part. Does that make more sense @khamui ?

khamui:
now I got it. Thanks for explanation.

PR #987, opened 2022-02-15
Ks 2022 02 use internal api to fetch comments

Rineee:
There is still one axios call missing here: https://github.com/liqd/adhocracy4/blob/39acf0711aa7a614c596e2ee943f1528a8951ab7/adhocracy4/comments_async/static/comments_async/comment_box.jsx#L452

I wasnt sure how to handle it, because the url comes from pagination. I guess there are 2 options:
1. I could modify pagination that it doesnt return the url, but the page name and then pass that as data and use api.js
2. Or use fetch directly here if that is the long term plan anyway (is it? :))?

What do you think @khamui @phillimorland @fuzzylogic2000 ?

Update: as discussed I now use fetch directly for this.

### DIFF
--- a/adhocracy4/comments_async/static/comments_async/comment_box.jsx
+++ b/adhocracy4/comments_async/static/comments_async/comment_box.jsx
@@ -2,7 +2,6 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import django from 'django'
 import update from 'immutability-helper'
-import axios from 'axios'
 
 import CommentForm from './comment_form'
 import CommentList from './comment_list'
@@ -54,6 +53,10 @@ export default class CommentBox extends React.Component {
     this.handleHideReplyError = this.handleHideReplyError.bind(this)
     this.updateStateComment = this.updateStateComment.bind(this)
     this.handleToggleFilters = this.handleToggleFilters.bind(this)
+    this.urlReplaces = {
+      objectPk: this.props.subjectId,
+      contentTypeId: this.props.subjectType
+    }
 
     this.state = {
       comments: [],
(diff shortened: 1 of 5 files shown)


phillimorland (commented):
@Rineee this is looking good and working nicely, however as discussed this morning in standup I found this morning that fetch() may not be supported in IE11, although we do not officially 'have' to support it anymore, i'm now hesitent to break a major interactive element, @khamui will check it out and see if my first read was correct, sorry guys!!! :(

khamui:
@phillimorland @Rineee yes indeed `fetch` is not implemented in IE11. Sorry that I didn't check this before. we would have to decide what to do. So there are three options (probably more, but i will stick to the modern and production proven ones):

1.) Using fetch and taking in account IE11 users will have issues with comments
2.) Using axios again, because it is using XHR under the hood
3.) Using fetch-polyfill, which makes fetch working with older browsers

Another option would be to use XHR directly instead of having axios as wrapper. But this is not recommended, because the implementation would be very different. Axios makes XHR work with Promises, which is what we want to have to avoid `callback hell`. What do you think?

Tagging all, but if this annoys you let me know :) @fuzzylogic2000  @goapunk @sabinammm

khamui (changes requested):
Nice! Works fine.

Just one little thing: axios package can be removed from package.json.
Once that is done feel free to merge!

Rineee:
> @phillimorland @Rineee yes indeed `fetch` is not implemented in IE11.
> 
> ~Sorry that I didn't check this before. we would have to decide what to do. So there are three options (probably more, but i will stick to the modern and production proven ones):~
> 
> ~1.) Using fetch and taking in account IE11 users will have issues with comments~ ~2.) Using axios again, because it is using XHR under the hood~ ~3.) Using fetch-polyfill, which makes fetch working with older browsers~
> 
> ~Another option would be to use XHR directly instead of having axios as wrapper. But this is not recommended, because the implementation would be very different. Axios makes XHR work with Promises, which is what we want to have to avoid `callback hell`. What do you think?~
> 
> UPDATE: option 3.) is already implemented where needed (e.g. meinBerlin).

Cool, thanks for taking care of that @khamui @phillimorland !

PR #989, opened 2022-02-21
comments_async translation keys changed and adding

khamui:
distinguish between replying and commenting


fuzzylogic2000 (commented):
Looks good to me, but I cannot get it working. :woman_shrugging:

Rineee (approved):
Yeah, nice!

PR #992, opened 2022-02-22
[#5404] adhocracy4/projects: make slug field editable in the django admin

sabinammm:
Still testing this in projects


fuzzylogic2000 (commented):
Even though, we overwrite this in most projects, I think, the slug field should also be added here: https://github.com/liqd/adhocracy4/blob/b1b6a6a6f37996cc49bb876b1ea5a737b23fc5fc/adhocracy4/projects/admin.py#L63 So that we can change it in all projects if we want to.

sabinammm:
> Even though, we overwrite this in most projects, I think, the slug field should also be added here:
> 
> https://github.com/liqd/adhocracy4/blob/b1b6a6a6f37996cc49bb876b1ea5a737b23fc5fc/adhocracy4/projects/admin.py#L63
> 
> So that we can change it in all projects if we want to.

Yes, I'll add this field here as well.

PR #993, opened 2022-02-22
WIP: comments/date: i18n for dates of comments

khamui:
would it be possible to add the `locale`, e.g. `de-DE` or `ky-KG` (for Kirgistan) etc., to the api of the comments?
currently it falls back to german.  @fuzzylogic2000 @Rineee @sabinammm 

Just adding some screenshots, because testing this is a little time consuming. In case if the browser does not support a locale code, it falls back to numerical format (see below)

**safari with ky-KG**

safari_ky-KG

**chrome with ky-KG**

chrome_ky-KG

**chrome with de-DE**

chrome_de-DE

### DIFF
--- a/adhocracy4/api/mixins.py
+++ b/adhocracy4/api/mixins.py
@@ -3,6 +3,7 @@ from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.http import Http404
 from django.shortcuts import get_object_or_404
+from django.utils.translation import get_language
 from rest_framework import status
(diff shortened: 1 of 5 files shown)


khamui:
@Rineee cool, will try to add that.

Rineee:
> @Rineee cool, will try to add that.

@khamui Cool! If you add the LocaleMixin to CommentViewSet, make sure to add it before/above PaginationCommentLinkMixin, otherwise I think pagination might get broken..

PR #995, opened 2022-02-23
Ks 2022 02 refactor to representation in comment serializer

Rineee:


Rineee:
@fuzzylogic2000 You meant this, that I should also check, right? https://github.com/liqd/adhocracy4/blob/18aa99eb97c7d0c0f51582135d28cebaf93e6bca/adhocracy4/comments_async/static/comments_async/comment.jsx#L99 I dont really know how to change this.. you have an idea?

fuzzylogic2000:
> @fuzzylogic2000 You meant this, that I should also check, right?
> 
> https://github.com/liqd/adhocracy4/blob/18aa99eb97c7d0c0f51582135d28cebaf93e6bca/adhocracy4/comments_async/static/comments_async/comment.jsx#L99
> I dont really know how to change this.. you have an idea?

No, I also stumbled over this. The only thing, I can think of is that the comment knows if it is a child comment. We probably also use that elsewhere? But I think, it can stay like that for now.

fuzzylogic2000 (approved):
Supernice! Love it!

PR #996, opened 2022-02-24
WIP comments/date: format to long month and translating date in serializer

khamui:
@fuzzylogic2000 I found a solution, but still some things needs to be discussed I guess.

because the serializer makes the translation we have to set a specific format there.
the format is a custom one and therefore does not react on country specifics.

For instance "en-US" time would rather be 4:21 p.m., than 16:21. 

LANGUAGE_CODE="ky-KG" --> "ru-RU" is exactly the same
Bildschirmfoto 2022-02-24 um 14 47 29

LANGUAGE_CODE="en-US"
Bildschirmfoto 2022-02-24 um 14 47 58

LANGUAGE_CODE="fr-FR"
Bildschirmfoto 2022-02-24 um 14 48 27

LANGUAGE_CODE="it-IT"
Bildschirmfoto 2022-02-24 um 14 50 44


fuzzylogic2000 (approved):
Nice!

PR #998, opened 2022-02-28
comments_async/filter-search: shifting input handling into component

khamui:


khamui:
@phillimorland to remove the browser specific `cancel-button` i changed the input type to `<input type="text"...` instead of `<input type="search"...`. FF and IE are not supporting that pseudo-selector:

https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-search-cancel-button

So only our circled icon is shown now in all browsers.

phillimorland:
@khamui nice find! however as using the correct 'type' attribute is semantically correct I feel like we should try to find a css way to deal with this as we are trying to focus on a11y, here is an ideal example (apparently) https://a11y-style-guide.com/style-guide/section-forms.html#kssref-forms-search

khamui:
actually i checked if this is relevant for a11y and it seems the type of the input is not useful. It seems that the type=search only brings features, but not helping with the semantic. So basically we are turning it into a searchinput but then turn off those features because FF and IE don't support them.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search

so to make it useful for screenreaders I guess it is better to follow what your link describes? what do you think?

phillimorland (approved):
Nice! :)

PR #1003, opened 2022-03-01
comments_async/date: using '<time>element'

### LINKED ISSUE #918: Accessibility-relevant HTML-tags in JSX
**URL:** https://meinberlin-dev.liqd.net/ideas/2021-00628/ See DATES in comments from a4
**user:** any
**expected behaviour:** accessible date through html tag
**behaviour:** 
**important screensize:**
**device & browser:** 
**Comment/Question:** 
In reference to https://github.com/liqd/a4-meinberlin/issues/4017, is this also generally the case in the projects for html rendered by jsx-files that we are missing the accessibility-relevant html-tags @phillimorland, i.e. can you estimate the scope of this issue?  In the above case, it should be changed to include `<time datetime=".xxxx"></time> ` right?
Screenshot?

image

khamui:
it seems that currently only date in jsx is used in `comments_async`.
Not sure if sending two extra fields is the right way? I tried converting it on frontend
but that only works with locale set to 'de'.

fixes #918 

- [x] tested with aplus --> liqd/adhocracy-plus#1739
- [x] tested with mB --> liqd/a4-meinberlin#4231
- [ ] tested with opin

### DIFF
--- a/adhocracy4/comments_async/static/comments_async/comment.jsx
+++ b/adhocracy4/comments_async/static/comments_async/comment.jsx
@@ -315,7 +315,9 @@ export default class Comment extends React.Component {
                       : <a href={userProfile}>{this.props.user_name}</a>}
                   </div>
                   {moderatorLabel}
-                  <div className="a4-comments__submission-date">{lastDate}</div>
+                  <time className="a4-comments__submission-date">
+                    {lastDate}
+                  </time>
                 </div>
 
                 <div className="col-1 col-md-1 ms-auto a4-comments__dropdown-container">


phillimorland:
@khamui as discussed in the issue, we do not require a datetime attribute as we do not need it to be machine readable, however we can keep the `<time>` tag as it does not nee the `datetime` attribute and is nice and semantic so this PR can be edited to just replace the `<div>`  with `<time>`  does that make sense?
see here for ref https://html.spec.whatwg.org/#the-time-element

khamui:
sounds good, i am on it

PR #1006, opened 2022-03-04
projects/fields: initialize TopicField choices even if topics not pre…

Rineee:
…sent in settings

When this is merged, opin can be updated to newest A4.


Rineee:
I will merge this, so I can update the other projects, I hope that is fine :grimacing:

PR #1007, opened 2022-03-04
api/permissions: allow options requests in ViewSetRulesPermission

Rineee:
We currently get a 500 for OPTIONS requests, see e.g. here https://sentry.liqd.net/organizations/liqd/issues/1939/?project=2

So we should either handle them or disallow them, but I guess its ok to allow them, especially since we still check permissions. 

See here for django restframework docs of OPTIONS request: https://www.django-rest-framework.org/api-guide/metadata/


fuzzylogic2000 (approved):
Nice! I think, it's good to allow them!

PR #1013, opened 2022-03-11
adhocracy4/comments_async: display of search results text uses correc…

sabinammm:
…t count

This is necessary for Issue [#4254 in meinBerlin](https://github.com/liqd/a4-meinberlin/issues/4254)

### DIFF
--- a/adhocracy4/comments_async/static/comments_async/comment_box.jsx
+++ b/adhocracy4/comments_async/static/comments_async/comment_box.jsx
@@ -89,9 +89,6 @@ export default class CommentBox extends React.Component {
           translated.entries =
             django.ngettext('entry', 'entries', data.count)
 
-          translated.entriesFound =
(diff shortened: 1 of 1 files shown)


adhocracy4/comments_async/static/comments_async/comment_box.jsx
```diff
@@ -89,8 +89,8 @@ export default class CommentBox extends React.Component {
           translated.entries =
             django.ngettext('entry', 'entries', data.count)
 
-          translated.entriesFound =
-            django.ngettext('entry found for ', 'entries found for ', data.count)
+          console.log(data)
+          console.log(this.state.commentCount)
```
fuzzylogic2000, line 93: We don't need this anymore. :)

fuzzylogic2000 (changes requested):

fuzzylogic2000 (approved):
So cool!

PR #1014, opened 2022-03-11
apps/async_comments: make sure local time is used for comment dates

fuzzylogic2000:
will fix https://github.com/liqd/a4-meinberlin/issues/4256 once updated

@khamui Luckily someone already explained it in here: https://github.com/liqd/a4-meinberlin/blob/23922a3797db0e06b6de6cd1aa87bc73224cbc17/meinberlin/apps/contrib/templatetags/contrib_tags.py#L50


sabinammm (approved):
Perfect!!

PR #1039, opened 2022-04-06
comments/models: detect if comment text has changed and add update_fi…

Rineee:
…elds

Inspired by [this solution](https://stackoverflow.com/questions/1355150/when-saving-how-can-you-check-if-a-field-has-changed) :)

I realized that even if a child comment is added, the parent comment is again sent to api as the last_discussed field is changed. So I was looking for a solution to detect if the comment text has changed.

### DIFF
--- a/adhocracy4/comments/models.py
+++ b/adhocracy4/comments/models.py
@@ -56,6 +56,11 @@ class Comment(base.UserGeneratedContentModel):
         ordering = ('created',)
         index_together = [('content_type', 'object_pk')]
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._former_comment = transforms.clean_html_all(
+            self.comment)
+
     def __str__(self):
         if len(self.comment) > 200:
             return "{} ...".format(self.comment[:200])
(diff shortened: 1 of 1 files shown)


Rineee:
ahhh, will fix tests tomorrow

fuzzylogic2000:
> …elds
> 
> Inspired by [this solution](https://stackoverflow.com/questions/1355150/when-saving-how-can-you-check-if-a-field-has-changed) :)
> 
> I realized that even if a child comment is added, the parent comment is again sent to api as the last_discussed field is changed. So I was looking for a solution to detect if the comment text has changed.

Right! I was confused, because the `last_discussed` does not change the modified field. (https://github.com/liqd/adhocracy4/blob/2cb93f01c8b237e5abc7587cc99f6798eb9be1c4/adhocracy4/comments/signals.py#L17). But that will obviously still send a signal. :roll_eyes: 

Cool solution for checking if it was really the comment-text that was changed!

Rineee:
@fuzzylogic2000 Alright, once again the tests were right! The problem was with removed or censored comments, I now changed it to not send update_fields if comment was removed or censored, as we dont need to check the empty string, I guess. Also: the transform_clean_html was missing when saving the _former_comment..

fuzzylogic2000 (approved):
Nice!

PR #1044, opened 2022-04-12
downgrade dependencies for compatibility with react 17.x

goapunk:
* downgrade dependencies for compatibility with react 17.x
* move shared dependencies into peerDependencies to prevent version conflicts/chaos between a4 and a+


package.json
```diff
@@ -15,38 +15,52 @@
...
     "eslint-plugin-react": "7.29.4",
     "eslint-plugin-standard": "4.1.0",
     "husky": "7.0.4",
+    "immutability-helper": "3.1.1",
     "jest": "27.5.1",
-    "lint-staged": "12.4.1"
+    "leaflet": "1.7.1",
+    "jquery": "3.6.0",
+    "leaflet-draw": "1.0.4",
+    "leaflet.markercluster": "git+https://github.com/liqd/Leaflet.markercluster#5ed89b26922c51083fc9632a2c01425b9261a0f5",
```
goapunk, line 41: @phillimorland  we use our fork in a+ but the upstream repo in a4, was there a reason for that? I changed it to the fork as well for now

goapunk:
this unfortunately duplicates the dependencies because for the tests to work they need to be in `devDependencies`.

package.json
```diff
@@ -15,38 +15,52 @@
...
+    "react-markdown": "8.0.3",
+    "react-popper": "2.2.5",
+    "react-slick": "0.28.1",
+    "shpjs": "4.0.2",
+    "slick-carousel": "git+https://github.com/liqd/slick.git#pm-2019-07-overwrites"
+  },
+  "peerDependencies": {
+    "immutability-helper": "3.1.1",
+    "jquery": "3.6.0",
+    "leaflet": "1.8.0",
```
goapunk, line 54: a4 uses 1.7.1 but a+ 1.8.0 so I set it to 1.8.0 as well

package.json
phillimorland: @goapunk only thing I can think of is a+ have specific styling but that should be in this package

fuzzylogic2000 (approved):
Cool! It works!

PR #1045, opened 2022-04-14
apps/comments_async/moderator_statement.jsx: adding new moderator sta…

khamui:
…tement component and integrate into comments

### DIFF
--- a/adhocracy4/comments_async/static/comments_async/comment.jsx
+++ b/adhocracy4/comments_async/static/comments_async/comment.jsx
@@ -12,6 +12,7 @@ import CommentCategoryEditForm from './comment_category_edit_form'
 import CommentForm from './comment_form'
 import CommentManageDropdown from './comment_manage_dropdown'
 import CommentList from './comment_list'
+import { ModeratorStatement } from './moderator_statement'
 
 import { RatingBox } from '../../../ratings/static/ratings/react_ratings'
 
@@ -24,7 +25,9 @@ const translated = {
   readMore: django.gettext('Read more...'),
   readLess: django.gettext('Read less'),
   share: django.gettext('Share'),
-  report: django.gettext(' Report')
+  report: django.gettext(' Report'),
+  showModStatement: django.gettext('Show moderator\'s statement'),
+  hideModStatement: django.gettext('Hide moderator\'s statement')
 }
 
 function getAnswerForm (hide, number) {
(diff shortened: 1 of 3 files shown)


adhocracy4/comments_async/static/comments_async/moderator_statement.jsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <div className="row">
+      <div className="a4-comments__moderator-statement__container">
+        <i className="fas fa-share a4-comments__moderator-statement__icon" />
+        <div className="a4-comments__moderator-statement__content">
+          <div className="a4-comments__moderator-statement__moderator">
+            {translated.moderator}
+          </div>
+          <div className="a4-comments__moderator-statement__text">
+            {lastEdit}
```
khamui, line 16: do we want to get it from backend like we did similarily here?
https://github.com/liqd/adhocracy4/pull/996/commits/c0af4f052ca5e726db97cf4128568a713629e66b

the other way would reintroduce the javascript `localeString` thingy, but I guess we rather want to have once consistent way how we make dates localized and timzoned?

adhocracy4/comments_async/static/comments_async/comment.jsx
```diff
@@ -24,7 +25,9 @@ const translated = {
   readMore: django.gettext('Read more...'),
   readLess: django.gettext('Read less'),
   share: django.gettext('Share'),
-  report: django.gettext(' Report')
+  report: django.gettext(' Report'),
+  showModStatement: django.pgettext('kosmo', 'Show moderator\'s statement'),
+  hideModStatement: django.pgettext('kosmo', 'Hide moderator\'s statement')
```
fuzzylogic2000, line 30: Uh, ehm, actually, I think, as this will also end up an a+ and thus in kosmo in the locale-source branch, we should not mark this as translation from komo. While getting the translations will be a longer way now, I think that will save us from confusion at a later point. 
So, let's:
1. remove the stuff marking it for kosmo
2. Also update a4 in a+, push strings to transifex, translate, pull. (That way we will get the translations with the next rebase.)
3. Probably test this story without translations, but add a reminder about that to the release story

fuzzylogic2000 (commented):
Yes, looks good! Will try it out in kosmo and have a small request that makes translating muuuuch harder, but will hopefully help us in the long run.

adhocracy4/comments_async/static/comments_async/moderator_statement.jsx
fuzzylogic2000: Ups, I didn't see your comment. But yes! Let's do it the same way!

fuzzylogic2000 (commented):
Nice! But can you please squash the first and second commit? In the history it will be easier to read and understand, when there is only one commit changing something.

goapunk (approved):
very nice! and even with js test :)

fuzzylogic2000 (approved):
Supercool!

PR #1052, opened 2022-04-21
apps/comments_async/moderator_statement.jsx: fix passing props when u…

khamui:
…sing component and in js tests


phillimorland (approved):
very nice :)

PR #1053, opened 2022-04-21
jstests: adding tests to components Alert.jsx and ErrorList.jsx

khamui:


fuzzylogic2000:
@goapunk @khamui But we can merge? And you add more tests later? Or we wait and you add them here?

goapunk:
@fuzzylogic2000 fine with me to merge and add tests later/separately

khamui:
@fuzzylogic2000 yes lets merge then, thought it would be fast, but with `timeouts` in tests i need to research a bit.

PR #1056, opened 2022-04-22
Alert: run timer in useEffect to prevent multiple triggers

goapunk:


adhocracy4/static/Alert.jsx
```diff
@@ -1,11 +1,17 @@
-import React from 'react'
+import React, { useRef, useEffect } from 'react'
 import django from 'django'
 
 const Alert = ({ type, message, onClick, timer }) => {
+  const timer0 = useRef()
```
khamui, line 5: maybe a different naming instead of timer0? 

for instance the argument `timeInMs` and the actual ref `timer`?

khamui (approved):
Good stuff @goapunk ! just a tiny little thing about the naming:

adhocracy4/static/Alert.jsx
goapunk: yeah, good idea! `timer` and `timer0` is not a good choice

khamui: i know we would need to change it where we use it. or you "rename" it here for instance ({ type, message, onClick, timer: timeInMs }). I admit its sketchy.

goapunk: we only use it in kosmo, right? I can change it there

goapunk: opened https://github.com/liqd/a4-kosmo/pull/203 for it

khamui: perfect thanks!

PR #1057, opened 2022-04-22
tests/factories: move FollowFactory to be used in projects

fuzzylogic2000:


goapunk:
looks good and makes more sense there!

PR #1065, opened 2022-04-26
test/factories: move follow factory to own file

fuzzylogic2000:
I don't know why I didn't do it like this in the first place. Now also works in again in projects that don't have the a4 follows installed (like opin).


goapunk (approved):
nice fix, looks good to me!

PR #1068, opened 2022-04-29
modules/predicates: make some predicates work for projects instead of…

Rineee:
… items

I needed this for the moderation comments permission as we don't have a module there, so the permission object (for creating) needs to be the project.


fuzzylogic2000 (approved):
Ah, I needed a while to get why you need that. But ok! :)

fuzzylogic2000:
Right! You should add it to the tests, though. I think we do test the predicates and rules quite extensively so far. You can do it in this PR or add a task and merge this one.

Rineee:
> Right! You should add it to the tests, though. I think we do test the predicates and rules quite extensively so far. You can do it in this PR or add a task and merge this one.

@fuzzylogic2000 Voilà!

fuzzylogic2000 (approved):
Nice!

PR #1074, opened 2022-05-09
comments/async: check setting if moderator_marked is used and add to …

Rineee:
…sorting filter if it is

### DIFF
--- a/adhocracy4/comments_async/static/comments_async/comment_box.jsx
+++ b/adhocracy4/comments_async/static/comments_async/comment_box.jsx
@@ -62,7 +62,7 @@ export default class CommentBox extends React.Component {
       showFilters: false,
       filter: 'all',
       filterDisplay: django.gettext('all'),
-      sort: 'new',
+      sort: props.useModeratorMarked ? 'mom' : 'new',
       loading: true,
       search: '',
       anchoredCommentId: props.anchoredCommentId ? parseInt(props.anchoredCommentId) : null,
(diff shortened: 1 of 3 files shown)


adhocracy4/comments_async/static/comments_async/comment_box.jsx
```diff
@@ -62,7 +62,7 @@ export default class CommentBox extends React.Component {
       showFilters: false,
       filter: 'all',
       filterDisplay: django.gettext('all'),
-      sort: 'new',
+      sort: props.useModeratorMarked ? 'mom' : 'new',
```
fuzzylogic2000, line 65: So, this is also the default filter when it is used?

fuzzylogic2000 (commented):
Nice!
But the tests are failing...

adhocracy4/comments_async/static/comments_async/comment_box.jsx
Rineee: Yes. I thought we can also change that once we use it somewhere else, where it is not supposed to be default? That would be another setting, I guess? Or the setting would be a tuple?

Rineee: On the tests btw..

fuzzylogic2000: Yeah, we'll worry about that when we get there. It might just be default and have the same wording anyway. :)

fuzzylogic2000 (approved):
Whoop!

PR #1075, opened 2022-05-09
deps: rm remark-gfm as uneeded, comments_async/comment: add <mark> ta…

phillimorland:
…g to moderator marked comments

### DIFF
--- a/adhocracy4/comments_async/static/comments_async/comment.jsx
+++ b/adhocracy4/comments_async/static/comments_async/comment.jsx
@@ -2,7 +2,6 @@ import React from 'react'
 import ReactMarkdown from 'react-markdown'
 import PropTypes from 'prop-types'
 import django from 'django'
-import gfm from 'remark-gfm'
 
 import Modal from '../../../static/Modal'
 import { ReportModal } from '../../../reports/static/reports/react_reports'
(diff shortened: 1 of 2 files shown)


adhocracy4/comments_async/static/comments_async/comment.jsx
```diff
@@ -201,13 +200,21 @@ export default class Comment extends React.Component {
       }
       comment = (
         <div className={'a4-comments__text' + (this.state.anchored ? ' a4-comments__text--highlighted' : '')}>
```
phillimorland, line 202: we should change this to a4-comments__text--linked as that what the styling actually shows but we would need to change it everywhere so not sure about doing now?!

adhocracy4/comments_async/static/comments_async/comment.jsx
```diff
@@ -201,13 +200,21 @@ export default class Comment extends React.Component {
...
-          <ReactMarkdown
-            disallowedElements={['h1', 'h2', 'h3', 'h4', 'input', 'table', 'thead', 'tr', 'th']}
-            unwrapDisallowed
-            plugins={[[gfm, { singleTilde: false }]]}
-          >
-            {content}
-          </ReactMarkdown>
+          {this.props.is_moderator_marked
+            ? <mark>
```
khamui, line 204: this seems to highlight/mark the entire line in cases where the comment is shorter. Is this what we want?

phillimorland: nope but really not sure how else to do it with the markdown, did have a play yesterday but it's not a big story, also I imagine a moderator marked comment that is amazing probable will be longer then 2-3 words hopefully

phillimorland: it's not ideal I know!

PR #1079, opened 2022-05-11
comments/serializers: adapt update method of CommentModerateSerialize…

Rineee:
…r to fixed version

There was a bugfix in the rest_framework.serializers.ModelSerializer.update():
https://github.com/encode/django-rest-framework/commit/6a95451d72c2e1a87e0c5c1d70b2c21e7daa70ea

Since we copied the old method over, this needed to updated here to get the bugfix.


fuzzylogic2000 (approved):
Nice!

PR #1083, opened 2022-05-18
comments_async: remove bs padding class as not all projects use and m…

phillimorland:
…akes btn too small


fuzzylogic2000 (approved):
Cool!

PR #1084, opened 2022-05-19
exports/mixins/base: add option to specify which attributes of a rela…

Rineee:
…ted field should be exported. 

Will add tests later!


fuzzylogic2000 (commented):
It does look good! But where can I try it out? Quickly scanning the exports of a+, I couldn't find anything to replace. Did you have somewhere to try it or another PR already using it?

@fuzzylogic2000 Yes, sorry! It is used here now: https://github.com/liqd/a4-kosmo/pull/242

PR #1094, opened 2022-06-01
comments_async//polls/api: add orga terms of use info to response

Rineee:
@fuzzylogic2000 Not super happy with this double code, but since it is once added to retrieve and once to list, could not really put it into a mixin.. 

@khamui I hope this is what we discussed. It gives you
- use_org_terms which comes from the settings, should only be true for aplus
- user_has_agreed: null if anonymous user
                                true/false if logged in and has/has not agreed


adhocracy4/comments_async/api.py
```diff
@@ -153,3 +152,23 @@ def destroy(self, request, *args, **kwargs):
                              instance=comment)
         serializer = self.get_serializer(comment)
         return Response(serializer.data)
+
+    def list(self, request, *args, **kwargs):
+        """Add organisation terms of use info to response.data."""
+        response = super().list(request, args, kwargs)
+        if response.status_code == 400:
+            return response
```
fuzzylogic2000, line 160: Is this for the pagination in combination with linking to a comment? That works still fine, then?

fuzzylogic2000 (commented):
I think it is fine that we add it in different places and twice for the comments and polls. They are very different.

adhocracy4/polls/api.py
```diff
@@ -33,6 +34,26 @@ def get_permission_object(self):
...
+    def retrieve(self, request, *args, **kwargs):
+        """Add organisation terms of use info to response.data."""
+        response = super().retrieve(request, args, kwargs)
+        if response.status_code == 400:
+            return response
+
+        use_org_terms_of_use = False
+        if hasattr(settings, 'A4_USE_ORGANISATION_TERMS_OF_USE'):
+            user_has_agreed = None
+            use_org_terms_of_use = settings.A4_USE_ORGANISATION_TERMS_OF_USE
```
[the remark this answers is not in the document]
Rineee: Yes, true! I had it like that before and dont remember why I changed it, but that should work! Will change it back!

adhocracy4/comments_async/api.py
Rineee: @fuzzylogic2000 I added that, because a test was failing. It failed, because for a bad request the response is empty, but I was still trying to add something to it. So I copied that from the PermissionInfoMixin above.. I dont think that changes anything for pagination and comment linking, right? Or what do you mean exactly?

khamui:
nice!! working fine

PR #1095, opened 2022-06-01
apps/comments_async/checkbox terms of use: adding new component and m…

khamui:
…ake it hidable

- [x] to be adjusted once backend data is available
- [ ] make other checkboxes disappear after one was accepted (after submission)

### DIFF
--- a/adhocracy4/comments_async/static/comments_async/comment_form.jsx
+++ b/adhocracy4/comments_async/static/comments_async/comment_form.jsx
@@ -2,6 +2,7 @@ import React from 'react'
 import django from 'django'
 
 import CategoryList from './category_list'
+import { TermsOfUseCheckbox } from '../../../static/TermsOfUseCheckbox'
 
 import * as config from '../../../static/config'
 import Alert from '../../../static/Alert'
@@ -20,7 +21,6 @@ const translated = {
 export default class CommentForm extends React.Component {
   constructor (props) {
     super(props)
-
     this.state = {
       comment: '',
       commentCharCount: 0,
@@ -87,6 +87,15 @@ export default class CommentForm extends React.Component {
   render () {
     const textareaStyle = { height: (this.state.textareaHeight) + 'px' }
     const hasParent = this.props.parentIndex !== undefined
+    const actionButton = (
+      <button
+        type="submit"
+        className="btn a4-comments__submit-input ms-auto"
+        disabled={this.props.useTermsOfUse && !this.props.agreedTermsOfUse && !this.state.checkedTermsOfUse}
+      >
+        {translated.post}
+      </button>
+    )
 
     if (this.props.hasCommentingPermission) {
       return (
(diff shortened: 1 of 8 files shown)


adhocracy4/comments_async/static/comments_async/comment_form.jsx
```diff
@@ -87,6 +87,18 @@ export default class CommentForm extends React.Component {
   render () {
     const textareaStyle = { height: (this.state.textareaHeight) + 'px' }
     const hasParent = this.props.parentIndex !== undefined
+    const actionValue = this.props.commentCategoryChoices
+      ? translated.post
+      : translated.comment
```
Rineee, line 92: @khamui I know you did't introduce this, but do you know what it is used for? I mena, the different action value?

Rineee:
Yes, cool, that works! But frontend tests are failing ..

adhocracy4/comments_async/static/comments_async/comment_form.jsx
```diff
@@ -112,12 +124,24 @@ export default class CommentForm extends React.Component {
...
+                  <div className="a4-comments__submit d-flex col-6">
+                    {actionButton}
+                  </div>)
+                : (
+                  <TermsOfUseCheckbox
+                    id={`terms-of-use-checkbox-${typeof this.props.parentIndex === 'number' ? this.props.parentIndex : 'rootForm'}`}
+                    onChange={val => this.setState({ agreedTermsOfUse: val })}
+                    actionButton={actionButton}
+                  />
+                  )}
```
Rineee, line 144: Would it also work to remove the button from the TermsOfUseCheckbox component and always show the same actionButton here that is disabled if useTermsOfUse and not agreedTermsOfUse? And then only have the TermsOfUseCheckbox after the condition `!this.props.useTermsOfUse || this.props.agreedTermsOfUse` ?

khamui: the problem here was rather the position of the button. in case of terms of use being used, it is not on the same level as the counter `.../4000`. does that make sense @Rineee ?

adhocracy4/comments_async/static/comments_async/comment_form.jsx
khamui: not really sure, i will double check that

adhocracy4/comments_async/static/comments_async/comment_form.jsx
phillimorland: @khamui @Rineee I think simplifying the js here should take priority, to solve the level issue I think we can take `col-6` class from char count label and make container div with that class for the char count and checkbox, and leaving the action button out of it. This is better practice too then having a structure class on a col. Dies that make sense?

adhocracy4/comments_async/static/comments_async/comment_form.jsx
Rineee: Yes... but why? Is that maybe just some leftover? Could you try and take it out and check if it still works (with and without categories)?

khamui: hm, do you think it is enough to verify? because taking it out here, can break in another project where it might be used?

So if the flag `withCategories` is false:
Bildschirmfoto 2022-06-07 um 12 05 38

if its true:
Bildschirmfoto 2022-06-07 um 12 04 39

I don't know when this flag is set (or in which projects?). But i am hesitating taking this out and just verify in aplus.

Rineee: Well the code for the comments is all in A4, so it should be fine to test it in one project..
I just don't get why the value of the button changes depending on `withCategories`? But anyway, I can also check that!

khamui:
> Supercool, thanks for also deleting the button value stuff...
> 
> I have one question: both in comments and polls, you did not add an initial state for the checkedTermsOfUse. I was just wondering if that was on purpose? But it all works fine, so will merge now!

yes actually I did it on purpose, because I thought in most projects (except of aplus for now) this flag remains false and will never change to true. But also not totally sure if making this less obvious is worth the saving of a state variable.

PR #1096, opened 2022-06-02
babel.config: add peer deps as externals

phillimorland:


khamui:
merging and testing afterwards (this might fix, why i can't build a4 locally :)

PR #1101, opened 2022-06-08
comments/wording: change wording for moderators statement/feedback fo…

fuzzylogic2000:
…r comments (used in kosmo)

We decided to have this wording in the next kosmo release. But to get proper translations in a+ transifex and not to forget about it, it's already here!


goapunk (approved):
nice

PR #1102, opened 2022-06-08
apps/projects: increase max length of tile image name to fix bplan is…

fuzzylogic2000:
…sue in mB (https://github.com/liqd/a4-meinberlin/issues/4356)

It will appear in the bplans error messages automatically because it gets the length from a4
https://github.com/liqd/a4-meinberlin/blob/e8b129ec7daac2890af6e76b95269a78c5a707d8/meinberlin/apps/bplan/serializers.py#L42


adhocracy4/projects/models.py
```diff
@@ -241,9 +242,11 @@ class Project(ProjectContactDetailMixin,
         ),
         upload_to='projects/tiles',
         blank=True,
-        max_length=300)
+        max_length=512
+    )
```
phillimorland, line 246: will check this again mB tiles in morning as if it's same test then might break styling

fuzzylogic2000: It's the name of the image, which is hopefully not displayed anywhere...

phillimorland (approved):
looks good, doesn't work with mB because of node but looking fine on a+

fuzzylogic2000:
> looks good, doesn't work with mB because of node but looking fine on a+

Yeah, we need to cherry-pick this into an extra branch to actually get it into mB without updating all a4.

PR #1105, opened 2022-06-13
Jd 2022 06 bleach 5.x

goapunk:
* add some basic tests for transforms
* upgrade bleach to 5.0.0


goapunk:
There's been a "breaking change" in Bleach:  removing a block-level element will now insert a newline. I assume that's fine for our use case

fuzzylogic2000 (approved):
Very nice!

PR #1107, opened 2022-06-13
Ks 2022 06 save user agreement

Rineee:
We still need to handle the update of the user agreement over the different components.. not sure how to do this best? But probably we therefore need to refetch the comments after updating instead of just changing state.. :woman_shrugging: 

You can test it here https://github.com/liqd/adhocracy-plus/pull/1882


fuzzylogic2000 (approved):
Woohoo! It works!

PR #1108, opened 2022-06-13
Pm 2022 06 react 18 update

phillimorland:


adhocracy4/comments/static/comments/react_comments.jsx
```diff
@@ -1,8 +1,11 @@
...
+import { createRoot } from 'react-dom/client'
+
 const CommentBox = require('./CommentBox')
-const React = require('react')
-const ReactDOM = require('react-dom')
 
 module.exports.renderComment = function (el) {
   const props = JSON.parse(el.getAttribute('data-attributes'))
-  ReactDOM.render(<CommentBox {...props} />, el)
+  const container = el
```
khamui, line 8: we could reduce one variable and either rename the parameter to `container` or using `el`?

adhocracy4/comments_async/static/comments_async/comment_box.jsx
```diff
@@ -209,12 +215,16 @@ export default class CommentBox extends React.Component {
...
 
+    // flushSync stops react18 batching state updates and ensures
+    // re-render when no error
     return api.comments.change(modifiedComment, comment.id)
       .done(changed => {
-        this.updateStateComment(
-          index,
-          parentIndex,
-          changed)
+        flushSync(() => {
```
khamui, line 222: Do we really need this? is the automatic batching problematic here?

phillimorland: yeah, it's what I spent a day trying to fix :p without it it doesn't re-render when you edit comments or categories. 
Other option is instead of having 2 set states we combine them, not sure what happens then with errors and re-render but could test. Also thought having this was a good reminder about the batching behaviour

adhocracy4/comments/static/comments/react_comments.jsx
phillimorland: will do!

khamui:
very nice!

PR #1110, opened 2022-06-13
validators/validate_image(): adding optional max_resolution for images

khamui:


adhocracy4/images/validators.py
```diff
@@ -44,6 +44,12 @@ def validate_image(
         msg = _('Image must be at least {min_height} pixels high')
         errors.append(ValidationError(msg.format(min_height=min_height)))
 
+    if max_resolution:
+        max_width, max_height = max_resolution
+        if image_height > max_height:
+            msg = _('Image must be at most {max_height} pixels high')
+            errors.append(ValidationError(msg.format(max_height=max_height)))
+
```
Rineee, line 52: Could you maybe also add the validation for the width? I know, its not really necessary, but just to have it the same way as for min resolution..

fuzzylogic2000 (approved):
NIce! I will add some tests.

PR #1114, opened 2022-06-14
comment_box: update setstates and include flushsync where required

phillimorland:


phillimorland:
also corrected linting errors requesting `property shorthand`

PR #1119, opened 2022-06-15
add custom tos agreed event

goapunk:
add event which indicates that the terms of use have been accepted to other components

### DIFF
--- a/adhocracy4/polls/assets/PollQuestions.jsx
+++ b/adhocracy4/polls/assets/PollQuestions.jsx
@@ -38,6 +38,8 @@ class PollQuestions extends React.Component {
       loadingPage: true
     }
 
+    this.handleTermsOfUse = this.handleTermsOfUse.bind(this)
+
     this.linkToPoll = (
       <button type="button" className="btn poll__btn--link" onClick={() => this.handleToggleResultsPage()}>
         {django.gettext('To poll')}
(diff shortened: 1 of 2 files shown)


phillimorland (commented):
I know this is unfinished but checked it, code looks good but i'm still getting the same behaviour that after f5 refresh the checkbox appears again but thinking it might be my set up?! @khamui can you check this?

khamui:
for me it works, also with refreshing afterwards! 👏

adhocracy4/polls/assets/PollQuestions.jsx
```diff
@@ -38,6 +38,8 @@ class PollQuestions extends React.Component {
       loadingPage: true
     }
 
+    this.handleTermsOfUse = this.handleTermsOfUse.bind(this)
```
khamui, line 41: maybe this can be omitted if you write it as arrow function expression on line 194?

khamui (commented):
very cool! working fine

adhocracy4/polls/assets/PollQuestions.jsx
goapunk: hmm, I'd keep it like this till we rewrite it in functional as we do the same in the other components

goapunk:
needs #1120 to be merged first

goapunk:
@phillimorland Just rebased and unwipped as it was depending on the revert of the react update

phillimorland (approved):
Looking good, linting we can fix when update configs to actually use all our linting packages

PR #1121, opened 2022-06-20
apps/projects: decrease filename length of tile image

fuzzylogic2000:
- undo 2389b15ee834634bdc0bbdb114cc7d84aa7cdb13


goapunk (approved):
we need this migration for dev, right?

fuzzylogic2000:
> we need this migration for dev, right?

@goapunk Yes, I thought about reverting and deleting the old migration, but thought as this is a4, that's too dangerous and might result in weird stuff. So far, it should only be on a+ dev, though. But still.

PR #1122, opened 2022-06-20
apps/comments async: add link to organisation's terms of use to label

fuzzylogic2000:


adhocracy4/static/__tests__/TermsOfUseCheckbox.jest.jsx
```diff
@@ -1,10 +1,11 @@
 import React from 'react'
 import { render, screen, fireEvent } from '@testing-library/react'
 import { TermsOfUseCheckbox } from '../TermsOfUseCheckbox'
+import django from '../../../../../__mocks__/djangoMock'
 
 test('One TermsOfUseCheckbox is showing', () => {
   render(<TermsOfUseCheckbox id="test-checkbox" />)
-  const component = screen.getByLabelText(/mock text/)
+  const component = screen.getByLabelText(django.interpolate(), /mock text/)
```
phillimorland, line 8: because the test is getting the label text and we now use interpolate in it, it needed a django mock to be able to run the test. Is that right @khamui @goapunk ? i'm still bit confused by them myself?!

phillimorland:
styling for helptext https://github.com/liqd/adhocracy-plus/pull/1934

adhocracy4/static/TermsOfUseCheckbox.jsx
```diff
@@ -25,9 +31,11 @@ export const TermsOfUseCheckbox = (props) => {
       <label
         className="a4-termsofuse__checkbox-label"
         htmlFor={props.id}
-      >
-        {translated.termsOfUse}
-      </label>
+        dangerouslySetInnerHTML={{ __html: termsOfUseLabel }}
```
khamui, line 34: just read about this prop. Seems to be fine to use in this case

khamui (approved):
Looking good to me!

adhocracy4/static/TermsOfUseCheckbox.jsx
fuzzylogic2000: As we set it, I think it is fine!

fuzzylogic2000 (commented):
I cannot see the checkbox in the category edit form? It is there for the category form, though!

phillimorland:
@fuzzylogic2000 just tested it, was sure it was there yesterday :p will investigate!

fuzzylogic2000 (commented):
Very nice! Works! 
Now it is still missing from editing a reply. :roll_eyes: Our comments are bit crazy. But that can be another fix, so that we can use and merge your PR in a+ with this!

PR #1123, opened 2022-06-20
apps/comments async: save agreement in comment update

fuzzylogic2000:


fuzzylogic2000:
@phillimorland I guess, this is also missing from the category edit form. Whichever PR is merged first, we can rebase the other and add it.

phillimorland (approved):
nice! will rebase the PR i'm omn and add to categories

PR #1128, opened 2022-06-21
comments_async: update logic for termsofuse checkbox to ensure it's s…

phillimorland:
…hown on edit

### DIFF
--- a/adhocracy4/comments/static/comments/CommentBox.jsx
+++ b/adhocracy4/comments/static/comments/CommentBox.jsx
@@ -26,7 +26,7 @@ class CommentBox extends React.Component {
       diff[index] = { $merge: updatedComment }
     }
     comments = update(comments, diff)
-    this.setState({ comments: comments })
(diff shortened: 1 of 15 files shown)


fuzzylogic2000 (commented):
It still does not show the checkbox to me, when I edit a reply. But it changes the state and then does not show the checkbox at all (also in the create forms) anymore when I update the reply. It does not save the agreement, though. Reloading the comments shows the checkbox again in the create forms.

phillimorland:
@fuzzylogic2000 what??? so no checkbox on reply edit and no checkbox on comment??? I'm seeing this right now: will try a reinstall and link though as it was being very slow, maybe I removed something before commit.
Screenshot from 2022-06-21 15-43-28

phillimorland:
sorry ran lint fix and it did a whole bunch of other crap but that should now work

fuzzylogic2000 (approved):
Nice!

PR #1130, opened 2022-06-27
WIP: make sure comment and poll can only be added with egreement and…

fuzzylogic2000:
… add tests


adhocracy4/polls/api.py
```diff
@@ -93,6 +113,10 @@ def vote(self, request, pk):
                             self.request.data['agreed_terms_of_use']
                     }
                 )
+            else:
+                raise ValidationError({
```
fuzzylogic2000, line 117: For the poll, it does not matter which error, I use, it always gives "Your answer could not be saved. Please check the data you entered again.". I will add an issue for that, but would not like to handle that in this PR.

fuzzylogic2000:
> @fuzzylogic2000 but the tests are working now?

@goapunk Yes! I had to patch the reverse that is imported into the api. Which does make sense, but took me a while...

adhocracy4/polls/api.py
```diff
@@ -42,23 +43,46 @@ def rules_method_map(self):
...
+                    }
+                )
+            except NoReverseMatch:
+                raise NotImplementedError('Add org terms of use view.')
             if hasattr(request, 'user'):
                 user = request.user
                 if user.is_authenticated:
-                    user_has_agreed = \
-                        user.has_agreed_on_org_terms(organisation)
+                    user_has_agreed = self._user_has_agreed(user)
```
Rineee, line 85: see comment in comments

adhocracy4/comments_async/api.py
```diff
@@ -192,16 +217,18 @@ def list(self, request, *args, **kwargs):
...
+                        'organisation_slug': organisation.slug
+                    }
+                )
+            except NoReverseMatch:
+                raise NotImplementedError('Add org terms of use view.')
             if hasattr(request, 'user'):
                 user = request.user
                 if user.is_authenticated:
-                    user_has_agreed = \
-                        user.has_agreed_on_org_terms(organisation)
```
Rineee, line 204: Maybe we could still use user.has_agreed_on_org_terms here? Because we already have the organisation object here and then dont have to get it again in self._user_has_agreed.

tests/comments/test_async_api_org_terms.py
```diff
@@ -0,0 +1,320 @@
...
+    assert response.status_code == status.HTTP_201_CREATED
+    assert response.data['comment'] == 'comment comment'
+
+
+@pytest.mark.django_db
+@override_settings(
+    A4_USE_ORGANISATION_TERMS_OF_USE=True
+)
+def test_agreement_fields_not_implemented(
+        apiclient, question_ct,
```
Rineee, line 261: Can you explain to me what this test does? Somehow dont get it, sorry..

tests/comments/test_async_api_org_terms.py
```diff
@@ -0,0 +1,320 @@
...
+        organisation=question.module.project.organisation,
+        has_agreed=True,
+    )
+    url = reverse(
+        'comments_async-list',
+        kwargs={'content_type': question_ct.pk,
+                'object_pk': question.pk})
+    data = {'comment': 'comment comment'}
+    with active_phase(question.module, AskPhase):
+        apiclient.post(url, data, format='json')
```
Rineee, line 274: Should this raise permission denied as user is not authenticated?

Rineee (approved):
Supercool! Better structured and so many tests!

I just have one tiny suggestion..

adhocracy4/comments_async/api.py
fuzzylogic2000: Weeeeeell, I changed it because otherwise the tests do not work. Which is a weird reason, but I just could not find anything faking a property. And I would still like to test in a4, so we cannot really use the property that we only have in a+. So, don't know. :cry:

tests/comments/test_async_api_org_terms.py
fuzzylogic2000: It tests the new line in the API that throws the otImplementedError (lines 226 and 227 of adhocracy4/comments_async/api.py)
I think, I should redo the test naming. It's really bad and does not make a lot of sense.

tests/comments/test_async_api_org_terms.py
fuzzylogic2000: Yeah, it is in there from other tests to add a comment and then test the info. But I am not actually using it, so it can just be removed. Same in the next test. Will do!

adhocracy4/comments_async/api.py
Rineee: Ah ok, I see! But doing all these queries to make the test work also doesnt seem right, no? Maybe we can just move the failing tests to Aplus? Or is it all of them?

fuzzylogic2000: Yeah, for the poll it would be all of them. And I think, if we use `user.has_agreed_on_org_terms` in the API, we should also check that it is implemented and otherwise throw an error. Just relying on that it's there in the project also does not seem like the best way to go.

PR #1131, opened 2022-06-27
polls/assets/PollQuestions: do not show terms of use checkbox to unau…

Rineee:
…thenticated users

As the agreedTermsOfUse is null when the user is not authenticated, the checkbox was also shown to unauthenticated users.


goapunk (approved):
nice catch!

PR #1136, opened 2022-06-28
Ks 2022 06 blueprint types

Rineee:
I thought you could already have a look @fuzzylogic2000 :) Can be tested with https://github.com/liqd/adhocracy-plus/pull/1954

~~The migration of the old modules are still missing.. and actually I am not sure how to do it, because it needs to happen in A4, but for all projects? Well, problem for tomorrow..~~


adhocracy4/modules/admin.py
```diff
@@ -26,6 +27,11 @@ class ModuleAdmin(admin.ModelAdmin):
         phase_admin.PhaseInline
     ]
     list_filter = ('project__organisation', 'project')
+    if hasattr(settings, 'A4_BLUEPRINT_TYPES'):
+        list_filter += ('blueprint_type',)
+    else:
+        exclude = ['blueprint_type']
```
Rineee, line 33: I just realized, that maybe we dont want to show the type at all in admin? Or make it not editable? So modules cant be assigned a wrong type.. What do you think @fuzzylogic2000?

fuzzylogic2000: Hmm, if we filter by it, we should also show it.
Yes, and I think, making it not editable is a good idea. If we absolutely need to change it, we can still do that in the shell.

fuzzylogic2000 (commented):
NIce, good idea how you did the migration! But can't we put it into the project? Maybe with ideas, as these are kind of the basic for most our modukes?

fuzzylogic2000 (approved):
It looks good! 
And works in mB (tried it with an extra branch without the React update https://github.com/liqd/adhocracy4/tree/mB-v2203.1-test-branch).
I will still also test in a+!

adhocracy4/modules/admin.py
```diff
@@ -26,6 +27,12 @@ class ModuleAdmin(admin.ModelAdmin):
         phase_admin.PhaseInline
     ]
     list_filter = ('project__organisation', 'project')
+    if hasattr(settings, 'A4_BLUEPRINT_TYPES'):
+        list_filter += ('blueprint_type',)
+        readonly_fields = ('blueprint_type',)
+    else:
+        exclude = ['blueprint_type']
+
     list_display = ('__str__', 'name')
```
fuzzylogic2000, line 36: Can you also add the blueprint type to the list display when the settings are there?

fuzzylogic2000 (approved):
Also works in a+! :tada: 
Just one teenytiny thing and then we can merge.

Rineee:
> It looks good! And works in mB (tried it with an extra branch without the React update https://github.com/liqd/adhocracy4/tree/mB-v2203.1-test-branch). I will still also test in a+!

Cool, thank you for testing it in mB! :pray:

fuzzylogic2000 (approved):
Supercool!

PR #1140, opened 2022-07-05
comments: move date conversion to own function and use in serializer

Rineee:
I moved the date conversion to a function, so we can use it in other places. I guess it would better fit into contrib, but we dont have that in a4, so left it in comments app for now.. could also create contrib folder?

### DIFF
--- a/adhocracy4/api/dates.py
+++ b/adhocracy4/api/dates.py
@@ -1,0 +1,19 @@
+from django.template import defaultfilters
+from django.utils import timezone
+
+
+def get_datetime_display(datetime):
+    """Convert datetime instance to localtime and return as str."""
+    local_datetime = timezone.localtime(datetime)
+    local_date = defaultfilters.date(local_datetime)
+    local_time = defaultfilters.time(local_datetime)
+
+    return local_date + ', ' + local_time
+
+
+def get_date_display(datetime):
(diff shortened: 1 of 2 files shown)


fuzzylogic2000 (commented):
Looks good!
Could it go to api as a helper as that is where we use it?
And when do we add tests? :)

Rineee:
> Looks good! Could it go to api as a helper as that is where we use it? And when do we add tests? :)

Yeah, I also thought about that, but we are actually using it in the serializers and in the api folder, there is only router stuff.. so I thought it does also not quite fit there? But maybe better than comments.. So: you prefer putting it to api instead of making a contrib folder?

I will add a test with the other tests, its on the list!

Rineee:
> > Looks good! Could it go to api as a helper as that is where we use it? And when do we add tests? :)
> 
> Yeah, I also thought about that, but we are actually using it in the serializers and in the api folder, there is only router stuff.. so I thought it does also not quite fit there? But maybe better than comments.. So: you prefer putting it to api instead of making a contrib folder?
> 
> I will add a test with the other tests, its on the list!

Alright, moved it to api now as definitely better than comments and then we can merge!

fuzzylogic2000 (approved):
Very nice!

PR #1145, opened 2022-07-11
Add aria-label to rating buttons

goapunk:


fuzzylogic2000 (commented):
Nice! 
Only thing: we try to add the strings to translate as variables in the beginning of the jsx-files as far as possible, because if we don't, the strings sometimes disappear from the translation files.

goapunk:
> Nice! Only thing: we try to add the strings to translate as variables in the beginning of the jsx-files as far as possible, because if we don't, the strings sometimes disappear from the translation files.

@fuzzylogic2000 will change it! I just looked at a random other file in a4 how they are handled, I guess I picked the wrong one

PR #1153, opened 2022-07-14
comments/models: update _former_comment after save

Rineee:
Had to change this again as need to update former_comment after saving :see_no_evil: 

The super.save() does not return anything, so fine to remove the return here, I think.

Used here: https://github.com/liqd/a4-kosmo/pull/268

### DIFF
--- a/adhocracy4/comments/models.py
+++ b/adhocracy4/comments/models.py
@@ -58,6 +58,7 @@ class Comment(base.UserGeneratedContentModel):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        # save former comment to detect if comment text has changed
         self._former_comment = transforms.clean_html_all(
             self.comment)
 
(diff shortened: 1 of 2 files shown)


fuzzylogic2000 (commented):
I don't get that. Why after save? Why isn't `self.comment` in the new state after save?

Rineee:
> I don't get that. Why after save? Why isn't `self.comment` in the new state after save?

@fuzzylogic2000 So in kosmo, I want to compare comment and former_comment. The code in kosmo listens to post_save signal. But if I set former_comment to comment before save, they are always both the same.. so I need to do that afterwards. Does that make sense?

fuzzylogic2000:
> > I don't get that. Why after save? Why isn't `self.comment` in the new state after save?
> 
> @fuzzylogic2000 So in kosmo, I want to compare comment and former_comment. The code in kosmo listens to post_save signal. But if I set former_comment to comment before save, they are always both the same.. so I need to do that afterwards. Does that make sense?

Today it doesn't. :blush: But I will give it another go tomorrow and ask you again if it still doesn't make sense. :smile:

Rineee:
Maybe just check the code without the change.. while saving, former_comment and comment are always the same.

Rineee:
> > > I don't get that. Why after save? Why isn't `self.comment` in the new state after save?
> > 
> > 
> > @fuzzylogic2000 So in kosmo, I want to compare comment and former_comment. The code in kosmo listens to post_save signal. But if I set former_comment to comment before save, they are always both the same.. so I need to do that afterwards. Does that make sense?
> 
> Today it doesn't. blush But I will give it another go tomorrow and ask you again if it still doesn't make sense. smile

Maybe just check the code without the change.. while saving, former_comment and comment are always the same then.

Rineee:
@fuzzylogic2000 I thought about this again and realized that setting the _former_comment to comment after save was not necessary as that is then done in init, next time the comment is instantiated. So I removed that now, maybe now it also makes more sense for you? Good that you asked! I also added that to the tests..

fuzzylogic2000 (approved):
Yes! Nice!

PR #1155, opened 2022-07-18
apps/phases: validate that phase dates in modules are ordered by weig…

### LINKED ISSUE #919: No Validation or Error when Phases in Dashboard are set in Illogical Order
URL: https://meinberlin-dev.liqd.net/projekte/module/burgerinnenhaushalt-2-phasen/?mode=list
user: Project initiator
expected behaviour: If I have more than one phase in a module, I would expect them to only be able to be set to occur in 1st proposal, 2nd rating, and 3rd voting time slots. If I make a mistake, I hope I am alerted and cannot publish this.
behaviour: Both 3-phase and 2-phase modules can be published with phases that make no sense in their time line. Users can set up and publish modules with voting phases that occur before rating and proposal phases. There is no validation or error.
important screensize: any
device & browser: any
Comment/Question: moved from mB: https://github.com/liqd/a4-meinberlin/issues/4029

Screenshot?
2-Phase module published with reverse phases:
Bildschirmfoto 2021-12-08 um 12 41 47

fuzzylogic2000:
…ht - fixes #919

### DIFF
--- a/adhocracy4/phases/forms.py
+++ b/adhocracy4/phases/forms.py
@@ -4,9 +4,7 @@ from django.utils.translation import gettext_lazy as _
 
 class PhaseInlineFormSet(BaseInlineFormSet):
     def clean(self):
-        """
-        Make sure phases of the same module don't overlap.
-        """
+        """Make sure phases of the same module don't overlap."""
         super().clean()
         phase_dates = []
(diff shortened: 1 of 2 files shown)


Rineee (approved):
Supercool!

PR #1161, opened 2022-07-25
apps/phases: remove deprecated property is_first_of_project

fuzzylogic2000:


goapunk (commented):
Looks good to me but I'm unsure how to test it?

goapunk (approved):

fuzzylogic2000:
> Looks good to me but I'm unsure how to test it?

If you use it in the old state with the old property, the added test will fail. I thought that is enough confirmation...
@goapunk

fuzzylogic2000:
Right! But I also have to add tests for the new property! So, don't merge yet!

Rineee (approved):
Cool!

adhocracy4/phases/models.py
```diff
@@ -119,15 +114,9 @@ def is_over(self):
...
-        phase order is determined by weight.
-        """
-        warnings.warn(
-            "is_first_of_project is deprecated as it relies "
-            "on the project only having one module",
-            DeprecationWarning
+    def starts_first_of_project(self):
+        """Test if this is the phase that starts first in the project."""
+        phases = self.module.project.phases.order_by(
+            F('start_date').asc(nulls_last=True)
```
fuzzylogic2000, line 121: @Rineee @goapunk It was good, that I added more tests, because doing that I noticed that phases without dates were actually ordered first. I changed that. 
Now there is only the edge-case left when there are only phases without dates. Then a phase without date will be the "first one starting", which might be surprising, but is also not wrong. Especially if there is only one phase it's right. But I think, we can leave it like that.
And then I found another surprising thing, I fixed and got carried away with the tests a bit...

fuzzylogic2000: @Rineee And now it is really finished (I hope!!!!!!11!!!!! :p) and will filter out all phases from unpublished modules. :)

adhocracy4/phases/models.py
```diff
@@ -34,12 +34,13 @@ def future_phases(self):
...
 
-        Note: Phases without a start date are assumed to start in the future.
+        Note: Phases without a start date are assumed to start in the future,
+        but are appended to the end of the list.
         """
         return self\
             .filter(models.Q(start_date__gt=timezone.now())
                     | models.Q(start_date=None))\
-            .order_by(F('start_date').asc(nulls_first=True))
+            .order_by(F('start_date').asc(nulls_last=True))
```
Rineee, line 43: makes sense!

adhocracy4/phases/models.py
```diff
@@ -119,15 +114,17 @@ def is_over(self):
...
-        )
-        return self == self.module.project.phases.first()
+        if not self.module.is_draft:
+            phases = self.module.project.phases.filter(
+                module__is_draft=False
+            ).order_by(
+                F('start_date').asc(nulls_last=True)
+            )
+            return self == phases.first()
+        return False
```
Rineee, line 131: Cool!

Rineee (approved):
Yeah cool!

And thank you for testing the voting predicates!

PR #1176, opened 2022-08-04
comments_async/serializers//static: serve fallback image in own field

Rineee:
So we can distinguish if a user image is actually present or not.. 

Unfortunately cant test the fallback_image here, as there is no user model in a4 (or is there a way @fuzzylogic2000 ?), so coverage decreases :neutral_face: But I will add a test in aplus!

### DIFF
--- a/adhocracy4/comments_async/static/comments_async/comment_list.jsx
+++ b/adhocracy4/comments_async/static/comments_async/comment_list.jsx
@@ -14,7 +14,7 @@ const CommentList = (props) => {
               user_name={comment.user_name}
               user_pk={comment.user_pk}
               user_profile_url={comment.user_profile_url}
-              user_image={comment.user_image}
+              user_image={comment.user_image ? comment.user_image : comment.user_image_fallback}
               child_comments={comment.child_comments}
               created={comment.created}
               modified={comment.modified}
--- a/tests/comments/test_async_api.py
+++ b/tests/comments/test_async_api.py
@@ -361,7 +361,7 @@ def test_fields(user, apiclient, question_ct, question):
     assert not response.data['use_org_terms_of_use']
 
     commentDict = response.data['results'][0]
-    assert len(commentDict) == 21
+    assert len(commentDict) == 22
     assert 'child_comments' in commentDict
     assert 'comment' in commentDict
     assert 'comment_categories' in commentDict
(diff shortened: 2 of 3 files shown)


tests/comments/test_async_api.py
```diff
@@ -361,7 +361,7 @@ def test_fields(user, apiclient, question_ct, question):
     assert not response.data['use_org_terms_of_use']
 
     commentDict = response.data['results'][0]
-    assert len(commentDict) == 21
+    assert len(commentDict) == 22
     assert 'child_comments' in commentDict
     assert 'comment' in commentDict
     assert 'comment_categories' in commentDict
```
fuzzylogic2000, line 367: I don't get why we only test a subset of the fields- Can we add all and also the new one?

adhocracy4/comments_async/static/comments_async/comment_list.jsx
```diff
@@ -14,7 +14,7 @@ const CommentList = (props) => {
               user_name={comment.user_name}
               user_pk={comment.user_pk}
               user_profile_url={comment.user_profile_url}
-              user_image={comment.user_image}
+              user_image={comment.user_image ? comment.user_image : comment.user_image_fallback}
```
fuzzylogic2000, line 17: Haha, I was a bit confused initially, because this affects all projects. But right, we also have the react comments in a4. :smile:

fuzzylogic2000 (commented):
Nice!
For the testing: no, there is no way. I tried quite a while to get an image only into the factory without adding an extra user in the test-apps, but didn't succeed. So, I think, we can only test that the field is there and `None`.

fuzzylogic2000 (approved):
Well, let's actually merge without any more tests, but like this! It doesn't need to be done in this PR. Maybe @goapunk can have a look and merge if he also thinks it's fine?

tests/comments/test_async_api.py
Rineee: Yeah, to be honest I find that test a bit annoying, I had to update that number a hundred times already.. :sweat_smile: But I guess if we do test, it should be all fields, yes.

fuzzylogic2000: Yeah, I started that, and then I thought, it doesn't make sense to test these fields here, but in the serializer test and then I started that and noticed, that we only do exclude the creator field instead of only adding the fields we actually want and then I just gave up and thought, that that's for another time. :woman_shrugging:

PR #1184, opened 2022-08-15
requirements: include extras for Bleach and easy-thumbnail

goapunk:
bleach needs the tinecss2 library which was moved to the css extra - this was previously installed as dependency of easy-thumbnail but there it also moved into the extra svg. Install both explicitely


Rineee (approved):
lgtm!

PR #1199, opened 2022-09-01
apps/actions: make sure only actions for published modules are created

fuzzylogic2000:
will fix (after cherry-picking to hotfix branch) https://github.com/liqd/a4-meinberlin/issues/4383


phillimorland (approved):
Nice, not sure on testing but reads like it should work! :)

PR #1207, opened 2022-09-06
dashboard/views: add module type in ProjectCreateView

Rineee:
Needed for external projects, bplans and project containers.

Kind of belongs to https://github.com/liqd/a4-meinberlin/pull/4449

### DIFF
--- a/adhocracy4/dashboard/views.py
+++ b/adhocracy4/dashboard/views.py
@@ -1,4 +1,5 @@
 from django.apps import apps
+from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth import get_user_model
 from django.contrib.messages.views import SuccessMessageMixin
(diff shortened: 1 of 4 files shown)


fuzzylogic2000 (commented):
Looks good!
I am a bit unsure about adding the types to the settings, but I also see no other way to do it. (And let's discuss that in the other PR.) 
I also wasn't aware that we overwrite the ProjectCreateView for the mB-Projects, but then still use this one for the external projects and so on. Makes sense, but is also a bit surprising. Maybe we should add a comment here or somewhere, were it's actually used and also what it's for (Adding one-module-projects).
Should we merge right away or do you want to add the test first?

Rineee:
> Looks good! I am a bit unsure about adding the types to the settings, but I also see no other way to do it. (And let's discuss that in the other PR.) I also wasn't aware that we overwrite the ProjectCreateView for the mB-Projects, but then still use this one for the external projects and so on. Makes sense, but is also a bit surprising. Maybe we should add a comment here or somewhere, were it's actually used and also what it's for (Adding one-module-projects). Should we merge right away or do you want to add the test first?

@fuzzylogic2000 I thought, the other possibility would be to just add empty blueprint type for bplans etc (so just type = '' in the defintions of the blueprints for bplan, ext proj, proj container).. then we would not need this and also not need to add those in the settings? But then they would show up empty in django admin.. dont really know, what would you prefer?

Rineee:
@fuzzylogic2000 Ok, I finally added a test..

fuzzylogic2000 (approved):
Nice!

PR #1216, opened 2022-09-19
apps/dashboard: add for_superuser property to module component menu

fuzzylogic2000:
needed for https://github.com/liqd/a4-meinberlin/pull/4484
I thought about making more use of this also here and using it in https://github.com/liqd/adhocracy4/blob/04466cb3b14ca7a952c09e060a6c615c1c9173ae/adhocracy4/dashboard/templates/a4dashboard/base_dashboard_project.html#L27 
But adding all that complexity in a4, when we don't need it, did not make sense to me.

### DIFF
--- a/adhocracy4/dashboard/components/__init__.py
+++ b/adhocracy4/dashboard/components/__init__.py
@@ -40,6 +40,7 @@ class DashboardComponent:
     identifier = ''
     weight = 0
     label = ''
+    for_superuser_only = False
 
     def is_effective(self, project_or_module):
         """Return if the component is effective for the current context.
--- a/adhocracy4/dashboard/__init__.py
+++ b/adhocracy4/dashboard/__init__.py
@@ -126,5 +126,6 @@ class ProjectDashboard:
                     'is_active': is_active,
                     'url': url,
                     'is_complete': is_complete,
+                    'for_superuser_only': component.for_superuser_only,
                 })
         return module_menu or None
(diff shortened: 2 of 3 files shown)


adhocracy4/dashboard/components/__init__.py
```diff
@@ -40,6 +40,7 @@ class DashboardComponent:
     identifier = ''
     weight = 0
     label = ''
+    for_superuser = False
```
khamui, line 43: think the wording is fine and easy to understand, imho it would be a tiny little bit more accurate:
`for_superuser_only`, because `for_superuser = False` sounds like all other users are allowed, except of superusers.

khamui (commented):
cool!

adhocracy4/dashboard/components/__init__.py
fuzzylogic2000: Yeah, I was also thinking about doing it the other way round and saying `for_initiators = True` or `initiators_allowed = True`. Yes, but I think your version is even clearer, so will do that after there is a second review. :)

Rineee (approved):
Looks good to me! And then you found a way to check if the user is superuser?

fuzzylogic2000:
@khamui Could you have another look and then merge? Then I will  update the mB PR. Or @Rineee if you don't have time?

khamui:
cool!

PR #1221, opened 2022-09-22
rewrite image_uploader.js in vanilla js

goapunk:


adhocracy4/images/assets/image_uploader.js
```diff
@@ -1,27 +1,29 @@
...
+          text.value = name
+        }
         previewImage.alt = name
         if (window.FileReader) {
           const reader = new window.FileReader()
-          reader.onload = function (e) {
-            previewImage.attr('src', e.target.result)
+          reader.addEventListener('load', function (e) {
+            previewImage.setAttribute('src', e.target.result)
             clearInput.prop('checked', false)
```
khamui, line 22: in line 26 you rewrote a similar assignment, maybe you want to do it here as well?

khamui (commented):
nice! just one comment

adhocracy4/images/assets/image_uploader.js
goapunk: oh, good catch `prop` is still jquery :facepalm:

khamui (approved):
cool! looking good from reading code.

I can't test it because currently i am having a4 linking issues

fuzzylogic2000 (approved):
Yeah, I tried it out, it works!
But I needed a while to find out how we use it (through the images.widgets.ImageInputWidget, which is used in images.forms.ImageField, which is then used in images.fields.ConfiguredImageField, which we use for all images in the model files of mB and probably also elsewhere). I don't know if that should go into a comment or docstring somewhere?

goapunk:
> Yeah, I tried it out, it works! But I needed a while to find out how we use it (through the images.widgets.ImageInputWidget, which is used in images.forms.ImageField, which is then used in images.fields.ConfiguredImageField, which we use for all images in the model files of mB and probably also elsewhere). I don't know if that should go into a comment or docstring somewhere?

sounds like a good idea, do you want to add it? You actually investigated the usage much better than I did :sweat_smile:

fuzzylogic2000:
> > Yeah, I tried it out, it works! But I needed a while to find out how we use it (through the images.widgets.ImageInputWidget, which is used in images.forms.ImageField, which is then used in images.fields.ConfiguredImageField, which we use for all images in the model files of mB and probably also elsewhere). I don't know if that should go into a comment or docstring somewhere?
> 
> sounds like a good idea, do you want to add it? You actually investigated the usage much better than I did sweat_smile

Ah, another time. :p

PR #1241, opened 2022-10-17
apps/labels: add label filter

fuzzylogic2000:


fuzzylogic2000:
> Lgtm! Do you want to add tests first or do we break the rule again? stuck_out_tongue_winking_eye

Ah, yes, I don't know. But yes, let's break the rule! I am still doing more refactoring in mB and would like to finish that before writing the tests... @Rineee

PR #1247, opened 2022-10-20
Jd 2022 10 react fixes

goapunk:
* remove ref from component initialization as it doesn't work for functional components
* rename index.jsx of comments to react_comments.jsx to be more in line with the other components
* move polls from assets/ to static/ to be in line with other components


index.js
```diff
@@ -6,11 +6,11 @@ import './adhocracy4/static/global_jquery'
 export { default as alert } from './adhocracy4/static/Alert'
 export { default as api } from './adhocracy4/static/api'
 export { default as comments } from './adhocracy4/comments/static/comments/react_comments'
-export * as commentsAsync from './adhocracy4/comments_async/static/comments_async'
+export * as commentsAsync from './adhocracy4/comments_async/static/comments_async/react_comments'
```
phillimorland, line 9: so we don't double up names I would call this react_comments_async or something?

phillimorland (commented):
looks good but one nameing thing and may cause conflicts with poll Pr which is ready again

index.js
goapunk: done

index.js
```diff
@@ -10,7 +10,7 @@ export * as commentsAsync from './adhocracy4/comments_async/static/comments_asyn
 export { default as config } from './adhocracy4/static/config'
 export { default as follows } from './adhocracy4/follows/static/follows/react_follows'
 export * as maps from './adhocracy4/maps/static/a4maps/a4maps_common'
-export { default as polls } from './adhocracy4/polls/assets/react_polls'
```
phillimorland, line 13: @goapunk sorry missed this before, need to also rename the folder :p then feel free to merge

goapunk: sorry, don't understand?

goapunk: ahh, forgot to commit :man_facepalming:

phillimorland (approved):
Nice

PR #1251, opened 2022-10-24
Ks 2022 10 support map popup

Rineee:


adhocracy4/modules/predicates.py
```diff
@@ -172,3 +172,11 @@ def is_allowed_change_item(user, item):
...
                  phase_predicates.phase_allows_change(user, item)))
     return False
+
+
+def module_is_between_phases(past_phase_type, future_phase_type, module):
+    return (past_phase_type
+            in [phase.type for phase in module.phases.past_phases()]
+            and future_phase_type
+            in [phase.type for phase in module.phases.future_phases()]
+            and module.phases.active_phases().count() == 0)
```
fuzzylogic2000, line 182: Nice! But maybe start with the active phase check, so that it can already stop when that's the case?

Rineee: @fuzzylogic2000 Do you mean because of ended projects?

Rineee:
@fuzzylogic2000 I changed it quite a bit now (also did not like the nesting before) and added a test, do you want to have another look?

adhocracy4/modules/predicates.py
fuzzylogic2000: YEs, I thought most modules will not have an active phase and thus that would be the condition mostly failing

fuzzylogic2000 (approved):
Supercool! I love it!

PR #1255, opened 2022-10-25
updating our jsx errors to be a11y compliant

phillimorland:
part of fix for https://github.com/liqd/a4-meinberlin/issues/4232


fuzzylogic2000 (approved):
Supercool!

PR #1261, opened 2022-11-07
replace template strings with regular string concat

goapunk:
This will fix the issue with the disappearing translations. Template strings like `some string ${someVar}` will break our translations so we should avoid them

### DIFF
--- a/adhocracy4/comments_async/static/comments_async/comment.jsx
+++ b/adhocracy4/comments_async/static/comments_async/comment.jsx
@@ -262,7 +262,7 @@ export default class Comment extends React.Component {
     if (this.props.has_deleting_permission) {
       return (
         <Modal
-          name={`comment_delete_${this.props.id}`}
+          name={'comment_delete_' + this.props.id}
           partials={{ title: django.gettext('Do you really want to delete this comment?') }}
           handleSubmit={() => this.props.onCommentDelete(this.props.index, this.props.parentIndex)}
           action={django.gettext('Delete')}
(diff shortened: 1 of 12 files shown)


adhocracy4/comments_async/static/comments_async/comment.jsx
```diff
@@ -26,19 +26,29 @@ const translated = {
...
+  deleteComment: django.gettext('Do you really want to delete this comment?'),
+  delete: django.gettext('Delete'),
+  abort: django.gettext('Abort'),
+  deletedyByCreatorOn: django.gettext('Deleted by creator on'),
+  deletedByModeratorOn: django.gettext('Deleted by moderator on'),
+  blockedByModeratorOn: django.gettext('Blocked by moderator on'),
+  lastEditOn: django.gettext('Latest edit on'),
+  moderator: django.gettext('Moderator'),
+  hideReplies: django.gettext('hide replies'),
+  reply: django.pgettext('verb', 'Reply')
```
goapunk, line 39: I pulled out all the translations into this because I thought we decided that translations should always be on top but then realized there are a lot of files which don't have the `translated` at all, so I didn't continue as this would have taken quite long

fuzzylogic2000: Yes, the translations need to be on top not to get lost in the makemessages. 
I think, we wanted to change that when we touch the files, but we might have forgotten again.

fuzzylogic2000 (approved):
Nice!

adhocracy4/comments_async/static/comments_async/comment.jsx
goapunk: how do we handle plurals? as they depend on a variable they can't be in the `translated` block? If we get rid of all template strings I guess that's not a problem anymore?

PR #1269, opened 2022-11-16
[#6785] categories/assets/category_formset: refactor category_formset

phillimorland:
Refactor to mostly remove jquery. It is still conditionally called for the select category icon element. After investigating alternatives, most options don't offer any advantage, the more accessible option that was found [here](https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html) would require a custom django widget and is still not actually recommended due to possible gaps in support, we can look into it as a technical backlog story.

### DIFF
--- a/adhocracy4/categories/assets/category_formset.js
+++ b/adhocracy4/categories/assets/category_formset.js
@@ -1,10 +1,5 @@
-(function (init) {
-  document.addEventListener('DOMContentLoaded', init, false)
-  document.addEventListener('a4.embed.ready', init, false)
-})(function () {
-  // Dynamically add or remove subforms to a formset.
-  const $ = window.jQuery
-  const $formsets = $('.js-formset')
+window.addEventListener('load', function () {
+  const formsets = document.querySelectorAll('.js-formset')
   const PLACEHOLDER = /__prefix__/g
   const dynamicFormSets = []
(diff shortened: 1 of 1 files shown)


adhocracy4/categories/assets/category_formset.js
```diff
@@ -1,82 +1,88 @@
...
+    let nextSibling = form.nextElementSibling
+    // console.log(form.children)
+    while (nextSibling && nextSibling !== this.formTemplate) {
+      const index = Array.from(this.formset.querySelectorAll('.js-form')).indexOf(nextSibling)
+      const indexTest = index - 1
+      console.log(index)
+      Array.from(nextSibling.firstElementChild.children).forEach(function (element, i) {
+        updateAttr(element, 'name', i)
+        updateAttr(element, 'for', i)
+        updateAttr(element, 'id', i)
```
phillimorland, line 70: `i` is actually the index of the field child elements, not the index of the parent element, so only the first element has it's attributes replaced as the following index's to not match `_old` variable from `updateAttr`

phillimorland:
@goapunk should I tag in here?

goapunk:
@phillimorland yes, please

goapunk (approved):
looks good to me!

PR #1272, opened 2022-11-22
categories//labels/models: add default ordering

Rineee:
This should fix https://github.com/liqd/a4-meinberlin/issues/4654, I will also add it for the moderation tasks in mB then.

Apparently there is no default ordering, I checked the broken example that was linked in the issue and indeed the categories were in random order there, although in most cases it works? Since we do not have a created field here, I guess pk is the best bet!


fuzzylogic2000 (approved):
Cool!

PR #1273, opened 2022-11-29
[updates] sorting poll classnames so uses same styling in documents and updating old btn class for following

phillimorland:


fuzzylogic2000 (approved):
Nice!
Are there already PRs in a+ and mB? Or should I merge and we do them then?

PR #1274, opened 2022-11-29
apps/modules: add some properties to determine days left of running m…

fuzzylogic2000:
…odule and if it is in the future


phillimorland (approved):
awesome!

PR #1285, opened 2022-12-05
mapbox to maplibre styles

goapunk:


adhocracy4/maps/static/a4maps/a4maps_common.js
```diff
@@ -22,11 +23,20 @@ export function createMap (L, e, {
     tap: false
   })
 
+  const transformRequest = (url, resourceType) => {
```
phillimorland, line 26: can we add a note here explaining why we need it to ensure we don't forget?

package.json
```diff
@@ -17,6 +17,7 @@
     "classnames": "2.3.2",
     "file-saver": "2.0.5",
     "js-cookie": "3.0.1",
+    "maplibregl-mapbox-request-transformer": "^0.0.1",
```
phillimorland, line 20: caret needs to go

phillimorland (approved):
coolio, well done! I will merge after little change and update my PR with new hash :)

phillimorland (commented):
sweet!

PR #1288, opened 2022-12-06
tests/modules and projects: add property tests and reorder to mirror …

fuzzylogic2000:
…property order from models

Some of the properties are still used in OPIN, added an issue there. And there is one deprecated one we still use in mB, so I did not yet remove it here, but added an issue in mB that we should remove it.


tests/modules/test_module_models.py
```diff
@@ -91,15 +91,7 @@ def test_last_active_phase(module, phase_factory):
...
-def test_first_phase_start_date(phase, phase_factory):
-    module = phase.module
-    first_phase = phase_factory(
-        module=module, start_date=phase.start_date - timedelta(days=1))
-    assert module.first_phase_start_date == first_phase.start_date
-
-
-@pytest.mark.django_db
-def test_module_dates(phase_factory):
+def test_module_start_and_module_and(phase_factory):
```
Rineee, line 94: test_module_start_and_module_end

Rineee (approved):
Really nice clean-up and testing.
I found one little typo, but good to merge!

PR #1290, opened 2022-12-08
assets/dashboard.js: rewrite updatedashboard to remove jQuery + update EditPollManagent.jsx

khamui:


phillimorland (approved):
very nice :) I think we should look at a system for js selectors when classes are used but that can be done later

khamui:
> very nice :) I think we should look at a system for js selectors when classes are used but that can be done later

true. I thought about that, but wasn't sure of a better solution for now.

PR #1292, opened 2022-12-12
deps: adding js-cookie and maplibregl-mapbox-request-transformer as p…

phillimorland:
…eer dependencies


phillimorland (commented):
coolio!

PR #1297, opened 2022-12-19
apps/modules: make sure abbreviated string gets into translations

fuzzylogic2000:
will fix (after update) https://github.com/liqd/adhocracy-plus/issues/2151


Rineee (approved):
Nice!

PR #1299, opened 2022-12-19
comments_async/comment_form.jsx: changing grid for sm to fix weird li…

khamui:
…ne breaking in aplus

fixes liqd/adhocracy-plus#1980


fuzzylogic2000 (commented):
Looks good to me. But does it work and look good in all projects? (Or at least mB and a+)

khamui:
@fuzzylogic2000 i tested again. and indeed found the mobile variant (with terms of use) breaking. That should work for all variants now (in mB and aplus).

fuzzylogic2000 (approved):
Cool!

PR #1300, opened 2022-12-21
apps/modules: add some more explanation to property module_running_ti…

fuzzylogic2000:
…me_left_abbreviated


phillimorland (approved):
fab! thank you! :)

PR #1303, opened 2022-12-27
setup/deps: include all packages only used in a4

fuzzylogic2000:


goapunk (approved):
cool. looks good to me

PR #1308, opened 2022-12-30
filters: when invalid filter value, use default value if there is one

Rineee:
When the filter argument is invalid, django adds and shows the label 'All', also in the case that this isnt a 'real' choice. 
With this, it would now use the default value of the filter if there is one.


fuzzylogic2000 (approved):
Very nice! 
Do you want to add the tests here or later? :upside_down_face:

Rineee:
So this is not working, the tests are failing as it still takes the ordering by pks.. maybe this could help? https://django-filter.readthedocs.io/en/stable/ref/filterset.html#custom-forms-using-form

PR #1313, opened 2023-01-03
WIP Pm 2023 icon select

phillimorland:
possible options:
[select only combobox](https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html) - still not recommended
[a11y compliant option but a bit old](https://24ways.org/2019/making-a-better-custom-select-element/)


phillimorland:
testing attempt https://github.com/liqd/adhocracy-plus/pull/2200

PR #1321, opened 2023-01-11
[6669] apps/exports: change wording for the moderator feedback fields

fuzzylogic2000:
Well, we kind of switched the names now, so the statement is called feedback and what we call feedback in the code is the status now. :cry: I guess, we also need to at least adapt the verbose names on a+ or it will be veeeery confusing. But well, for the current mB story, this is it.


Rineee:
Uh yes, confusing.. but we did confirm this wording change with Aplus, right? @CarolingerSeilchenspringer

fuzzylogic2000:
@khamui is currently doing the renaming. He can also do it in here when the models are through in mB and a+.

tests/apps/ideas/models.py
```diff
@@ -21,9 +21,9 @@ class Idea(Item):
     name = models.CharField(max_length=120,
                             default='Can i haz cheezburger, pls?')
     description = RichTextField(verbose_name='Description', blank=True)
-    moderator_feedback = models.CharField(max_length=256, blank=True)
-    moderator_statement = models.OneToOneField(
-        ModeratorStatement,
+    moderator_status = models.CharField(max_length=256, blank=True)
+    moderator_feedback_text = models.OneToOneField(
```
Rineee, line 25: To me `moderator_feedback` makes more sense here? And then the text is `moderator_feedback.feedback_text` .. why did you name it `moderator_feedback_text`?

Rineee (approved):
Yes, nice!

PR #1333, opened 2023-01-19
[issues] polls/PollResults: rm the highlight class as not usable with current …

phillimorland:
…slick structure


phillimorland:
@khamui i fucked up with adding the highlight class, realised I can't actually use it also it returns false when not true?! so better to remove, sorry!

PR #1335, opened 2023-01-23
poll/PollQuestions//PollResults: change alert behavior so it makes se…

phillimorland:
…nse, add clearer variable names, only show slider when there is an answer

### DIFF
--- a/adhocracy4/polls/static/PollDetail/PollQuestion.jsx
+++ b/adhocracy4/polls/static/PollDetail/PollQuestion.jsx
@@ -48,11 +48,12 @@ export const PollQuestion = (props) => {
   const handleMultiChange = (event, isOther) => {
     const choiceId = parseInt(event.target.value)
     setUserChoices(prevState => [...prevState, choiceId])
+    setOtherChoiceAnswer(otherChoiceAnswer)
     props.onMultiChange(props.question.id, choiceId)
-    if (!isOther) {
-      props.onOtherChange(props.question.id, '', findOtherChoice())
-    } else {
+    if (isOther) {
       setOtherChoiceAnswer('')
+    } else {
+      props.onOtherChange(props.question.id, otherChoiceAnswer, findOtherChoice())
     }
   }
 
@@ -91,6 +92,7 @@ export const PollQuestion = (props) => {
                       disabled={!props.question.authenticated || props.question.isReadOnly}
                     />
                     <span className="radio__text">{choice.is_other_choice ? translated.other : choice.label}</span>
+
                     {choice.is_other_choice &&
                       <>
                         <input
--- a/adhocracy4/polls/static/PollDetail/PollQuestions.jsx
+++ b/adhocracy4/polls/static/PollDetail/PollQuestions.jsx
@@ -45,13 +45,21 @@ class PollQuestions extends React.Component {
     this.handleTermsOfUse = this.handleTermsOfUse.bind(this)
 
     this.linkToPoll = (
-      <button type="button" className="btn poll__btn--link" onClick={() => this.handleToggleResultsPage()}>
+      <button
+        type="button"
+        className="btn poll__btn--link"
+        onClick={() => { this.handleToggleResultsPage(); this.removeAlert() }}
+      >
         {django.gettext('To poll')}
       </button>
     )
 
     this.linkChangeVote = (
-      <button type="button" className="btn poll__btn--link" onClick={() => this.handleToggleResultsPage()}>
+      <button
+        type="button"
+        className="btn poll__btn--link"
+        onClick={() => { this.handleToggleResultsPage(); this.removeAlert() }}
+      >
         {django.gettext('Change answer')}
       </button>
     )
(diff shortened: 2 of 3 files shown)


adhocracy4/polls/static/PollDetail/PollQuestion.jsx
```diff
@@ -48,11 +48,12 @@ export const PollQuestion = (props) => {
...
     setUserChoices(prevState => [...prevState, choiceId])
+    setOtherChoiceAnswer(otherChoiceAnswer)
     props.onMultiChange(props.question.id, choiceId)
-    if (!isOther) {
-      props.onOtherChange(props.question.id, '', findOtherChoice())
-    } else {
+    if (isOther) {
       setOtherChoiceAnswer('')
+    } else {
+      props.onOtherChange(props.question.id, otherChoiceAnswer, findOtherChoice())
```
phillimorland, line 56: should this be inside another if? although the initial condition should stop anything being saved that shouldn't be, right?

adhocracy4/polls/static/PollDetail/PollQuestions.jsx
```diff
@@ -45,13 +45,21 @@ class PollQuestions extends React.Component {
     this.handleTermsOfUse = this.handleTermsOfUse.bind(this)
 
     this.linkToPoll = (
-      <button type="button" className="btn poll__btn--link" onClick={() => this.handleToggleResultsPage()}>
+      <button
+        type="button"
+        className="btn poll__btn--link"
+        onClick={() => { this.handleToggleResultsPage(); this.removeAlert() }}
+      >
```
phillimorland, line 52: success alert was still showing when going back to poll, was confusing

adhocracy4/polls/static/PollDetail/PollQuestions.jsx
```diff
@@ -239,25 +247,28 @@ class PollQuestions extends React.Component {
...
-    const modifiedQuestions = this.state.questions.filter(q => q.modified)
-    const validatedQuestions = modifiedQuestions.filter(q => {
-      if (!q.is_open) {
-        const otherChoice = q.choices.find(c => c.is_other_choice)
-        const otherChoiceSelected = otherChoice && q.userChoices.filter(uc => uc === otherChoice.id).length > 0
+
+    const modifiedAnswers = this.state.questions.filter(question => question.modified)
+
+    const validatedQuestions = modifiedAnswers.filter(question => {
+      if (!question.is_open) {
```
phillimorland, line 259: thought using words was clearer specially in case of userChoice but happy to go with majority on this?

adhocracy4/polls/static/PollDetail/PollQuestions.jsx
```diff
@@ -347,8 +359,14 @@ class PollQuestions extends React.Component {
                     )
               ))}
               <Alert onClick={() => this.removeAlert()} {...this.state.alert} />
-              {!this.isReadOnly()
+
+              {this.isReadOnly()
```
phillimorland, line 363: wasn't sure why the false before the true here, didn't seem to have a purpose, unless I missed it?

adhocracy4/polls/static/PollDetail/PollQuestions.jsx
```diff
@@ -314,31 +325,32 @@ class PollQuestions extends React.Component {
...
             <div className="pollquestionlist-container">
-              {this.state.result.map((q, idx) => (
+              {this.state.result.map((question, idx) => (
                 <PollResults
                   key={idx}
-                  question={q}
+                  question={question}
                 />
               ))}
+              <Alert onClick={() => this.removeAlert()} {...this.state.alert} />
```
phillimorland, line 334: so success alert is shown on results page as discussed with Marcelo, else it wouldn't be seen

phillimorland:
I have not tested this well but have to go now, sorry if its crap! can also double check it tomorrow.

adhocracy4/polls/static/PollDetail/PollQuestions.jsx
khamui: yes, it is better :)

adhocracy4/polls/static/PollDetail/PollQuestions.jsx
khamui: better this way!

khamui (changes requested):
Cool! just one change request and a question

General question: if one multi select question is not answered at all, it is not going to the results page. Is this intended?

adhocracy4/polls/static/PollDetail/PollQuestion.jsx
phillimorland: I can't remember :(

adhocracy4/polls/static/PollDetail/PollQuestion.jsx
```diff
@@ -48,11 +48,12 @@ export const PollQuestion = (props) => {
   const handleMultiChange = (event, isOther) => {
     const choiceId = parseInt(event.target.value)
     setUserChoices(prevState => [...prevState, choiceId])
+    setOtherChoiceAnswer(otherChoiceAnswer)
```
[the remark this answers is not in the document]
phillimorland: @khamui when editing your choice and updating multiselect checkbox when the other option is set the validation fails because the other text has not been updated and so is 'empty' according to validation but if we set it when the multiselect is updated then it is not empty according the validation and so works

phillimorland:
@khamui what change?

phillimorland:
> Cool! just one change request and a question
> 
> General question: if one multi select question is not answered at all, it is not going to the results page. Is this intended?

sorry missed that question, not intended will have a look

fuzzylogic2000 (approved):
For me it does work in a+! And it looks good. Is there sth. I should look out for before merge?

fuzzylogic2000:
> > Cool! just one change request and a question
> > General question: if one multi select question is not answered at all, it is not going to the results page. Is this intended?
> 
> sorry missed that question, not intended will have a look - could not reproduce

I tried it in a poll with one multi question and one open one and answering either or did bring me to the results page. Only removing my answers for both alerted me, that the poll was saved, but stayd on the poll page. That makes sense, as I didn't give an answer.

phillimorland:
@fuzzylogic2000 if you aren't having either of the issue we described in the comments then I would say we try a merge @khamui what do you think?

PR #1368, opened 2023-03-06
Kl 2023 03 orga terms

fuzzylogic2000:
fixes https://github.com/liqd/adhocracy-plus/issues/2276
fixes https://github.com/liqd/a4-kosmo/issues/346

- [x] I still have to try it in kosmo if it really fixes it, but it has to. :)


Rineee (changes requested):
Cool, it works with checking the terms of use now!
Buuut, the terms of use link seems wrong, also only in the child comments, it is the right one in the partent comments.

Rineee (approved):
Yeah cool!

PR #1384, opened 2023-04-04
filters/rest_filters: add DefaultsRestFilterSet

Rineee:
We are already using the DefaultsRestFilterSet in mb, so I figured adding in a4 is better than in aplus. The DistinctOrderingFilter could also be reused I guess.

Once this is merged, will also change in mb to use it from a4.


fuzzylogic2000 (approved):
Nice!

PR #1390, opened 2023-04-12
static/widget.js: re-write to rm jquery

philli-m:


index.js
```diff
@@ -10,7 +10,6 @@ export * as commentsAsync from './adhocracy4/comments_async/static/comments_asyn
 export { default as config } from './adhocracy4/static/config'
 export { default as follows } from './adhocracy4/follows/static/follows/react_follows'
 export * as maps from './adhocracy4/maps/static/a4maps/a4maps_common'
-export { default as polls } from './adhocracy4/polls/static/react_polls'
 export { default as ratings } from './adhocracy4/ratings/static/ratings/react_ratings'
```
philli-m, line 13: @goapunk I am wondering as we should no longer be using ES5 if this can be done away with but I am unsure what 'import mess' they are referring to in the notes, any idea?

goapunk: hmm,  good question. It seems to work without the import, but I only tested it very quickly

goapunk (approved):
lgtm!

PR #1394, opened 2023-04-18
WIP: rename fontawesome icons to v6

goapunk:
requires updating to  fontawesome v6 in all projects


philli-m (changes requested):
as we are touching all the files anyway I would update all classes to their current naming convention

PR #1404, opened 2023-05-09
maps: if no attribution is given try to load it from style

goapunk:


fuzzylogic2000 (commented):
It looks good, but I cannot test as the [mB branch](https://github.com/liqd/a4-meinberlin/pull/5052) is not installing with this a4 version. Any tips on how to make it work? (`--legacy-peer-deps` doesn't do it, then it fails at `npm run build`)

goapunk:
> It looks good, but I cannot test as the [mB branch](https://github.com/liqd/a4-meinberlin/pull/5052) is not installing with this a4 version. Any tips on how to make it work? (`--legacy-peer-deps` doesn't do it, then it fails at `npm run build`)

@fuzzylogic2000  should be fixed by https://github.com/liqd/a4-meinberlin/pull/5054

fuzzylogic2000 (commented):
In mB it works for the project overview, but not in the map modules (at least not the Bürgerhaushalt), there I get in the dashboard and on the participation view `AttributeError: 'Settings' object has no attribute 'A4_MAP_ATTRIBUTION'`

fuzzylogic2000:
Now it works in the dashboard, but still not the participation view.

goapunk:
> Now it works in the dashboard, but still not the participation view.

arg, sorry! I hopefully fixed all places now. Also it didn't always work, I modified the code to make it a bit more robust.

fuzzylogic2000 (approved):
Yes! It works everywhere in mB!

PR #1412, opened 2023-05-25
action: add target_creator to improve filtering

goapunk:
warning: contains migration


adhocracy4/actions/models.py
```diff
@@ -46,12 +46,19 @@ def exclude_updates(self):
...
-
     # actor, if actor is None the action was create by the system
     actor = models.ForeignKey(
         settings.AUTH_USER_MODEL, on_delete=models.CASCADE, blank=True, null=True
     )
 
+    target_creator = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        related_name="target_creator",
+        on_delete=models.CASCADE,
```
goapunk, line 57: do we want to cascade here? I guess so?

fuzzylogic2000: On the other hand, when we don't do that for some reason, the action should also stay.
I mean we have: action - target - target creator
Normally when the target creator is deleted that cascades though the target to the action. But if for some reason, we do not delete the target when the creator is deleted, we also shouldn't delete the action. So I would say we want SET_NULL
Even though, we do the same for the actor, but that makes more sense as a n action always has an object and an actor, but not always a target. Right?

adhocracy4/actions/migrations/0006_set_action_target_creator.py
```diff
@@ -0,0 +1,29 @@
...
+
+def set_action_target_creator(apps, schema_editor):
+    Action = apps.get_model("a4actions", "Action")
+    for action in Action.objects.filter(verb="add"):
+        if hasattr(action, "target_content_type") and action.target_content_type:
+            model = apps.get_model(
+                app_label=action.target_content_type.app_label,
+                model_name=action.target_content_type.model,
+            )
+            content_object = model.objects.get(id=action.target_object_id)
```
Rineee, line 14: Couldnt you just get it with action.target?

adhocracy4/actions/migrations/0006_set_action_target_creator.py
fuzzylogic2000, line 1: Ohhh, I also just thought, that this might be a migration again that blocks the DB for a while. We should find a way to make the whole migration not atomic (like slice it into bits that are atomic or sth.) and we should also test it on the prod DB somehow. :grimacing:

adhocracy4/actions/migrations/0006_set_action_target_creator.py
goapunk: no, generic relations / fks are not available in migrations :/

adhocracy4/actions/migrations/0006_set_action_target_creator.py
goapunk: It takes less than 2 mins on prod

fuzzylogic2000: Cool! And on mB?

fuzzylogic2000: https://docs.djangoproject.com/en/4.2/howto/writing-migrations/#non-atomic-migrations

adhocracy4/actions/migrations/0006_set_action_target_creator.py
Rineee: Ah, right! And annoyoing!

adhocracy4/actions/migrations/0006_set_action_target_creator.py
goapunk: so a+ is 30s and mb 1m10s on my machine

goapunk: I added batch updating so we don't need the explicit atomic

fuzzylogic2000 (commented):
Very nice! 
Just two more little things: can you please squash the commits, and add a backwards function (just emptying the fields again)? Then, if we do fuck up, there's still a way back.

PR #1420, opened 2023-06-01
labels/model: add title and description field, categories/model: add …

philli-m:
…title and description field


philli-m:
@fuzzylogic2000 is this what you meant and then it'll be added to template in mB?

fuzzylogic2000:
> @fuzzylogic2000 is this what you meant and then it'll be added to template in mB?

Yes, nearly. The fields and form are good! But it needs to go in an extra model. The Category model has one entry for every category like "done", "in progress", ... and are then filtered by module. The title and description shouldn't be added to every category, but only once for each module. So an extra CategoryTitle (or some other name) model with the two fields you added would be good.

philli-m:
@fuzzylogic2000 ahh ok cheers, I've updated the task and will add a new model

adhocracy4/categories/models.py
```diff
@@ -27,6 +28,49 @@ def formfield(self, **kwargs):
...
+            "The description text is displayed to the "
+            "participants as help text when they have to "
+            "assign their ideas. The character limit is max. "
+            "300 characters (with spaces)."
+        ),
+    )
+    module = models.ForeignKey(
+        module_models.Module,
+        on_delete=models.CASCADE,
+    )
```
philli-m, line 64: we only want one per module so thought this was correct?

adhocracy4/categories/models.py
```diff
@@ -27,6 +28,49 @@ def formfield(self, **kwargs):
...
+
+    class Meta:
+        verbose_name_plural = "categorie alias"
+        ordering = ["pk"]
+
+    def __str__(self):
+        return self.name
+
```
philli-m, line 72: not sure if need to have the return?

adhocracy4/categories/dashboard.py
```diff
@@ -14,8 +14,23 @@ class ModuleCategoriesComponent(ModuleFormSetComponent):
...
+
+    def get(self, request, *args, **kwargs):
+        self.object = None
+        form_class = self.get_form_class()
+        form = self.get_form(form_class)
+        category_alias_form = forms.CategoryAliasForm()
+        return self.render_to_response(
+            self.get_context_data(form=form, category_alias_form=category_alias_form)
+        )
+
```
philli-m, line 33: @fuzzylogic2000 @Rineee this is my attempt to add the second form to the context, tested in mB https://github.com/liqd/a4-meinberlin/pull/5095/files however currently failing, does either one of you want to tag in?

adhocracy4/categories/forms.py
```diff
@@ -30,6 +30,15 @@ def show_categories(self):
         return module_has_categories
 
 
+class CategoryAliasForm(forms.ModelForm):
```
fuzzylogic2000, line 33: I changed this from ModuledashboardForm to the ModelForm (also for the labels) and remobed the second commit. The rest stayed the same.
The changes for the dashboard component are only added in mB now. We need to add it to a4 when we want to have it in a+ as well, and then we probably need to make sure it still works in opin and overwrite there.

Rineee (approved):
Really nice!

PR #1422, opened 2023-06-06
Jd 2023 06 comments add project fk

goapunk:


adhocracy4/comments/models.py
```diff
@@ -39,6 +39,9 @@ class Comment(base.UserGeneratedContentModel):
     is_moderator_marked = models.BooleanField(default=False)
     # used in moderation dashboard, indicates if comment is marked as read by moderator
     is_reviewed = models.BooleanField(default=False)
+    project = models.ForeignKey(
+        Project, on_delete=models.CASCADE, blank=True, null=True
```
Rineee, line 43: Do we really want blank here? Probably every comment should have the project set, right?

goapunk: yes and no, some comments don't have the module set, so getting the project in the migration fails. I'm not sure why but maybe if a module or project is deleted? Do we keep the comments in that case? At least we probably don't want the cascade here I think?

adhocracy4/comments/migrations/0013_set_project.py
```diff
@@ -0,0 +1,46 @@
...
+        )
+        content_object = model.objects.get(id=comment.object_pk)
+        if hasattr(content_object, "module"):
+            comment.project = content_object.module.project
+            comments.append(comment)
+            count += 1
+        if count == BATCH_SIZE:
+            Comments.objects.bulk_update(comments, ["project"])
+            count = 0
+            actions = []
```
Rineee, line 25: this one is not needed, right?

goapunk (commented):
really nice

adhocracy4/comments/models.py
goapunk: @Rineee @fuzzylogic2000 last thing we should check: do we want the cascade there? I guess we don't (if it should be deleted it will happen it will come from somewhere else) ?

Rineee: Hm, yeah, probably it will be deleted anyway then.. but still we could keep it here, I would say? What do you think @fuzzylogic2000 ?

adhocracy4/comments/models.py
```diff
@@ -60,6 +63,8 @@ def __str__(self):
     def save(self, *args, **kwargs):
         """Change comment.comment if comment was marked removed or censored."""
         self.comment = transforms.clean_html_all(self.comment)
+        if not self.project:
```
goapunk, line 66: @Rineee do we need  to distinguish between comment and chapter here as well?

Rineee: I think this should work, because both comments and paragraphs have a module property that then goes to parent or chapter.

adhocracy4/comments/models.py
Rineee: Sure, I am also fine with `SET_NULL` :)

PR #1425, opened 2023-06-13
[#7412] images/fields: add alt text image field to project form

philli-m:


philli-m:
@fuzzylogic2000 @Rineee this is the related a4 to the mB image PR, if this is correct i'll add migration

adhocracy4/images/mixins.py
```diff
@@ -0,0 +1,61 @@
...
+        max_length=80,
+        help_text=format_lazy(
+            _(
+                "An alternate text serves as a textual description "
+                "of the image content and is read out by screen "
+                "readers. Describe the image in max. 80 characters. "
+                "Example: A busy square with people in summer."
+            ),
+            image_name="image",
+        ),
```
philli-m, line 43: not sure if this is valid, i just kinda mushed it in and the image_name is just a stan-in as wanted to get it kind of running so was just putting thing in so it would start

adhocracy4/images/mixins.py
```diff
@@ -0,0 +1,61 @@
...
+    )
+
```
Rineee, line 45: I think you dont need to add/define the fields here, they should be there just from defining them in the model and then adding them to the form, so like [here](https://github.com/liqd/adhocracy4/blob/e864a4e2d505fa19c586f406f9b98944b0afe2a2/adhocracy4/dashboard/forms.py#L28), but I am not sure which the right form is here, you probably also have to add it manually to all the external, container etc forms.. 
So in the mixin only the __innit__ with the checks should be needed. And then you add the mixin to the ProjectCreateForms.

adhocracy4/images/forms.py
```diff
@@ -1,7 +1,9 @@
 from django.forms.fields import ImageField
 
+from adhocracy4.images.mixins import ImageMetadataMixin
+
 from . import widgets
 
 
-class ImageField(ImageField):
+class ImageField(ImageField, ImageMetadataMixin):
```
Rineee, line 8: I think you need to add the mixin in the ProjectCreateForms instead.

adhocracy4/images/mixins.py
```diff
@@ -0,0 +1,61 @@
...
+        image_copyright = cleaned_data.get("image_copyright")
+        image_alt_text = cleaned_data.get("image_alt_text")
+        if image and not image_copyright:
+            self.add_error(
+                "right_of_use",
+                _("Please add copyright information."),
+            )
+        if image and not image_alt_text:
+            self.add_error(
+                "right_of_use",
```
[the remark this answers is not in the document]
Rineee: Not sure if that works with the lazy formatting for header_image and tile_image. Maybe if not, just do the checks for both of them after another?

adhocracy4/images/mixins.py
Rineee: I think I found the form, its in A4 [here](https://github.com/liqd/adhocracy4/blob/e864a4e2d505fa19c586f406f9b98944b0afe2a2/adhocracy4/dashboard/forms.py#L62). If we add it there, we also add it to all other projects, so we probably have to overwrite again in mB :frowning_face:

adhocracy4/images/fields.py
```diff
@@ -80,39 +80,39 @@ def _help_text(self):
...
+#                     "readers. Describe the image in max. 80 characters. "
+#                     "Example: A busy square with people in summer."
+#                 ),
+#                 image_name=image_name,
+#             ),
+#             "max_length": 80,
+#             "blank": True,
+#         }
+#         defaults.update(kwargs)
+#         super().__init__(*args, **defaults)
```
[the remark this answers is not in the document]
philli-m: ahhhh I thought that was one and the same but that makes sense :)

adhocracy4/images/mixins.py
```diff
@@ -0,0 +1,61 @@
...
+    )
+
+    def clean(self):
+        cleaned_data = super().clean()
+        image = cleaned_data.get("image")
+        image_copyright = cleaned_data.get("image_copyright")
+        image_alt_text = cleaned_data.get("image_alt_text")
+        if image and not image_copyright:
+            self.add_error(
+                "right_of_use",
```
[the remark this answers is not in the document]
philli-m: sorry hangover brain

adhocracy4/images/mixins.py
philli-m: @Rineee do you mean a+ ect.? I think as we want to have a default for all old images it's probably fine as we will want it in a+ for ally, although i think it's best practice to leave it blank if it's decorative but again it is not used like that so having alt is probably better then nothing

adhocracy4/images/mixins.py
```diff
@@ -0,0 +1,59 @@
...
+    def __init__(self, *args, image="image", tile_image="tile_image", **kwargs):
+        self.image = image
+        self.tile_image = tile_image
+        super().__init__(*args, **kwargs)
+
+        if not self.instance.image:
+            self.initial["image_copyright"] = ""
+            self.initial["image_alt_text"] = ""
+        if self.instance.image and not self.instance.image_copyright:
+            self.initial["image_copyright"] = _("Copyright of project owner")
```
philli-m, line 15: This is not the final wording and it doens't actually work like a default, it just adds it when the image is saved rather then adding it to the old projects, I assume I am doing this wrong?

adhocracy4/images/mixins.py
```diff
@@ -0,0 +1,59 @@
...
+
+class ImageMetadataMixin(forms.ModelForm):
+    def __init__(self, *args, image="image", tile_image="tile_image", **kwargs):
+        self.image = image
+        self.tile_image = tile_image
+        super().__init__(*args, **kwargs)
+
+        if not self.instance.image:
+            self.initial["image_copyright"] = ""
+            self.initial["image_alt_text"] = ""
```
philli-m, line 13: did this so it deleted the entry when the img is deleted as made sense to me?

goapunk:
@philli-m needs a Changelog entry ;-)

CHANGELOG.md
```diff
@@ -7,3 +7,6 @@ This project (not yet) adheres to [Semantic Versioning](https://semver.org/spec/
 
 
 ## Unreleased
+* [#7412] images/fields: add alt text image field to project form
+  * Add alt text field to project model and form.
```
goapunk, line 11: the leading whitespace here and next line is redundant

philli-m:
@goapunk done, you can now review :D

goapunk (approved):
LGTM

PR #1429, opened 2023-06-22
Jd 2023 06 housekeeping

goapunk:


.flake8
```diff
@@ -0,0 +1,5 @@
+[flake8]
```
goapunk, line 1: annoyingly flake8 doesn't support reading the config from pyproject.toml so we have to move back to a separate file

philli-m (approved):
coolio!

PR #1436, opened 2023-06-28
[7385] Ks 2023 06 use category label alias

Rineee:


Rineee:
@goapunk @philli-m I added tests, although they dont do much, because we dont have a real view here in A4 that uses the new filter.. so will properly tests in mB then once this merged!

philli-m (approved):
Looking good! just the rebase issue but we can talk about that next

PR #1437, opened 2023-06-28
actions: only create actions for phases and project if the project is

goapunk:
not a draft

see https://github.com/liqd/a4-meinberlin/issues/4922


Rineee:
Nice!

PR #1443, opened 2023-07-04
[7428] projects: add intial_slide to view context of module detail views

Rineee:
Ok, this story is actually still in the backlog, was checking it together with @goapunk to see if is good for onboarding, but seemed a bit too complicated to find the right places.
But maybe @m4ra and @hklarnerliqd can review :) It needs another part in Aplus, will also tag you there!

### DIFF
--- a/adhocracy4/projects/models.py
+++ b/adhocracy4/projects/models.py
@@ -152,6 +152,14 @@ class TimelinePropertiesMixin:
             return []
         return []
 
+    def get_module_cluster_index(self, module):
+        for idx, val in enumerate(self.participation_dates):
+            if "modules" in val and module in val["modules"]:
+                return idx
+        if self.project.get_current_participation_date():
+            return self.project.get_current_participation_date()
+        return 0
+
 
 class Project(
     ProjectContactDetailMixin,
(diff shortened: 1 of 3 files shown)


adhocracy4/projects/models.py
```diff
@@ -152,6 +152,14 @@ def get_current_modules(self, idx):
...
         return []
 
+    def get_module_cluster_index(self, module):
+        for idx, val in enumerate(self.participation_dates):
+            if "modules" in val and module in val["modules"]:
+                return idx
+        if self.project.get_current_participation_date():
+            return self.project.get_current_participation_date()
+        return 0
+
```
m4ra, line 162: Is in that order bcz participation_dates may both exist? so we prioritise the modules over the project's?

Rineee: So the order is because first, I try to find the module cluster that the module is in, if that doesnt work, I take the module cluster that is the 'active' one (this is what get_current_participation_date() does, I copied it from initial_slide() in DisplayProjectOrModuleMixin). It is more like a fallback.

adhocracy4/projects/models.py
```diff
@@ -152,6 +152,14 @@ def get_current_modules(self, idx):
             return []
         return []
 
+    def get_module_cluster_index(self, module):
+        for idx, val in enumerate(self.participation_dates):
+            if "modules" in val and module in val["modules"]:
+                return idx
+        if self.project.get_current_participation_date():
+            return self.project.get_current_participation_date()
+        return 0
```
hklarner, line 161: You could also use
```python
return self.project.get_current_participation_date() or 0
```
- calls `get_current_participation_date()` only once

hklarner:
looks good to me

PR #1459, opened 2023-07-25
comments_async: fix broken scroll and tos callback

goapunk:
more of a hotfix for now


hklarner:
looks good to me + tested locally

PR #1473, opened 2023-08-01
comments_async: make fetchComments take arguments instead of relying

goapunk:
on state variables. Should fix the issue with the scrolling not working when an anchor is passed.


hklarner:
looks good to me

PR #1477, opened 2023-08-09
[7543] extends report serializer

hklarner:


adhocracy4/reports/serializers.py
```diff
@@ -1,9 +1,24 @@
 from rest_framework import serializers
 
+from adhocracy4.api.dates import get_datetime_display
+
 from .models import Report
 
 
 class ReportSerializer(serializers.ModelSerializer):
+    username = serializers.CharField(required=False, source="creator.username")
+    created_at = serializers.SerializerMethodField(required=False)
```
goapunk, line 10: maybe just use `created` here (and below) as we also use that in other places (e.g. comments)

goapunk (changes requested):
generally, it looks good to me. One small comment and it needs a changelog

adhocracy4/reports/serializers.py
hklarner: done

adhocracy4/reports/serializers.py
```diff
@@ -1,9 +1,24 @@
...
 from .models import Report
 
 
 class ReportSerializer(serializers.ModelSerializer):
+    username = serializers.CharField(required=False, source="creator.username")
+    created = serializers.SerializerMethodField(required=False)
+
+    @staticmethod
+    def get_created(report):
+        return get_datetime_display(datetime=report.created)
```
m4ra, line 14: we don't need that, see my comment above.

m4ra (changes requested):

adhocracy4/reports/serializers.py
m4ra: oh, is it for returning the date in a specific format? if yes then it makes sense. But take the staticmethod decorator away, and replace `report` with `self`.

adhocracy4/reports/serializers.py
```diff
@@ -1,9 +1,24 @@
 from rest_framework import serializers
 
+from adhocracy4.api.dates import get_datetime_display
+
 from .models import Report
 
 
 class ReportSerializer(serializers.ModelSerializer):
+    username = serializers.CharField(required=False, source="creator.username")
+    created = serializers.SerializerMethodField(required=False)
```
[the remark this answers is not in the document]
hklarner: resolved (`created` needs to be optional)

adhocracy4/reports/serializers.py
hklarner: done

m4ra (approved):

PR #1488, opened 2023-09-05
[7601] Install and configure Celery

hklarner:
to do:
- [x] replace old background tasks decorator with new shared task decorator
- ~custom migration that removes the existing tables of the un-installed app `background_tasks`~

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -16,3 +16,4 @@ python-dateutil==2.8.2
 python-magic==0.4.27
 rules==3.3
 XlsxWriter==3.1.2
+celery==5.3.1
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ coverage/
 
 # Local instance specific settings
 /tests/project/local.py
+
+build/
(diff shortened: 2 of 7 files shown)


requirements/base.txt
```diff
@@ -16,3 +16,5 @@ python-dateutil==2.8.2
 python-magic==0.4.27
 rules==3.3
 XlsxWriter==3.1.2
+celery==5.3.1
+redis==5.0.0
```
goapunk, line 20: we might not need redis here as a4 doesn't use it / is transcient to which celery broker is used. The actual celery config will be in a+/mb, so maybe we just add it there?

hklarner: done

m4ra (approved):

m4ra:
@hklarner can you also add the changelog?

m4ra:
We said the coverage drop is very little, so we are good to merge.

goapunk (approved):
LGTM

m4ra (approved):

PR #1490, opened 2023-09-07
[TB 92 and 37]comments_async: update semantics and implement some fixes from a11y test

philli-m:
Some fixes for the tech backlog stories concerning the comments, has been tested with a+ main, may need additional fixes when mB is updated but can be done in design sprint


goapunk (approved):
looks good to me!

PR #1493, opened 2023-09-19
makes send_async serializable for celery workers 

hklarner:


adhocracy4/emails/tasks.py
```diff
@@ -1,16 +1,46 @@
...
+    elif object_pk is not None:
+        log.info(f"{task_id=}, {model_name=}, {object_pk=}")
+
+        model = apps.get_model(app_label, model_name)
+        obj = model.objects.get(pk=object_pk)
+        email_class().dispatch(obj, *args, **kwargs)
 
-    return cls().dispatch(obj, *args, **kwargs)
+    else:
+        log.error(f"called send_async without object_pk or data: {task_id=}")
```
m4ra, line 46: `dispatch` is expecting an object for the context, and context is then passed to the method `render` for the email template. if data is a dict, how will this work out?
https://github.com/liqd/adhocracy4/blob/ca6bdd995d74b380fee8a3f2ab971c6d57301b67/adhocracy4/emails/base.py#L56C1-L56C1

I would suggest we change what dispatch is returning, rather than omit return statements here. So it's good to know why dispatch is returning a list of mail objects.
In any case we need extra tests for covering the sending of emails better here and in a+

hklarner: > if data is a dict, how will this work out?

It is already the case that it works out today. The line you point to works fine with non-Django models, i.e. a dictionary:
```python
>>> d = dict()
>>> d.__class__.__name__.lower()
'dict'
```

More importantly, it is already used in this way today, because `send_no_object` also calls `dispatch`, see:
- https://github.com/liqd/adhocracy-plus/blob/bd33a7de561e45415b7585560ef153bf1585194c/apps/projects/tasks.py#L10

Note that the name `object` is misleading here because it is in fact just a data dict (that's why I suggest to change it to `data`). Here is an example of what `object` looks like in that function:
```python
{'initiators': ['admin@liqd.net'], 'name': 'yes', 'organisation_id': 6}
```

> So it's good to know why dispatch is returning a list of mail objects.

I agree, but it looks very much like there is no good reason. The function `send_async_no_object` is clearly just copy-pasted from `send_async` and modified (instead of modifying `send_asnyc` directly).

I have tested non-returning versions of `send_async` with the following events locally and received the emails:
- Feedback for your contribution
- Confirm your email address
- Deletion of project

m4ra: It works for dict object but is not returning anything meaningful for the email template is what I mean. But you pointed out that this is happening already, so I guess the email template doesn't need to render the object.

So, since this send_async method is not supposed to be passed an object as it is anyhow, but an object pk. To make it more generic, I suggest then change the name of object_pk to obj_data (and type hinting as expecting two types) and then check if this is int or dict inside the code and proceed accordingly, rather than adding an extra arg `data`. 

For not returning the dispatch, then this send_async will return an empty list right? And for a quick look in a4 and a+ don't see an issue. Do the tests pass?
https://github.com/liqd/adhocracy4/blob/ca6bdd995d74b380fee8a3f2ab971c6d57301b67/adhocracy4/emails/base.py#L90

m4ra: Or we could return the mails list as string in dispatch()

hklarner: > For not returning the dispatch, then this send_async will return an empty list right? And for a quick look in a4 and a+ don't see an issue. Do the tests pass?

I think it shouldn't return anything. I've grepped across al projects. The return is never assigned to anything. The line you point to is after the call of `send_async`. (the call of `send_async` is [here](https://github.com/liqd/adhocracy4/blob/ca6bdd995d74b380fee8a3f2ab971c6d57301b67/adhocracy4/emails/base.py#L81)).

Yes, tests pass. I've also tested the management command `send_test_emails` (it works fine) but this one uses another Email class that uses a mixin for synchronous sending which avoids the celery / encoding issue altogether.

> To make it more generic, I suggest then change the name of object_pk to obj_data (and type hinting as expecting two types) and then check if this is int or dict inside the code and proceed accordingly, rather than adding an extra arg data.

I can do that if you want. But note that the caller _always knows_ whether a real model instance or just some data is dispatched. It would therefore add clarity if it is explicitly passed to the function accordingly. I.e. it is clearer to use single purpose arguments than to overload one parameter and then check its type in the function body. But this is just my opinion, you can decide.

hklarner: EDIT: I am also suggesting that we start to use keyword-based function calls more if possible, rather than position-based calls. It makes the intention of the caller and also the extension by more arguments in the future easier ..

hklarner: I have changed the `if-then-else` to check first for missing arguments (talked with @m4ra)..

hklarner: I have removed the `is not None` check for `object_pk` (discussed with @m4ra).

adhocracy4/emails/tasks.py
```diff
@@ -1,16 +1,48 @@
...
+
+    args = args or []
+    kwargs = kwargs or {}
+
+    log.info(f"{task_id=}, {email_module_name=}, {email_class_name=}")
+
+    email_module = importlib.import_module(email_module_name)
+    email_class = getattr(email_module, email_class_name)
+
+    if data is not None:
```
goapunk, line 38: this can just be `if data:` I think, right?

adhocracy4/emails/tasks.py
```diff
@@ -1,16 +1,48 @@
...
+        log.info(f"{task_id=}, {data=}")
+
+        email_class().dispatch(data, *args, **kwargs)
+
+    elif object_pk:
```
goapunk, line 43: and this just `else` ?

adhocracy4/emails/tasks.py
hklarner: done.

adhocracy4/emails/tasks.py
```diff
@@ -1,16 +1,48 @@
...
-    email_module_name, email_class_name, app_label, model_name, object_pk, args, kwargs
+    self: Task,
+    email_module_name: str,
+    email_class_name: str,
+    app_label: str,
+    model_name: str,
+    object_pk: Optional[int] = None,
+    args: Optional[list] = None,
+    kwargs: Optional[dict] = None,
+    data: Optional[dict] = None,
```
m4ra, line 22: Since we refactoring, I would use the kwargs for the data (I mentioned it in our DM), otherwise, the data after the kwargs while both are dicts is redundant. Name it **kwargs or **data, as you prefer. The kwargs are passed to the context I saw, so that should be working.
Then put back the asterisks that denote what type those args and kwargs/data are, and skip the hinting for those, except we are sure what type args and kwargs expected to be inside the tuple/dict, but we aren't. [PEP484](https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values)
 In emails/base.py the dispatch() and get_context() need refactoring too, to work with object optionally. If it stays as it is, code remains confusing, as we pass data which is a dict, and then is called object in EmailBase.

adhocracy4/emails/tasks.py
```diff
@@ -1,16 +1,48 @@
...
+
+    log.info(f"{task_id=}, {email_module_name=}, {email_class_name=}")
+
+    email_module = importlib.import_module(email_module_name)
+    email_class = getattr(email_module, email_class_name)
+
+    if data:
+        log.info(f"{task_id=}, {data=}")
+
+        email_class().dispatch(data, *args, **kwargs)
```
m4ra, line 41: here either **kwargs or **data, no need for both. Depending which one name you chose, you may need to update the if statements above.

adhocracy4/emails/tasks.py
```diff
@@ -1,16 +1,48 @@
...
-    model = apps.get_model(app_label, model_name)
-    obj = model.objects.get(pk=object_pk)
+    task_id = self.request.id[:5]
+
+    if not (data or object_pk):
+        log.error(f"called send_async without object_pk or data: {task_id=}")
+        return
+
+    args = args or []
+    kwargs = kwargs or {}
```
m4ra, line 31: no need for these two once you put back the asteriscs in the function definition

hklarner: we agreed to do any real refactoring in a separate story and focus on making a4 work with celery in this PR.

adhocracy4/emails/tasks.py
hklarner: we agreed to do any real refactoring in a separate story and focus on making a4 work with celery in this PR.

goapunk (approved):
LGTM. Will leave for @m4ra to give a final review and merge

m4ra (approved):

PR #1499, opened 2023-10-18
[7274] migrate from django-ckeditor to django-ckeditor-5

goapunk:


adhocracy4/ckeditor/fields.py
```diff
@@ -1,23 +1,10 @@
...
 
+# FIXME: remove these fields / file
 class RichTextCollapsibleMixin:
-    def __init__(self, *args, **kwargs):
-        kwargs["extra_plugins"] = kwargs.get("extra_plugins", []) + _extra_plugins
-        kwargs["external_plugin_resources"] = (
-            kwargs.get("external_plugin_resources", []) + _external_plugin_resources
-        )
-        super().__init__(*args, **kwargs)
+    pass
```
goapunk, line 9: not sure how to get rid of these fields, as they will break old migrations. One way would be to squash / reset migrations but that also has its drawbacks

goapunk: @m4ra any idea here?

m4ra: The fields that will be deleted, the data they retain they don't need to be migrated to the new CK?

goapunk: They are migrated to `CKEditor5Field`. Will look into squashing, thanks!

philli-m (approved):
s

philli-m (approved):
it's definitely working, also displays old content too but as it's quite pythony, maybe @m4ra you want to give it a final look over?

adhocracy4/ckeditor/storage.py
```diff
@@ -0,0 +1,12 @@
...
+from urllib.parse import urljoin
+
+from django.conf import settings
+from django.core.files.storage import FileSystemStorage
+
+
+class CustomStorage(FileSystemStorage):
+    """Custom storage to store uploads in a subfolder called uploads"""
+
+    location = os.path.join(settings.MEDIA_ROOT, "uploads")
```
m4ra, line 11: I'd specify those are from the ck editor uploads, so maybe a prefix or a suffix to `uploads`

m4ra (commented):
LGTM

adhocracy4/ckeditor/fields.py
```diff
@@ -1,23 +1,10 @@
...
-_external_plugin_resources = [
-    (
-        "collapsibleItem",
-        "/static/ckeditor_collapsible/",
-        "plugin.js",
-    )
-]
-
 
+# FIXME: remove these fields / file
```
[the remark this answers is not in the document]
goapunk: They are empty (pass), but still there. The problem is that removing them breaks some migrations, so we need to squash them (I would do this in a separate PR when removing the old django-ckeditor package)

adhocracy4/ckeditor/storage.py
goapunk: This breaks existing behavior though, the old ckeditor uploads them to `uploads`. I'd prefer to keep the current behavior for now. We have another story which will touch the upload behavior (file names, path), so we can change it as part of that it that's ok?

PR #1504, opened 2023-11-07
[7776] django upgrade to 4.2 (LTS)

m4ra:
**NOT merge until the other repos are ready** 

Django 4 has dropped support for sha1 hashes. The deprecation warning was there before Django v3. 
From the release notes:

- Support for the pre-Django 3.1 password reset tokens in the admin site (that use the SHA-1 hashing algorithm) will be removed.
- Support for the pre-Django 3.1 django.core.signing.Signer signatures (encoded with the SHA-1 algorithm) will be removed.
- Support for the pre-Django 3.1 django.core.signing.dumps() signatures (encoded with the SHA-1 algorithm) in django.core.signing.loads() will be removed.
- Support for the pre-Django 3.1 user sessions (that use the SHA-1 algorithm) will be removed.

@goapunk will that affect existing passwords? or was it dealt previously while upgrading to 3.2? Cannot find if this is dealt by Django itself.
See transitional notes [here](https://docs.djangoproject.com/en/4.2/releases/3.1/#default-hashing-algorithm-settings), though is about updating multiple instances of same project.

Notes:
DONE: _need to update django-allauth for 4.2_


goapunk:
> Django 4 has dropped support for sha1 hashes. The deprecation warning was there before Django v3. From the release notes:
> 
>     * Support for the pre-Django 3.1 password reset tokens in the admin site (that use the SHA-1 hashing algorithm) will be removed.
> 
>     * Support for the pre-Django 3.1 django.core.signing.Signer signatures (encoded with the SHA-1 algorithm) will be removed.
> 
>     * Support for the pre-Django 3.1 django.core.signing.dumps() signatures (encoded with the SHA-1 algorithm) in django.core.signing.loads() will be removed.
> 
>     * Support for the pre-Django 3.1 user sessions (that use the SHA-1 algorithm) will be removed.
> 
> 
> @goapunk will that affect existing passwords? or was it dealt previously while upgrading to 3.2? Cannot find if this is dealt by Django itself. See transitional notes [here](https://docs.djangoproject.com/en/4.2/releases/3.1/#default-hashing-algorithm-settings), though is about updating multiple instances of same project.
> 
> Notes: need to update django-allauth for 4.2

I'm not entirely sure, but worst case the very few affected users would have to request a new password via `forgot password`. So I think that's ok. I also vaguely remembered checking for passwords with old hashes in the db and didn't find any.

goapunk (changes requested):
@m4ra there are some instances of `index_together` which should be replaced with `indexes` as described here: https://docs.djangoproject.com/en/4.2/releases/4.2/

m4ra:
> @m4ra there are some instances of index_together which should be replaced with indexes as described here: https://docs.djangoproject.com/en/4.2/releases/4.2/#index-together-option-is-deprecated-in-favor-of-indexes
@goapunk done, and rebased. Please have another look.

adhocracy4/modules/models.py
```diff
@@ -95,7 +93,7 @@ class Module(models.Model):
 
     objects = ModulesQuerySet.as_manager()
 
-    blueprint_type = BlueprintTypeField(
+    blueprint_type = models.CharField(
         max_length=255,
         blank=True,
     )
```
m4ra, line 99: then we delete the fields with the custom BlueprintType?
Also the new migration cancels the 0008. If we remove 0008, it won't generate 0009 here
@goapunk

Makefile
```diff
@@ -45,15 +45,11 @@ lint:
 	$(VIRTUAL_ENV)/bin/isort --diff -c $(SOURCE_DIRS) ||  EXIT_STATUS=$$?; \
 	$(VIRTUAL_ENV)/bin/flake8 $(SOURCE_DIRS) --exclude migrations,settings ||  EXIT_STATUS=$$?; \
 	npm run lint ||  EXIT_STATUS=$$?; \
-	$(VIRTUAL_ENV)/bin/python manage.py makemigrations --dry-run --check --noinput || EXIT_STATUS=$$?; \
```
goapunk, line 48: we don't need to remove these anymore, right?

adhocracy4/categories/form_fields.py
```diff
@@ -1,34 +1,10 @@
...
-
 from django import forms
+from django.conf import settings
 from django.forms import widgets
-from django.utils.functional import cached_property
 
 from adhocracy4.categories import get_category_icon_url
 
 
-class CategoryIconDict(abc.Mapping):
```
goapunk, line 10: I'm not really sure why this was needed so I removed it, I think we should point out to testers to verify everything still works as intended with the icons

adhocracy4/projects/migrations/0043_migrate_topics_to_m2m_topics.py
```diff
@@ -0,0 +1,38 @@
...
+
+def add_topics_to_m2m_table(apps, schema_editor):
+    if hasattr(settings, "A4_PROJECT_TOPICS"):
+        topicenum = settings.A4_PROJECT_TOPICS
+        project = apps.get_model("a4projects", "Project")
+        for project in project.objects.all():
+            for topic_code in project.topics:
+                project.m2mtopics.create(
+                    code=topic_code,
+                    name=[item[1] for item in topicsenum if item[0] == topic_code][0],
```
goapunk, line 15: topicsenum -> topicenum

goapunk (changes requested):
Looks good to me, we only need changelogs for the latest changes (topics, categories?, blueprint_types?) and maybe some of my commits should be squashed

m4ra:
> Looks good to me, we only need changelogs for the latest changes (topics, categories?, blueprint_types?) and maybe some of my commits should be squashed

@goapunk done, have another look please.
FYI I kept the models changes related to choices in separate commits in case we need to revert.

goapunk (approved):
really nice!

PR #1514, opened 2023-12-07
Use topics enum instead of tuples

m4ra:
**Tasks**
- [ ] Changelog


adhocracy4/projects/migrations/0043_migrate_topics_to_m2m_topics.py
```diff
@@ -1,19 +1,25 @@
...
 
 
 def add_topics_to_m2m_table(apps, schema_editor):
     if hasattr(settings, "A4_PROJECT_TOPICS"):
-        topicsenum = settings.A4_PROJECT_TOPICS
+        # get the module from the settings string path without the class
+        index = settings.A4_PROJECT_TOPICS.rfind(".")
+        # get the class name from the settings string path
+        module = importlib.import_module(settings.A4_PROJECT_TOPICS[:index])
+        topicsenum = getattr(module, settings.A4_PROJECT_TOPICS[index + 1 :])
```
goapunk, line 15: This seems a bit complicated and raises the question whether we need the `name` of the topic in the database. For our purposes the `code` would be enough, right?

And if we need the label (e.g. in the template or via the api) we can get it via `topicsenum(code).label`, for example in the serializer or by overwriting the `def _str_` of the Topic model.

m4ra: I agree, we don't need the name. I will remove it then.

adhocracy4/projects/models.py
```diff
@@ -25,13 +26,12 @@
...
-    name = models.CharField(
-        max_length=120,
-        verbose_name=_("Topic"),
-    )
 
     def __str__(self):
-        return self.name
+        if hasattr(settings, "A4_PROJECT_TOPICS"):
+            topics_enum = import_string(settings.A4_PROJECT_TOPICS)
+            return str(topics_enum(self.code).label)
```
m4ra, line 33: Nice!

goapunk (approved):
LGTM!

PR #1515, opened 2023-12-11
[7885] replace multiselect with m2m relation for project topics

goapunk:
**Tasks**
- [x] Documentation (docs and inline)
- [ ] Changelog


m4ra (approved):
LGTM. I would like to add a docs about how we handle topics for projects which make use of this. I will add an extra commit to this PR.

adhocracy4/projects/fields.py
```diff
@@ -1,13 +1,12 @@
-from multiselectfield import MultiSelectField
+from django.db import models
 
 
-class TopicField(MultiSelectField):
+class TopicField(models.CharField):
     """Deprecated, don't use"""
 
     # TODO: remove once topic migrations are rolled out
     def __init__(self, *args, **kwargs):
```
goapunk, line 8: should we remove all methods and just add `pass` as it's unused anyway?

PR #1517, opened 2024-01-23
[7274] projects/model: update helptext for information and result field

goapunk:
**Tasks**
- [x] Changelog

partly fixes #https://github.com/liqd/a4-meinberlin/issues/5395


m4ra (changes requested):
Field's attributes shall be changed from ModelForms as a best practice, rather than generating model migrations. Did you check if you could do so in this case?

m4ra:
@goapunk you can add the help_text here: https://github.com/liqd/adhocracy4/blob/main/adhocracy4/dashboard/forms.py#L101

adhocracy4/dashboard/forms.py
```diff
@@ -121,6 +128,19 @@ class Meta:
...
+            "The results description will be shown in the result-tab. "
+            "Please describe a participation promise beforehand "
+            "(what will happen with the outcome?) and inform afterwards "
+            "about the outcome. If you add an image, please provide "
+            "an alternate text. It serves as a textual description "
+            "of the image content and is read out by screen readers. "
+            "Describe the image in approx. 80 characters. "
+            "Example: A busy square with people in summer."
+        )
+
```
m4ra, line 143: There is a `help_texts` dict you can add in Meta:
https://docs.djangoproject.com/en/4.2/topics/forms/modelforms/#overriding-the-default-fields

and don't forget to remove the migrations :)

goapunk: the migrations are there to remove the old help_text on the model, otherwise I feel it will be confusing having old one on the model and a new one on the form

m4ra (approved):

PR #1519, opened 2024-01-23
comments_async/comment_box: fix comment count plural not updating pro…

goapunk:
…perly

The comment count (on proposals for example) would only have the correct plural after refreshing. This should fix it

**Tasks**
- [x] Changelog


philli-m (approved):
Cool!

PR #1520, opened 2024-01-29
[7274] images: add validator which enforces alt texts on img tags

goapunk:
Adds an alt text validator to project information and  result forms in dashboard. Submitting the ckeditor form should fail if no alt text was set on an image

partly fixes https://github.com/liqd/a4-meinberlin/issues/5394

**Tasks**
- [x] Changelog


adhocracy4/images/validators.py
```diff
@@ -74,3 +76,21 @@ def validate_image(
...
+    message = _("Please add an alternative text for all images.")
+
+    def __call__(self, value):
+        soup = BeautifulSoup(value, "html.parser")
+        images = soup("img", alt=False)
+        if len(images) > 0:
+            raise ValidationError(self.message)
+
+    def __eq__(self, other):
+        return isinstance(other, self.__class__)
```
m4ra, line 96: Here preferably refer to the name directly than `self.__class__`

adhocracy4/images/validators.py
```diff
@@ -74,3 +76,21 @@ def validate_image(
     if errors:
         raise ValidationError(errors)
     return image
+
+
+@deconstructible
+class ImageAltTextValidator:
+    """Validate that if the input contains html img tags that all have the alt
+    attribute set, otherwise raise ValidationError.
+    """
```
m4ra, line 85: It needs attributes `message`, `code`, so the user knows what to fix (e.g alt in img tag is missing) , then adding these in the `__eq__` function. And is missing an `__init__`.
The inline docs better be placed under the call function.

adhocracy4/images/validators.py
```diff
@@ -74,3 +76,21 @@ def validate_image(
...
+
+    message = _("Please add an alternative text for all images.")
+
+    def __call__(self, value):
+        soup = BeautifulSoup(value, "html.parser")
+        images = soup("img", alt=False)
+        if len(images) > 0:
```
m4ra, line 92: what is this checking? not obvious, mb add a comment?

m4ra (changes requested):

adhocracy4/images/validators.py
goapunk: what purpose does the code have? I can't really figure it out. And as code and messages are always the same, do we need to check for them in `__eq__` ? They will always match.

m4ra: code is `invalid`.
You can see how they do it in [django valiators](https://github.com/django/django/blob/main/django/core/validators.py)

adhocracy4/images/validators.py
goapunk: added a comment

m4ra (approved):
Nice!

PR #1533, opened 2024-02-12
dashboard/mixins: change form_valid back to delete in

goapunk:
DashboardComponentDeleteSignalMixin and add documentation.

**Tasks**
- [x] Documentation (docs and inline)
- [x] Changelog

### DIFF
--- a/adhocracy4/dashboard/mixins.py
+++ b/adhocracy4/dashboard/mixins.py
@@ -1,3 +1,4 @@
+import warnings
 from copy import deepcopy
 
 from django.apps import apps
@@ -107,21 +108,32 @@ class DashboardComponentMixin(base.ContextMixin):
 
 
 class DashboardComponentFormSignalMixin(edit.FormMixin):
+    """Mixin which sends a project_component_updated or module_component_updated signal
+    when creating, updating or deleting an object via the HTTP POST method
+    (e.g. from the dashboard).
+    """
+
     def form_valid(self, form):
+        # The call to super.form_valid() may delete self.project or self.module,
+        # to be able to still reference them below we need to store them in separate
+        # variables before they are deleted.
+        project = self.project
+        module = self.module
+
         response = super().form_valid(form)
 
         component = self.component
         if component.identifier in components.projects:
             signals.project_component_updated.send(
                 sender=component.__class__,
-                project=self.project,
+                project=project,
                 component=component,
                 user=self.request.user,
             )
         else:
             signals.module_component_updated.send(
                 sender=component.__class__,
-                module=self.module,
+                module=module,
                 component=component,
                 user=self.request.user,
             )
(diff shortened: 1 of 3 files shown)


adhocracy4/dashboard/mixins.py
```diff
@@ -107,35 +107,52 @@ def get_context_data(self, **kwargs):
...
                 component=component,
                 user=self.request.user,
             )
         return response
 
 
 class DashboardComponentDeleteSignalMixin(edit.DeletionMixin):
-    def form_valid(self, request, *args, **kwargs):
+    """Mixin which sends a project_component_updated or module_component_updated signal
+    when deleting an object via the HTTP DELETE method.
```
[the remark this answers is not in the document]
m4ra: see https://www.django-rest-framework.org/api-guide/generic-views/#destroyapiview

adhocracy4/dashboard/mixins.py
```diff
@@ -107,35 +107,52 @@ def get_context_data(self, **kwargs):
...
 
 class DashboardComponentFormSignalMixin(edit.FormMixin):
+    """Mixin which sends a project_component_updated or module_component_updated signal
+    when creating, updating or deleting an object via the HTTP POST method
+    (e.g. from the dashboard).
+    """
+
     def form_valid(self, form):
+        # Project and module have to be stored before delete is called as
+        # they may rely on the still existing db object.
```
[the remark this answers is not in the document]
goapunk: was copied from the mixin below, it means that if we don't assign them to variables here, they will be deleted from the db in the super call and accessing `self.project` will give an error . I can try to reword it

adhocracy4/dashboard/mixins.py
```diff
@@ -107,35 +108,67 @@ def get_context_data(self, **kwargs):
...
 
 
 class DashboardComponentDeleteSignalMixin(edit.DeletionMixin):
-    def form_valid(self, request, *args, **kwargs):
+    """Deprecated, use DashboardComponentFormSignalMixin for POST requests instead.
+    This mixin will be removed in a future version.
+    """
+
+    def __init__(self):
+        warnings.warn(
```
goapunk, line 149: @m4ra I first deleted the Mixin but then thought maybe adding a warning is a good idea in case we do actually use it somewhere? Not sure if that's necessary though, what do you think?

m4ra: @goapunk I see it is used in mB for topic, maptopic and offlineevent delete views, with a template. We should replace this mixin with `DashboardComponentFormSignalMixin` in the mB views.
see e.g: https://github.com/liqd/a4-meinberlin/blob/main/meinberlin/apps/maptopicprio/views.py#L143

goapunk: @m4ra I did that in https://github.com/liqd/a4-meinberlin/pull/5454, however, my concern is that when removing `DashboardComponentDeleteSignalMixin` we only remove the signal, you can still delete objects because the `idea_views.AbstractIdeaDeleteView` uses `DeleteView` (which has `delete()`) so we wouldn't get any logging/warning if some js code still calls it.

changelog/_00010.md
```diff
@@ -0,0 +1,7 @@
+### Changed
+
+- **Breaking Change** Deprecated
+  dashboard.mixins.DashboardComponentDeleteSignalMixin,
+  for forms / POST requests use
+  dashboard.mixins.DashboardComponentFormSignalMixin instead.
+  DashboardComponentDeleteSignalMixin will be removed in the version.
```
m4ra, line 7: which version? next version you mean? next release mb? @goapunk

adhocracy4/dashboard/mixins.py
m4ra: @goapunk I see the `AbstractIdeaDeleteView` uses `form_valid()` not `delete()`:
https://github.com/liqd/a4-meinberlin/blob/b9aa9929fa5446032ce4ccb247003f5aeb1a7974/meinberlin/apps/ideas/views.py#L163

goapunk: yes, but it inherits `delete` from `generic.DeleteView` -> `generic.BaseDeleteview` which in turn inherits `delete` from `DeletionMixin`

m4ra: `DeletionMixin` will call delete if it gets DELETE http header, whereas the views that inherent from `AbstractIdeaDeleteView` use a template and send POST http.
https://docs.djangoproject.com/en/4.2/ref/class-based-views/mixins-editing/#deletionmixin

goapunk: yes, for POST / Forms it's fine, we still don't know about any javascript calls though (althoug it's very unlikely there are any). In any case, I'm also fine with removing the mixing, I just thought I add the deprecation as a precaution in case we missed some js code which uses the view.

PR #1589, opened 2024-04-03
fix migrations

m4ra:
**Tasks**
- [x] Changelog


m4ra:
closing this as I found out that django 4.0 generates migrations of related fields with Camel naming, while django 4.2 generates them with lower case names.

m4ra:
closing this as I found out that django 4.0 generates migrations of related fields with Camel naming, while django 4.2 generates them with lower case names.

PR #1590, opened 2024-04-09
Fix: remove reverse relation choices check in Question

m4ra:
**Tasks**
- [x] Changelog

### DIFF
--- a/adhocracy4/polls/models.py
+++ b/adhocracy4/polls/models.py
@@ -70,17 +70,6 @@ class Question(models.Model):
                         )
                     }
                 )
-            elif Choice.objects.filter(question=self.pk).count() > 0:
-                raise ValidationError(
-                    {
-                        "is_open": _(
-                            "Question with choices cannot become "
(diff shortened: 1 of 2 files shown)


adhocracy4/polls/models.py
```diff
@@ -70,17 +70,6 @@ def clean(self, *args, **kwargs):
...
-                    {
-                        "is_open": _(
-                            "Question with choices cannot become "
-                            "open question. Delete choices or add new "
-                            "open question."
-                        )
-                    }
-                )
-
         super().clean(*args, **kwargs)
```
m4ra, line 73: @goapunk this check here is causing errors to save a new open questions instances, as the choices can only be added after a question is created. For existing questions, a check if a question is open happens when creating the choices below in the Choice model

m4ra: hmm, actually in my installed version (earlier version for 4.2 upgrade) was referring to choice as self.choices. But the code in main, looks fine.

PR #1592, opened 2024-04-22
categories: make icon form field optional 

goapunk:
make icon form field optional  to allow categories without icon

fixes https://github.com/liqd/adhocracy-plus/issues/2580 and the same issue in mb

**Tasks**
- [x] Changelog


m4ra (approved):
LGTM

Issue #1588, opened 2024-03-21
Bug in comment deletion


TeaJayyy:
**URL:** e.g. https://kfd-dev.liqd.net/ideas/2024-00001/; same problem in a+ & mB
**user:** any
**expected behaviour:** when confirming that a/my comment should be deleted, the security question whether a comment should be deleted stays on the screen (see screenshot). you need to press "abort" to get back.
**behaviour:** the security question disappears after confirming 
**important screensize:** any
**device & browser:** any
**Comment/Question:** 

![Bildschirmfoto 2024-03-21 um 10 40 07](https://github.com/liqd/adhocracy4/assets/147062890/fb4611e9-ed2e-4b53-b7a7-0a9c911f0a4a)

PR #1596, opened 2024-04-29
static/Modal.jsx: Handle default value for dismissOnSubmit prop

hom3mad3:
second try at fixing https://github.com/liqd/adhocracy4/issues/1588 & https://github.com/liqd/adhocracy-plus/issues/2565

previous PR https://github.com/liqd/adhocracy4/pull/1594

### DIFF
--- a/adhocracy4/static/Modal.jsx
+++ b/adhocracy4/static/Modal.jsx
@@ -2,7 +2,7 @@ import React from 'react'
 import django from 'django'
 
 const Modal = (props) => {
-  const dismiss = props.dismissOnSubmit ? 'modal' : 'false'
+  const keepOpen = props.keepOpenOnSubmit || false
   const abortStr = props.abort || django.gettext('Abort')
   const bodyCssClass = props.partials.bodyClass || ''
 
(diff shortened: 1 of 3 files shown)


adhocracy4/static/Modal.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import django from 'django'
 
 const Modal = (props) => {
-  const dismiss = props.dismissOnSubmit ? 'modal' : 'false'
+  const dismiss = props.dismissOnSubmit !== undefined ? (props.dismissOnSubmit ? 'modal' : 'false') : 'modal'
```
goapunk, line 5: While this works (and I'm generally fine with merging it) this seems a bit complicated to read to me, I would suggest to either pass `dismissOnSubmit` on the comment delete modal (no change necessary here then) or we change the prop from `dismissOnSubmit` to something like `keepOpenOnSubmit`, so the default behavior is `modal` and we only need to set this prop on modals which need non-default behavior, what do you think?

goapunk (approved):
LGTM!

PR #1619, opened 2024-05-29
package.json: remove peerDependencies as they are currently to strict

goapunk:
The peerDependencies currently make updating quite complicated in a+/mb so it seems easier to remove them for now and maybe start fresh with less strict version requirements later.

**Tasks**
- [ ] Changelog


goapunk:
fixed the ci errors (the dependencies should be in the normal ones, not the devDependencies)

PR #1657, opened 2024-08-06
move comment categories from templatetag to api

goapunk:
**Tasks**
- [ ] Changelog


adhocracy4/comments_async/api.py
```diff
@@ -70,7 +71,7 @@ def list(self, request, content_type=None, object_pk=None):
                     return Response(status=status.HTTP_400_BAD_REQUEST)
             return response
         serializer = self.get_serializer(queryset, many=True)
-        return Response(serializer.data)
+        return Response({"results": serializer.data})
```
goapunk, line 74: This was bugged but is basically never reached anyway. I fixed it now and added a test for it

Issue #1043, opened 2022-04-12
comments_async: info message wrong


philli-m:
**expected behaviour**: members of private and semi public projects should see correct message
**behaviour**: PhilUser in Opin is listed as a member but sees incorrect message in rating phase 'Only invited users can actively participate.'
**Comment/Question**: When project is public the message reads 'The currently active phase doesn't allow to comment.'
In all projects see original issue https://github.com/liqd/a4-opin/issues/2525

goapunk:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/adhocracy4-6697bb5b650717053895c6ba/issues/liqd/adhocracy4/1043?planning-poker) with Zenhub @hom3mad3

TeaJayyy:
@hom3mad3 No I don't, close it

TeaJayyy:
@hom3mad3 You can reproduce the issue following these steps: 

1. create a private project
2. add idea challenge or spatial idea challenge (problem occurs in both modules)
3. create an idea
4. add a normal participant to the project
5. set the phase of the module to rating phase
6. go as a participant to the details of the idea
7. The comment section displays the message "Only invited users can actively participate." although it should state "rating phase does not allow comments".

![Bildschirmfoto 2024-08-21 um 13 35 00](https://github.com/user-attachments/assets/af78098a-a5e3-4e51-a071-97a76ba98234)

![Bildschirmfoto 2024-08-21 um 13 31 31](https://github.com/user-attachments/assets/8f429d3f-caa7-4a70-86e7-f7cc92db3bdc)

PR #1670, opened 2024-08-27
comments_async: display correct message for private group users

hom3mad3:
fixes : https://github.com/liqd/adhocracy4/issues/1043

steps to reproduce: https://github.com/liqd/adhocracy4/issues/1043#issuecomment-2301842120


m4ra (approved):
LGTM

PR #1679, opened 2024-10-21
polls/PollOpenQuestion.jsx: change input type to textarea

hom3mad3:
fix issue with the open question input field


goapunk (approved):
LGTM!

PR #1683, opened 2024-10-29
[8381] add option to allow unregistered users to vote in polls

goapunk:
**Tasks**
- [ ] Changelog


goapunk:
@m4ra @hom3mad3 here's a version where everything is integrated into the existing poll module for comparison. It's less big of a change than I expected. No changes are necessary on the a+ side so can be tested with main.

adhocracy4/polls/api.py
```diff
@@ -93,57 +100,108 @@ def retrieve(self, request, *args, **kwargs):
...
+                        user=self.request.user,
+                        organisation=self.get_object().project.organisation,
+                        defaults={
+                            "has_agreed": self.request.data["agreed_terms_of_use"]
+                        },
+                    )
             else:
                 raise ValidationError(
                     {_("Please agree to the organisation's terms of use.")}
                 )
```
m4ra, line 154: how the non-registered user will agree to the terms? This happens by default when they register IIRC

adhocracy4/polls/models.py
```diff
@@ -164,7 +171,7 @@ def __str__(self):
 
     class Meta:
         ordering = ["id"]
-        unique_together = ("question", "creator")
+        unique_together = ("question", "creator", "content_id")
```
m4ra, line 174: does this work if either creator or content_id is None?

m4ra (commented):
I left some comments for the django part, but need to do a local test too.
@hom3mad3 can you look more closely to the react part? 
BTW can we have the makefile and the github workflows in a separate PR?

adhocracy4/polls/models.py
goapunk: no, None is treated as a normal value in the unique check as far as I understand, therefore e.g. `question1 + user1 + None` is a valid unique value. Same for `question1 + None + unique uuid` is a valid unique value. And both trigger an error if voted again with the same question and user/content_id

goapunk:
> I left some comments for the django part, but need to do a local test too. @hom3mad3 can you look more closely to the react part? BTW can we have the makefile and the github workflows in a separate PR?

I'll make extra PRs but we will have to merge the workflow before this then and rebase it

adhocracy4/polls/serializers.py
```diff
@@ -130,24 +137,24 @@ class PollSerializer(serializers.ModelSerializer):
...
                     ).count()
-                    + models.Answer.objects.filter(
-                        question__poll=poll, creator=user
-                    ).count()
+                    + Answer.objects.filter(question__poll=poll, creator=user).count()
                 ) > 0
         return False
 
     def update(self, instance, data):
+        instance.allow_unregistered_users = data.get("allow_unregistered_users", False)
```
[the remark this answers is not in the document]
goapunk: You mean to check whether the unregistered user trying to vote is allowed to do that ? That's done by the rules/predicates already.

m4ra:
@goapunk tried to test it locally, but we cannot set the `allowed_unregistered_users` from the poll's dashboard yet. Maybe could be a checkbox in the basic settings? For registered users works as expected.

goapunk:
> @goapunk tried to test it locally, but we cannot set the `allowed_unregistered_users` from the poll's dashboard yet. Maybe could be a checkbox in the basic settings? For registered users works as expected.

@m4ra it works for me, the checkbox is there (in the `Poll` setting). Did you do `npm link <path to adhocracy4>` ?

m4ra (approved):
LGTM, thank you!

PR #1687, opened 2024-11-04
[#8443] polls: refactor PollQuestion.jsx and add TextareaWithCounter.jsx

hom3mad3:
### Changes

expandabletextarea

1. **Refactored Poll Textarea Component**  
   - The `Poll` textarea is now handled in `TextareaWithCounter.jsx`.  
   - Initially intended to create a global reusable textarea with a counter, but for now, this implementation is specific to `PollDetail`. Future cases will be addressed if this approach is adopted globally.  
   - The implementation was inspired by the [GOV.UK Design System's character count component](https://design-system.service.gov.uk/components/character-count/)

2. **Breaking Change: Textarea Visibility Logic**  
   - Updated logic to affect the entire visibility of the textarea container.  
   - The container is now displayed only when the "other" option is selected. Improves UX by following progressive disclosure pattern and performance/load by avoiding loading hidden/unnecessary elements.
   - Added dynamic handling for textarea height and optimized for persistent data when the option is untoggled by mistake with `React.memo`.

3. **Accessibility and Semantic Enhancements**  
   - Improved accessibility and semantic structure.  
   - Added label to textarea
   - Added aria support
   - Separated logic into smaller, more maintainable chunks for better readability.

4. **Component Renaming and Refactoring**  
   - `PollQuestion.jsx` is now renamed to `PollChoice.jsx` for consistency with `PollDashboard`.  
   - The component was broken down into smaller sub-components to enhance readability and maintainability.

5. **Test Updates**  
   - Updated test data dump and adjusted tests to align with the new changes.

---

### Improvement Ideas

- **`CharCounter.jsx`:** Consider better handling for screen readers to improve accessibility and provide clearer feedback for users with assistive technologies.
- Questions should be optimized for accessibility as well or at least improvement of semantics.


adhocracy4/polls/static/PollDetail/PollQuestions.jsx
```diff
@@ -383,7 +383,7 @@ class PollQuestions extends React.Component {
         <>
           {this.state.showResults
             ? (
-              <div className="pollquestionlist-container">
+              <div className="poll__preliminary-results">
```
goapunk, line 386: just to check: there are no styles for this in a+, that's intentional? and by removing the old class we lose a a padding-top, that's also intentional?

goapunk (changes requested):
LGTM, thanks! I just left some small notes regarding the mock and the textarea

adhocracy4/polls/static/PollDetail/PollQuestions.jsx
goapunk: Thanks, I can see it in the mb prs but in the a+ one the new class hasn't been added anywhere: https://github.com/liqd/adhocracy-plus/pull/2848/files

goapunk (approved):
LGTM, thanks!

PR #1688, opened 2024-11-06
polls: add new signal which is called after a user voted on a poll

goapunk:
**Tasks**
- [x] Changelog

### DIFF
--- a/adhocracy4/polls/api.py
+++ b/adhocracy4/polls/api.py
@@ -30,6 +30,7 @@ from .models import Poll
 from .models import Question
 from .models import Vote
 from .serializers import PollSerializer
+from .signals import poll_voted
 
 
 class PollViewSet(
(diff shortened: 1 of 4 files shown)


adhocracy4/polls/signals.py
```diff
@@ -0,0 +1,3 @@
+from django.dispatch import Signal
+
+poll_voted = Signal()
```
m4ra, line 3: Can you add here the arguments that are needed for the signal.

adhocracy4/polls/api.py
```diff
@@ -178,6 +179,8 @@ def vote(self, request, pk):
             self.save_vote(question, vote_data, creator, content_id)
 
         poll = self.get_object()
+        print("##########################################")
+        poll_voted.send(sender=None, poll=poll, creator=creator, content_id=content_id)
```
m4ra, line 183: sender is the poll view, I have added it as the `self.__class__` in my PR for mB  kiezradar.

m4ra: and the print statement can go I believe :)

m4ra (changes requested):
a few things

adhocracy4/polls/signals.py
goapunk: added a docstring

adhocracy4/polls/api.py
goapunk: thanks, added

m4ra (commented):
an extra thing for the tests

m4ra (approved):
Nice, thanks!

PR #1689, opened 2024-11-07
remove deprecated ckeditor fields

goapunk:
Bit of an old PR but ideally we get rid of the old ckeditor4 package. I replaced all the ckeditor4 fields in the migrations so we can remove the package, the ckeditor4 field extends CharField so it should be fine

**Tasks**
- [x] Changelog


m4ra (approved):
lgtm

PR #1691, opened 2024-11-11
poll: hide back to poll button for unregistered users

goapunk:
**Tasks**
- [ ] Changelog


adhocracy4/polls/static/PollDetail/PollQuestions.jsx
```diff
@@ -388,9 +388,10 @@ class PollQuestions extends React.Component {
                   <PollResults key={idx} question={question} />
                 ))}
                 <Alert onClick={() => this.removeAlert()} {...this.state.alert} />
-                <div className="poll">
-                  {this.state.hasUserVote ? this.linkChangeVote : this.linkToPoll}
-                </div>
+                {this.state.questions.length > 0 && this.state.questions[0].authenticate &&
```
hom3mad3, line 391: we could put `this.state.questions[0].authenticate` into a variable with a meaningful name (i.e. unregisteredUser)? 
also a note for eventual refactoring: `{this.state.hasUserVote ? this.linkChangeVote : this.linkToPoll}` doesn't immediate show what type of element we are rendering, we should probably have the <button> have an onclick event which handles things for us.

goapunk: yeah all the conditions in here could need a refactor, the chains are terribly long and hard to read. (Some of them could ideally be already done in the backend as well I think), I just made it work with minimal effort for now

PR #1728, opened 2024-12-11
react/ratings: hide accessible text for rating buttons to only be visible for screen readers

vellip:
this fixes an error with the new rating buttons that causes the screen reader text to be visible


vellip:
see https://github.com/liqd/a4-meinberlin/issues/5876

PR #1741, opened 2024-12-30
[#8527] E-Mail attachment setting

vellip:
Add new setting `A4_EMAIL_ATTACHMENTS` for when you need more than the `email_logo.png` attachment in your emails.

No breaking change as a fallback is included when the setting is not existing.

See https://github.com/liqd/a4-meinberlin/pull/5908 for usage

**Tasks**
- [x] Changelog


goapunk (approved):
nice, thanks!

PR #1758, opened 2025-01-22
static/Modal.jsx refactor: migrate from Bootstrap to native dialog element adapt to mB redesign

hom3mad3:
**_please dont merge this once approved, waiting to touch a-plus first_**

Replace Bootstrap modals with native HTML <dialog> element  and update styles for the mB redesign.

This change simplifies our modal implementation while improving the mobile experience. The <dialog> element provides better accessibility and native support for modal behaviors.

https://github.com/user-attachments/assets/3fb8d9a0-9e58-4510-b780-dc92d0c98ef7

https://github.com/user-attachments/assets/0215e666-2eab-40f6-9323-e0f9f697e734

TODO:
- [ ] a-plus dependency

### DIFF
--- a/adhocracy4/comments/static/comments/Comment.jsx
+++ b/adhocracy4/comments/static/comments/Comment.jsx
@@ -1,12 +1,12 @@
+const React = require('react')
+const django = require('django')
+
+const { default: Modal } = require('../../../static/Modal')
 const ReportModal = require('../../../reports/static/reports/react_reports').ReportModal
 const RatingBox = require('../../../ratings/static/ratings/RatingBox')
-const Modal = require('../../../static/Modal')
 const CommentEditForm = require('./CommentEditForm')
 const CommentForm = require('./CommentForm')
 const CommentManageDropdown = require('./CommentManageDropdown')
-
-const React = require('react')
-const django = require('django')
 
 const safeHtml = function (text) {
   return { __html: text }
@@ -156,7 +156,6 @@ class Comment extends React.Component {
     if (this.isOwner() || this.context.isModerator) {
       return (
         <Modal
-          name={'comment_delete_' + this.props.id}
           partials={{
             title: deleteTag,
             description: confirmDeleteText
--- a/adhocracy4/comments_async/static/modals/UrlModal.jsx
+++ b/adhocracy4/comments_async/static/modals/UrlModal.jsx
@@ -1,49 +1,63 @@
-import React, { useState } from 'react'
+import React, { useState, useRef } from 'react'
 import django from 'django'
 
 import Modal from '../../../static/Modal'
 
+const translated = {
+  share: django.gettext('Share'),
+  buttonTextCopy: django.gettext('Copy'),
+  buttonTextCopied: django.gettext('Copied')
+}
+
 export const UrlModal = (props) => {
   const [clicked, setClicked] = useState(false)
-  const buttonTextCopy = django.gettext('Copy')
-  const buttonTextCopied = django.gettext('Copied')
-  const partials = {}
-  partials.hideHeader = true
-  partials.hideFooter = true
-  partials.title = props.title
-  partials.body = (
-    <div className="input-group">
-      <input
-        id={'share-url-' + props.objectId}
-        type="text"
-        className="form-control"
-        value={props.url}
-        readOnly
-      />
-      <button
-        className="btn btn--primary input-group-append"
-        data-bs-toggle="button"
-        aria-pressed="false"
-        autoComplete="off"
-        onClick={(e) => copyUrl(e)}
-      >
-        {clicked ? buttonTextCopied : buttonTextCopy}
-      </button>
-    </div>
-  )
+  const inputRef = useRef(null)
 
-  const copyUrl = () => {
-    const copyText = document.getElementById('share-url-'.concat(props.objectId))
-    copyText.select()
-    document.execCommand('copy')
-    setClicked(true)
+  const copyUrl = (e) => {
+    e.preventDefault()
+    if (inputRef.current) {
+      navigator.clipboard.writeText(inputRef.current.value)
+        .then(() => {
+          setClicked(true)
+          return true
+        })
+        .catch((err) => {
+          console.error('Failed to copy text: ', err)
+        })
+    }
+  }
+
+  const partials = {
+    hideHeader: true,
+    hideFooter: true,
+    title: props.title,
+    body: (
+      <div className="input-group a4-url-modal__body">
+        <input
+          id={'share-url-' + props.objectId}
+          ref={inputRef}
+          type="text"
+          className="form-control"
+          value={props.url}
+          readOnly
+        />
+        <button
+          className="a4-url-modal__button"
+          aria-pressed={clicked}
+          autoComplete="off"
+          onClick={(e) => copyUrl(e)}
+        >
+          {clicked ? translated.buttonTextCopied : translated.buttonTextCopy}
+        </button>
+      </div>
+    )
   }
 
   return (
     <Modal
       abort={props.abort}
-      name={props.name}
       partials={partials}
+      toggle={<><i className="fas fa-share" aria-hidden="true" /> {translated.share}</>}
       keepOpenOnSubmit
     />
   )
--- a/adhocracy4/reports/static/reports/ReportModal.jsx
+++ b/adhocracy4/reports/static/reports/ReportModal.jsx
@@ -1,27 +1,41 @@
 import React, { useState } from 'react'
 import django from 'django'
-import Modal from '../../../static/Modal'
 import api from '../../../static/api'
+import Modal from '../../../static/Modal'
 
 const translations = {
+  reportTitle: django.gettext('Report Content'),
   reportSent: django.gettext('Report sent'),
   thankyouText: django.gettext('Thank you! We are taking care of it.'),
   placeholderText: django.gettext('Your message'),
-  sendReport: django.gettext('Send Report')
+  sendReport: django.gettext('Send Report'),
+  report: django.gettext('Report'),
+  validationError: django.gettext('Please enter a message before submitting.')
 }
 
 export const ReportModal = (props) => {
   const [report, setReport] = useState('')
   const [showSuccessMessage, setShowSuccessMessage] = useState(false)
   const [showReportForm, setShowReportForm] = useState(true)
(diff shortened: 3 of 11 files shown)


hom3mad3:
@CarolingerSeilchenspringer @mcastro-lqd would it be ok to leave the mobile design like suggested above? the full width and radius makes it look more native.

CarolingerSeilchenspringer:
@mcastro-lqd can you have a look?

adhocracy4/comments_async/static/modals/UrlModal.jsx
```diff
@@ -21,7 +40,7 @@ export const UrlModal = (props) => {
         readOnly
       />
       <button
-        className="btn btn--primary input-group-append"
+        className="btn btn--primary input-group-append button button--fullwidth-palm"
```
vellip, line 43: Is that not a meinberlin class?

adhocracy4/comments_async/static/modals/UrlModal.jsx
```diff
@@ -3,16 +3,35 @@ import django from 'django'
...
+  share: django.gettext('Share')
+}
+
 export const UrlModal = (props) => {
   const [clicked, setClicked] = useState(false)
   const buttonTextCopy = django.gettext('Copy')
   const buttonTextCopied = django.gettext('Copied')
+
+  const copyUrl = () => {
+    const copyText = document.getElementById('share-url-'.concat(props.objectId))
```
vellip, line 16: Why not use a ref?

vellip (changes requested):
Very good work, looks nice. It was interesting to see the `dialog` element being used. Haven't done that before. I left a some questions for you, but it looks good already.

adhocracy4/comments_async/static/modals/UrlModal.jsx
```diff
@@ -3,16 +3,35 @@ import django from 'django'
...
+    }
+  }
+
   const partials = {}
   partials.hideHeader = true
   partials.hideFooter = true
   partials.title = props.title
   partials.body = (
-    <div className="input-group">
+    <div className="input-group a4-modal__url-modal">
```
vellip, line 34: Maybe call it `url-body` instead? Technically the modal is the very top element, right? Or give the modal a `className` prop and pass it there?

adhocracy4/comments_async/static/modals/UrlModal.jsx
hom3mad3: good catch!

adhocracy4/comments_async/static/modals/UrlModal.jsx
hom3mad3: not sure how to go about it tbh, dont wanna extend classes, do you think that smells too bad?

adhocracy4/reports/static/reports/ReportModal.jsx
```diff
@@ -1,27 +1,41 @@
...
+import Modal from '../../../static/Modal'
 
 const translations = {
+  reportTitle: django.gettext('Report Content'),
   reportSent: django.gettext('Report sent'),
   thankyouText: django.gettext('Thank you! We are taking care of it.'),
   placeholderText: django.gettext('Your message'),
-  sendReport: django.gettext('Send Report')
+  sendReport: django.gettext('Send Report'),
+  report: django.gettext(' Report'),
```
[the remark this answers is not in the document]
hom3mad3: copied from old code

changelog/_0003.md
```diff
@@ -0,0 +1,13 @@
...
+- Migrated from Bootstrap classes to custom CSS classes
+- Modified `react_reports.py` to adapt data passing pattern for new modal structure
+
+### Added
+- Mobile-specific modal styling with bottom sheet animation
+- New CSS classes following `a4-modal` prefix convention
+- Input validation and error handling for ReportModal.jsx
+
+### Removed
+- Removed Bootstrap modal dependencies
```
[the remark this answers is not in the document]
hom3mad3: 🦾 one step at a time and then in a year react is the new bootstrap anyways 🤣

adhocracy4/static/Modal.jsx
```diff
@@ -1,69 +1,116 @@
...
+    if (!keepOpenOnSubmit) {
+      handleClose()
+    }
+  }
+
+  const dialogModal = (
+    <dialog
+      ref={dialogRef}
+      id={id}
+      aria-label={partials.title}
```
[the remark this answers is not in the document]
hom3mad3: good point 👍🏼

adhocracy4/comments_async/static/modals/UrlModal.jsx
vellip: I am not sure. I think it's strange to add two button classes. How are we doing it elsewhere? These can't be the only buttons in a4 that we also have in mb right? I guess I thought we are already extending them.

adhocracy4/static/Modal.jsx
```diff
@@ -1,69 +1,116 @@
...
+    </dialog>
+  )
+
+  return (
+    <>
+      <button
+        style={{
+          whiteSpace: 'normal',
+          textAlign: 'center'
+        }}
```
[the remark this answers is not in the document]
hom3mad3: remember those annoying icon breaks in the comments? since the button is nowrap by default, it won't work. So, I decided to just inline those styles here. 

Do you think it’s okay, or makes the component itself it too opinionated? This way, I don’t need to declare any additional styles in aPlus or mB since it would follow the component (in an ideal world where we don't have different designs this would be the case). 

However, now that I think about it, aPlus might be different. I’ll look into the styles there today.

adhocracy4/static/Modal.jsx
```diff
@@ -1,69 +1,116 @@
...
-                data-bs-dismiss="modal"
-                onClick={props.handleClose}
-              >
-                {abortStr}
-              </button>
-            </div>}
-        </div>
+            <button
+              className="a4-modal__submit button"
+              onClick={onSubmit}
```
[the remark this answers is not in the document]
hom3mad3: agreed! i'm using all the old naming, so that makes for a proper refactor 🙏🏼

adhocracy4/static/Modal.jsx
```diff
@@ -1,69 +1,116 @@
...
+  }, [])
+
+  const handleOpen = (e) => {
+    e.preventDefault()
+    dialogRef.current?.showModal()
+  }
+
+  const handleClose = () => {
+    dialogRef.current?.close()
+  }
```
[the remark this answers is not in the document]
hom3mad3: could you share an example of how you would do it?

adhocracy4/static/Modal.jsx
```diff
@@ -1,69 +1,116 @@
...
 
-  return (
-    <div
-      className="modal fade"
-      id={props.name}
-      tabIndex="-1"
-      role="dialog"
-      aria-label={props.partials.title || 'Modal'}
-      aria-hidden="true"
+const Modal = ({ id, partials, handleSubmit, action, keepOpenOnSubmit, toggle }) => {
```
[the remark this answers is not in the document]
hom3mad3: yes, much better

hom3mad3:
> Very good work, looks nice. It was interesting to see the `dialog` element being used. Haven't done that before. I left a some questions for you, but it looks good already.

dialog <3! it simplifies life a lot since its basically two methods open and close 

the biggest issue with this story though, was that at some point i tested the finger swiper for mobile and that was basically adding half the lines of code, so i decided to keep simple. it's a great little native component, but if you need fancy stuff then you gotta write a lot more stuff. i'm super happy not to have to write custom keyboard and accessibility extensions anymore thought.

hom3mad3:
@CarolingerSeilchenspringer @mcastro-lqd  @vellip oh, i'll also see if i can already change the copy button styles to use the new "toggle" button. if not, i'll create an issue for after release.

adhocracy4/comments_async/static/modals/UrlModal.jsx
goapunk: We do us extend elsewhere but we ran into problems with that early in the redesign process. For some reason there was a bug/weird limit: after reaching a certain amount @extend the css file would suddenly jump to 50mb. Removing one (any) extend would bring it back down to normal size, so we tried to avoid too much extending since then. Maybe that bug was fixed now though

hom3mad3: @vellip `button button--fullwidth-palm` is used in a couple of places in mB. i don't have an issue with this, since this is the 'correct' BEM approach `class class--modifier`, usually together. however, i do see a problem/agree with a4 knowing about mB classes (not sure this is what you meant?) 

also forgot to mention: the remaining btn and btn--primary will probably be removed once i touch a-plus and then i will replace with whatever a4-modal__ class is needed. however that won't solve the meinBerlin classes problem to go around the extend.
 
@goapunk will try the extend thing again

adhocracy4/static/Modal.jsx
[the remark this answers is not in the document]
hom3mad3: never mind, I wasn’t at my sharpest this morning

adhocracy4/static/Modal.jsx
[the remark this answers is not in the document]
hom3mad3: nevermind, i fixed with a class :P i think it was late in the day when i thought of this

adhocracy4/comments/static/comments/Comment.jsx
```diff
@@ -156,7 +156,7 @@ class Comment extends React.Component {
     if (this.isOwner() || this.context.isModerator) {
       return (
         <Modal
-          name={'comment_delete_' + this.props.id}
+          useId={'comment_delete_' + this.props.id}
```
vellip, line 159: I think we misunderstood each other :D I was suggesting that instead of having to pass a prop to every modal, to generate unique IDs you could just use https://react.dev/reference/react/useId inside `Modal` and generate your own ids.

vellip (approved):
LGTM, just the useId thing

adhocracy4/comments/static/comments/Comment.jsx
hom3mad3: aha! 😬 that makes a lot of sense! i thought `id` was a protected word like key or index, can't remember which one anymore hehe
fixing right away! 🏃🏻‍♂️‍➡️

PR #1761, opened 2025-01-23
[#8639] address-search: move components into a4 from mein_berlin PT 1

vellip:
This is part 1 of 2. Please feel already free to review and merge this. This will be merged into another branch that stems from `main` and so will part 2. Doing it like this allows me to split this story up for easier review into...

1. Moving components around (from mb -> a4)
2. Making further code changes in these components to fulfil the user story.

Once both parts are reviewed & merged into this branch `pv-2025-01-address-search` I can then merge this into `dev` (mb)/`main` (a4)

**Describe your changes**
Moves components out of mein berlin and into a4, adapts code to be more generic. It moves over:

1. The useCombobox hook, that is used in a) AutoComplete (AddressSearch) and b) Multi-Select
2. the useDebounce hook which is used to create debounced callbacks that you can put on e.g. a onChange handler and then it waits x milliseconds before triggering the actual callback and if another change event happens it resets the timer.
4. The ProjectsMapSearch component (now called AddressSearch) – the main part of this story
5. The AutoComplete component – which the AddressSearch is built on top of.

meinberlin PR: https://github.com/liqd/a4-meinberlin/pull/5969

**Tasks**
- [ ] Changelog

### DIFF
--- a/adhocracy4/maps_react/static/a4maps_react/AddressSearch.jsx
+++ b/adhocracy4/maps_react/static/a4maps_react/AddressSearch.jsx
@@ -1,0 +1,104 @@
+import React, { useEffect, useState } from 'react'
+import django from 'django'
+import { useMap } from 'react-leaflet'
+import L from 'leaflet'
+
+import useDebounce from '../../../static/useDebounce'
+import { AutoComplete } from '../../../static/forms/AutoComplete'
+import { makeIcon } from './GeoJsonMarker'
+
+const addressSearchCapStr = django.gettext('Address Search')
+let activeMarker = null
+
+function fetchSuggestions (address) {
+  return fetch(apiUrl + '?address=' + address)
+    .then((response) => response.json())
+}
+
+function getSearchResultText (feature) {
+  return feature.properties.strname + ' ' + feature.properties.hsnr + ' in ' + feature.properties.plz + ' ' + feature.properties.bezirk_name
+}
+
+const apiUrl = 'https://bplan-prod.liqd.net/api/addresses/'
+
+const AddressSearch = ({
+  onSubmitHandler
+}) => {
+  const [geoJson, setGeoJson] = useState(null)
+  const [searchString, setSearchString] = useState('')
+  const map = useMap()
+
+  const debouncedOnChange = useDebounce(async () => {
+    const geoJson = await fetchSuggestions(searchString)
+    setGeoJson(geoJson)
+  })
+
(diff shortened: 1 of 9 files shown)


adhocracy4/maps_react/static/a4maps_react/AddressSearch.jsx
```diff
@@ -0,0 +1,104 @@
...
+function fetchSuggestions (address) {
+  return fetch(apiUrl + '?address=' + address)
+    .then((response) => response.json())
+}
+
+function getSearchResultText (feature) {
+  return feature.properties.strname + ' ' + feature.properties.hsnr + ' in ' + feature.properties.plz + ' ' + feature.properties.bezirk_name
+}
+
+const apiUrl = 'https://bplan-prod.liqd.net/api/addresses/'
```
hom3mad3, line 22: @goapunk any opinion on this url here? could this cause an issue further down the line?

adhocracy4/static/forms/useCombobox.js
```diff
@@ -0,0 +1,179 @@
...
+function toggleValue (value, values) {
+  const shouldRemove = values.includes(value)
+  let newValues = [...values, value]
+  if (shouldRemove) {
+    newValues = values.filter(item => item !== value)
+  }
+  return newValues
+}
+
+const useCombobox = ({
```
hom3mad3, line 22: i'm uncertain about the name choice, since it's quite unclear on first encounter what a combobox is/does.multiselect combobox and has since been expanded to handle other elements? i think it would be helpful to add a comment explaining what this component is and how it interacts with other elements, or perhaps reconsider the name to provide better clarity.

adhocracy4/maps_react/static/a4maps_react/AddressSearch.jsx
goapunk: Hm somewhat tricky as this is not really a generic address search, it will only work with the mentioned url. However, as we will probably have to move away from the bplan map to fetch addresses in the long run, we could already lay out the foundation for it and load the url from a django setting + templatetag if that doesn't create a crazy long prop chain

hom3mad3: ok, thanks for the info 👍🏼

vellip: I will change this in pt 2

PR #1768, opened 2025-02-03
projects: add contact details flag to ProjectContactDetailMixin

hom3mad3:
added because of  https://github.com/liqd/a4-meinberlin/pull/6030

### DIFF
--- a/adhocracy4/projects/models.py
+++ b/adhocracy4/projects/models.py
@@ -82,6 +82,17 @@ class ProjectContactDetailMixin(models.Model):
 
     contact_url = models.URLField(blank=True, verbose_name=_("Website"))
 
+    def has_contact_details(self):
+        return any(
+            [
+                self.contact_name,
+                self.contact_address_text,
+                self.contact_email,
+                self.contact_phone,
+                self.contact_url,
+            ]
+        )
+
 
 class ProjectLocationMixin(models.Model):
     class Meta:


hom3mad3:
@m4ra @partizipation I needed to clean up the templates for the contact information refactor https://github.com/liqd/a4-meinberlin/pull/6030 and ended up making those changes. i would love an opinion on whether this is something we should do at all. and happy to get suggestions on best practices as i learn more about django backend specifics 🧑‍🎓 

coverages decrease a little, should i add tests?

adhocracy4/projects/models.py
```diff
@@ -82,6 +82,22 @@ class Meta:
...
+            [
+                self.contact_name,
+                self.contact_address_text,
+                self.contact_email,
+                self.contact_phone,
+                self.contact_url,
+            ]
+        )
+
+    def has_any_contact_info(self):
```
vellip, line 96: more of a wording question than anything else: I think `info` is not explicit enough to differentiate between this and `has_contact_details`?

you could also do the same as you did with `Project` here with `Organisation` and add a `has_contact_details` to it and in there return `any([self.organisation.address, self.organisation.url])`

hom3mad3: nice idea, thanks!!

m4ra: also we shouldn't check for self.organisation here as it is an attribute of Project model and here we are in the ProjectContactDetailMixin model

m4ra: @hom3mad3 @vellip looking at the Organisation, and how Micaela's new contact_organisation.html uses if checks for addresses and url, it doesn't make much sense to add another property method in the Organisation model to check for organisation\s contact details. Just an `if object.has_contact_details or if object.organisation` should be enough. I left comments in the other [PR](https://github.com/liqd/a4-meinberlin/pull/6030) too.

hom3mad3: @m4ra really appreciate it 💜

PR #1770, opened 2025-02-11
issue fix - modals/UrlModal: reset clicked state on close

hom3mad3:
fixes https://github.com/liqd/a4-meinberlin/issues/6062

### DIFF
--- a/adhocracy4/comments_async/static/modals/UrlModal.jsx
+++ b/adhocracy4/comments_async/static/modals/UrlModal.jsx
@@ -10,7 +10,7 @@ const translated = {
 }
 
 export const UrlModal = (props) => {
-  const [clicked, setClicked] = useState(false)
+  const [copied, setCopied] = useState(false)
   const inputRef = useRef(null)
 
   const copyUrl = (e) => {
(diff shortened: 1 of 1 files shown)


sevfurneaux (approved):

adhocracy4/comments_async/static/modals/UrlModal.jsx
```diff
@@ -27,6 +27,10 @@ export const UrlModal = (props) => {
     }
   }
 
+  const handleModalClose = () => {
+    setClicked(false) // Reset the clicked state when modal closes
```
sevfurneaux, line 31: Maybe we don't need a comment here if it was named `onClose`, then would be more explicit what it is doing.

hom3mad3: @sevfurneaux yes, the clicked state refers to the copy URL button. I could change it to setCopied—do you reckon that makes it clearer?

sevfurneaux: Yeah, `setCopied` makes sense!

PR #1774, opened 2025-02-19
projects: fix point migration by splitting it into separate migrations

goapunk:
**Tasks**
- [ ] Changelog


goapunk:
@m4ra sorry another fix, I forgot that data and schema migrations can't be in one migration, I think you even mentioned that

PR #1775, opened 2025-02-19
migrate geojson point to GeoDjango point

goapunk:
second attempt for https://github.com/liqd/adhocracy4/pull/1765, fixed some issues and tests


goapunk:
@m4ra this will also need a change in a+ to use the new serializer mixin here: https://github.com/liqd/adhocracy-plus/blob/main/apps/projects/serializers.py#L18

goapunk:
> So that we can store the point as django gis you mean?
> 
> > @m4ra this will also need a change in a+ to use the new serializer mixin here: https://github.com/liqd/adhocracy-plus/blob/main/apps/projects/serializers.py#L18

I think we don't allow creating/editing projects via the api, but we need the new mixin to serialize the gis point as geojson

adhocracy4/projects/migrations/0049_project_geos_point.py
```diff
@@ -0,0 +1,51 @@
...
+
+    dependencies = [
+        ("a4projects", "0048_project_geos_point_project_house_number_and_more"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            migrate_project_point_field, reverse_code=migrations.RunPython.noop
+        ),
+    ]
```
m4ra, line 51: can you rename this migration to reflect the action? e.g `0049_migrate_data_from_point_to_geos_point.py`

adhocracy4/projects/migrations/0050_replace_point_field.py
```diff
@@ -0,0 +1,30 @@
...
+            name="point",
+            field=django.contrib.gis.db.models.fields.PointField(
+                blank=True,
+                help_text="Locate your project. Click inside the marked area or type in an address to set the marker. A set marker can be dragged when pressed.",
+                null=True,
+                srid=4326,
+                verbose_name="Can your project be located on the map?",
+            ),
+        ),
+    ]
```
m4ra, line 30: please separate these in two migrations, one for removing the field, the other for adding it, as it can give inconsistent schema errors again.

adhocracy4/projects/migrations/0051_migrate_point_field.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.2.17 on 2025-02-19 15:49
+
+from django.db import migrations
+
+
+def migrate_project_geos_point_field(apps, schema_editor):
+    project = apps.get_model("a4projects", "Project")
+    for project in project.objects.all():
+        project.point = project.geos_point
```
m4ra, line 9: Shouldn't it also has a check if `project.geos_point` exists?

tests/conftest.py
```diff
@@ -26,6 +27,23 @@ def image_factory():
     return img_factories.ImageFactory()
 
 
+@pytest.fixture
+def gis_point():
+    return Point(13.397788148643649, 52.52958586909979)
```
m4ra, line 32: nit-picking, but can you stick to same naming either `gis_point` or `geos_point` as you have been naming it in the rest of the code?

tests/maps/test_map_validators.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # point within the berlin polygon
+    geojson_point = {
+        "type": "Feature",
+        "geometry": {
+            "type": "Point",
+            "coordinates": [13.411924777644563, 52.499598134440944],
+        },
+    }
+    point = GEOSGeometry(json.dumps(geojson_point["geometry"]))
```
m4ra, line 26: same, this could be named either `gis_point` or `geos_point`, just stay consistent.

tests/maps/test_map_validators.py
```diff
@@ -0,0 +1,44 @@
...
+        json.dumps(settings.BERLIN_POLYGON["features"][0]["geometry"])
+    )
+    validator = PointInPolygonValidator(polygon=polygon)
+
+    # point outside the berlin polygon
+    geojson_point = {
+        "type": "Feature",
+        "geometry": {"type": "Point", "coordinates": [13.459894, 51.574425]},
+    }
+    point = GEOSGeometry(json.dumps(geojson_point["geometry"]))
```
m4ra, line 42: same

m4ra (changes requested):
A few edits for extra precaution. And some naming consistency.

adhocracy4/projects/migrations/0051_migrate_point_field.py
goapunk: added

docs/location.md
```diff
@@ -0,0 +1,41 @@
...
+
+## GeoDjango Enhancements
+
+### Spatial Queries
+
+With `PointField`, we can now efficiently perform spatial queries. For example, retrieving projects within a 10km radius of a given location:
+
+```python
+from django.contrib.gis.geos import Point
+from django.contrib.gis.db.models.functions import Distance
```
m4ra, line 22: Not sure why the Distance is imported here in the example

m4ra (approved):
Just a comment on the documentation, but I will go ahead and merge now. We can edit the docs later again.

PR #1778, opened 2025-02-27
refactor PointFormMixin and make MapChoosePointWidget handle geo points

goapunk:
**Tasks**
- [ ] Changelog

### DIFF
--- a/adhocracy4/maps/mixins.py
+++ b/adhocracy4/maps/mixins.py
@@ -4,100 +4,188 @@ from collections import OrderedDict
 
 class GeoJsonPointMixin:
     """
-    A mixin that processes GeoJSON point data for compatibility with GeoDjango models.
+    Mixin to process GeoJSON point data for compatibility with GeoDjango models.
 
-    Classes using this mixin must define a `Meta` class with a `geo_field` attribute,
-    specifying the name of the model field that stores the point data.
-
-    Additionally, if `get_properties` returns a non-empty dictionary, the corresponding
-    GeoJSON properties are extracted and mapped to the specified model fields.
+    If `get_geojson_properties()` returns a non-empty dict, the corresponding GeoJSON
+    properties will be extracted and mapped to the specified model fields.
     """
 
-    def get_properties(self):
+    def get_geojson_properties(self):
         """
-        Defines a mapping of GeoJSON properties to model fields.
+        Return a mapping of GeoJSON property names to model field names.
 
         Returns:
-            dict: A dictionary where keys are GeoJSON property names and values are
-                  the corresponding model field names. If a model field name is identical
-                  to the GeoJSON property name, its value can be set to None.
+            dict: A dictionary where keys are GeoJSON property names and values are the
+                  corresponding model field names. Use None if the names are identical.
+        Example:
+            {"strname": "street_name", "hsnr": "house_number", "plz": "zip_code"}
         """
         return {}
 
-    def unpack_geojson(self, data):
+    def map_field_name(self, geojson_property, field_mapping, point_data=None):
         """
-        Extracts and reformats GeoJSON point data for use in a GeoDjango model.
+        Return the relevant model field name for a given GeoJSON property.
+
+        If point_data is provided and contains a "properties" dictionary, the mapping
+        is only applied if the geojson_property exists in those properties; otherwise, None is returned.
 
         Args:
-            data (dict): A dictionary containing GeoJSON data, including a point field
-                         specified by `Meta.geo_field`.
+            geojson_property (str): The GeoJSON property name.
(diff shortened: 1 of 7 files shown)


adhocracy4/maps/widgets.py
```diff
@@ -71,14 +75,23 @@ def render(self, name, value, attrs, renderer=None):
         if hasattr(settings, "A4_OPENMAPTILES_TOKEN"):
             omt_token = settings.A4_OPENMAPTILES_TOKEN
 
+        point = value
+        if isinstance(point, Point):
+            feature = OrderedDict()
```
goapunk, line 80: without this the map wouldn't render any more in the dashboard as the point wouldn't be in geojson

adhocracy4/maps/widgets.py
```diff
@@ -71,14 +75,23 @@ def render(self, name, value, attrs, renderer=None):
         if hasattr(settings, "A4_OPENMAPTILES_TOKEN"):
             omt_token = settings.A4_OPENMAPTILES_TOKEN
 
+        point = value
```
m4ra, line 78: why not just replace `value` with `point` above where the method is defined?

adhocracy4/maps_react/widgets.py
```diff
@@ -26,7 +32,16 @@ def render(self, name, value, attrs, renderer=None):
         }
 
         if value != "null" and value:
-            context["point"] = value
+            point = value
```
m4ra, line 35: same here, just replace the name `value` in the definition of the method.

adhocracy4/maps/mixins.py
```diff
@@ -63,13 +62,24 @@ def unpack_geojson(self, data):
...
 
-        return kwargs
+        # add geojson properties to widget
+        if "data" not in kwargs:
+            properties = self.get_geojson_properties()
+            if properties:
+                instance = kwargs["instance"]
+                props = OrderedDict()
+                for prop, mapping in properties.items():
+                    field = mapping if mapping else prop
```
m4ra, line 79: lil confusing here what's happening, what is the mapping or prop expected to be? Maybe add an inline comment

m4ra (commented):
just minor changes. You can merge once addressed.

adhocracy4/maps/mixins.py
```diff
@@ -13,7 +13,7 @@ class GeoJsonPointMixin:
     GeoJSON properties are extracted and mapped to the specified model fields.
     """
 
-    def get_properties(self):
+    def get_geojson_properties(self):
```
m4ra, line 16: also rename the docs string in line 12 to reflect the new method's name

adhocracy4/maps/widgets.py
goapunk: `value` comes from the super method, won't this break it if we change the name here?

m4ra: but you are replacing the super method, and it's just a positional argument

adhocracy4/maps_react/widgets.py
```diff
@@ -26,7 +32,16 @@ def render(self, name, value, attrs, renderer=None):
...
+            if isinstance(point, Point):
+                feature = OrderedDict()
+                feature["type"] = "Feature"
+                feature["geometry"] = json.loads(point.geojson)
+                if self.geo_json_properties:
```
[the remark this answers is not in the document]
goapunk: it's just the default, it will be overwritten by the mixin, see `mixins.py` line 85

adhocracy4/maps/widgets.py
goapunk: hmm, I just tested it and I did get an error: it's actually used as a named argument here: https://github.com/django/django/blob/main/django/forms/boundfield.py#L109

m4ra (dismissed):
Still needs some clarity the code.

adhocracy4/maps/mixins.py
```diff
@@ -45,12 +47,13 @@ def unpack_geojson(self, data):
                     point = geo_field
                 else:
                     point = json.loads(geo_field)
-                data = data.copy()
 
+                # data is mutable, so create a copy
```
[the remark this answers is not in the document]
goapunk: yes, sorry

adhocracy4/maps/mixins.py
```diff
@@ -91,7 +109,7 @@ def to_representation(self, instance):
                 feature["geometry"] = data[self.Meta.geo_field]
```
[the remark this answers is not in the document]
goapunk: but this one checks that the actual value of `data[self.Meta.geo_field]` isn't `None` vs `self.Meta.geo_field in data` checks if the key is in the dict? Or am I missing something?

PR #1780, opened 2025-03-06
follows: refactor FollowButton.js

hom3mad3:
**Description**
Refactor FollowButton to allow for alerts outside the container

### DIFF
--- a/adhocracy4/follows/static/follows/FollowButton.jsx
+++ b/adhocracy4/follows/static/follows/FollowButton.jsx
@@ -1,7 +1,7 @@
-import django from 'django'
 import React, { useState, useEffect } from 'react'
+import ReactDOM from 'react-dom'
+import django from 'django'
 import Alert from '../../../static/Alert'
-
 import api from '../../../static/api'
 import config from '../../../static/config'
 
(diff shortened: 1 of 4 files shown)


adhocracy4/follows/static/follows/FollowButton.jsx
```diff
@@ -18,73 +18,95 @@ const translated = {
...
+    
+    return ReactDOM.createPortal(
+      <Alert onClick={removeAlert} {...alert} />,
+      container
+    )
+  }
+
   return (
-    <span className="a4-follow">
+    <span className={`a4-follow ${useMbStyles ? 'a4-follow--proposal-style' : ''}`}>
```
vellip, line 84: Instead of asking for specific projects, can we not add like a `className` prop or `customClassName` or similar and then you pass `a4-follow--proposal-style` in MB as a prop

adhocracy4/follows/static/follows/FollowButton.jsx
```diff
@@ -18,73 +18,95 @@ const translated = {
...
+        type: follow.enabled ? 'success' : 'warning',
+        message: follow.enabled ? translated.followAlert : translated.followingAlert
       })
     })
   }
 
+  const buttonClasses = following ? 'a4-btn a4-btn--following' : 'a4-btn a4-btn--follow'
+
+  const AlertPortal = () => {
+    if (!alert || !alertTarget) return null
```
vellip, line 72: Should this actually be `&&` or `||`?

adhocracy4/follows/static/follows/FollowButton.jsx
```diff
@@ -18,73 +18,95 @@ const translated = {
...
+    
+    const container = document.getElementById(alertTarget)
+    if (!container) return null
```
vellip, line 75: Maybe add a `console.error` or throw an error, I think that's better for devs to understand where the issue is then.

vellip (approved):
Thank you! Looks good already, I just left some general questions. Also, should we add some tests?

adhocracy4/follows/static/follows/FollowButton.jsx
hom3mad3: oh good catch! that was unintended, i forgot to change it!

PR #1795, opened 2025-05-26
WIP: [#210] deps: update django to 5.2.1

shn-liqd:
**Tasks**
- [x] Changelog

### DIFF
--- a/changelog/210.md
+++ b/changelog/210.md
@@ -1,3 +1,3 @@
 ### Changed
 
-- Update Django version to 5.2.2
+- Update Django version to 5.2.1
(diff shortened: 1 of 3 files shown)


m4ra:
@shn-liqd we do upgrades incrementally, so for this task we need to upgrade first to 5.0, then 5.1 and finally to 5.2. For each upgrade we create separate commit, and we go through the [release notes](https://docs.djangoproject.com/en/5.2/releases/) to see what has been deprecated and need to change. Then for each upgrade we run the tests to check if any warnings still remain. I am making a documentation entry and will share it with you all soon. @partizipation

PR #1796, opened 2025-05-28
WIP: Possible Solutions for upgrading to Django 5.2.1

glanzel:
**Description**

Migration crashes when doing so with Django 5.2 caused by a not created index_together Index. (In Django 4 it still gets created).
I came up with the solution in 
[comments:migrations make migration work with django 5.2 and older version]

Afterwards ther are Still Warnings for other Apps that the old index_together instead of the new Meta.index is used. The Models are correct, recreating the migrations solves the problem. (But most likely, that is not a viable option.)
So i came up with recreating the index by hand: [actions:migrations: Forces from Index_together to new Meta.index by Hand]

PR just for information for more investigations

**Tasks**


glanzel:
Yes but as i wrote non of the models have Meta.index_together anymore.
The indextogether Warning is still there due to ther migrations.
I leave this PR open until i come up with a PR for 5.1 wich tackles these problems already

PR #1797, opened 2025-05-29
[ST-234] Polls:Export: Anonymous users are added to the Export 

glanzel:
**Tasks**
- [x ] PR name contains story or task reference
- [ x] Tests 
- [ x] Changelog

I attach the xls file with both registered and anonymous users. anonymous users are visible through the uuid instead of counting number. 

[test_20250528T161058.xlsx](https://github.com/user-attachments/files/20504294/test_20250528T161058.xlsx)

### DIFF
--- a/adhocracy4/polls/exports.py
+++ b/adhocracy4/polls/exports.py
@@ -52,18 +52,36 @@ class PollExportView(PermissionRequiredMixin, export_views.BaseItemExportView):
         return self.module
 
     def get_queryset(self):
-        creators_vote = poll_models.Vote.objects.filter(
-            choice__question__poll=self.poll
-        ).values_list("creator", flat=True)
-        creators_answer = poll_models.Answer.objects.filter(
-            question__poll=self.poll
-        ).values_list("creator", flat=True)
-        creator_ids = list(set(creators_vote).union(set(creators_answer)))
-        return User.objects.filter(pk__in=creator_ids)
+        return poll_models.Vote.objects.filter(choice__question__poll=self.poll)
+
+    def get_answers(self):
+        return poll_models.Answer.objects.filter(question__poll=self.poll)
+
+    def get_voters(self):
+        user_vote = (
+            self.get_queryset().exclude(creator=None).values_list("creator", flat=True)
+        )
+        user_answer = (
+            self.get_answers().exclude(creator=None).values_list("creator", flat=True)
+        )
+        user_ids = list(set(user_vote).union(set(user_answer)))
+        users = list(User.objects.filter(pk__in=user_ids))  # <-- User-Objekte!
+        anon_votes = (
+            self.get_queryset()
+            .filter(creator=None)
+            .values_list("content_id", flat=True)
+        )
+        anon_answers = (
+            self.get_answers().filter(creator=None).values_list("content_id", flat=True)
+        )
+        anon_ids = list(set(anon_votes).union(set(anon_answers)))
+        return users + anon_ids
 
     def get_object_list(self):
         # index is needed for (anonymous) user id
-        return [(index, user) for index, user in enumerate(self.get_queryset().all())]
+        list = [(index, user) for index, user in enumerate(self.get_voters())]
+        print(list)
+        return list
 
     @property
     def poll(self):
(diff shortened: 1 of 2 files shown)


adhocracy4/polls/exports.py
```diff
@@ -106,14 +124,24 @@ def get_field_data(self, item, field):
         index, user = item
 
         if field == "user_id":
-            value = index + 1
-
+            # Für anonyme Nutzer die UUID anzeigen, für registrierte wie bisher
+            if hasattr(user, "pk"):
+                value = index + 1
+            else:
+                value = str(user)  # UUID als String
```
m4ra, line 131: from what I understand, the field `user_id` is a counter of the users, so we don't need to add the value of the `content_id` in this field. And i don't see the `value` is called somewhere else, so is there a reason you are setting it here?

m4ra (changes requested):
Thank you @glanzel! Looks good, just a minor thing and a question. Also our comments shall be in English, so those need to be updated too.

adhocracy4/polls/exports.py
```diff
@@ -52,18 +52,36 @@ def get_permission_object(self):
...
+            self.get_answers().filter(creator=None).values_list("content_id", flat=True)
+        )
+        anon_ids = list(set(anon_votes).union(set(anon_answers)))
+        return users + anon_ids
 
     def get_object_list(self):
         # index is needed for (anonymous) user id
-        return [(index, user) for index, user in enumerate(self.get_queryset().all())]
+        list = [(index, user) for index, user in enumerate(self.get_voters())]
+        print(list)
```
m4ra, line 83: don't forget to remove this one

adhocracy4/polls/exports.py
```diff
@@ -52,18 +52,36 @@ def get_permission_object(self):
...
+        anon_answers = (
+            self.get_answers().filter(creator=None).values_list("content_id", flat=True)
+        )
+        anon_ids = list(set(anon_votes).union(set(anon_answers)))
+        return users + anon_ids
 
     def get_object_list(self):
         # index is needed for (anonymous) user id
-        return [(index, user) for index, user in enumerate(self.get_queryset().all())]
+        list = [(index, user) for index, user in enumerate(self.get_voters())]
```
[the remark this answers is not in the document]
partizipation: also just a style note, variable names should describe what they are used for. But I think both Mara and my comments here might be unnecessary, because I'm guessing this `list` was just for temporary use for debugging with the print statement :)

changelog/ST-234.md
```diff
@@ -0,0 +1,2 @@
+## Poll Module
+The non registered users are added to the Vote export. Votes and additional answers are included.
```
partizipation, line 2: Info on changelog style here: https://wiki.liqd.net/git?s[]=changelog

We have a script which concatenates changelogs before release which depends on things being under the e.g. "Added" or "Fixed" subheaders

adhocracy4/polls/exports.py
glanzel: i dont understand. 
as i understand by now the function get_field_data  walks throu every row of the export data and asks for the value for a given Field.
In the first Field "user_id" i set this value to a counter for real users (if) or the to the uuid if no user object is present ad there is only a uuid (else) 
value is the the return value which is placed in the first row (for the field user_id)

partizipation:
@glanzel check mattermost/signal lol

adhocracy4/polls/exports.py
```diff
@@ -106,14 +121,24 @@ def get_field_data(self, item, field):
         index, user = item
 
         if field == "user_id":
-            value = index + 1
-
+            # For anonymous Users Show UUID for registered show the count
+            if hasattr(user, "pk"):
+                value = index + 1
+            else:
+                value = str(user)
```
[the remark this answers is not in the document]
m4ra: and perhaps just add "ANON" or something in that direction.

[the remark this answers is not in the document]
glanzel: i though we could seperate them by the uuid but sure  i can write something else aswell.
ANON + index then ?

adhocracy4/polls/exports.py
```diff
@@ -106,14 +121,24 @@ def get_field_data(self, item, field):
         index, user = item
 
         if field == "user_id":
-            value = index + 1
-
+            # For anonymous Users Show Anon + count for registered show the count
+            if hasattr(user, "pk"):
+                value = index + 1
+            else:
+                value = "Anon" + str(index + 1)
```
m4ra, line 128: @t-j-q for the field of the users, when there are unregistered, we don't want to use their long UUID number, so is it fine to have the prefix "ANON"?

m4ra (approved):
Thank you!

PR #1798, opened 2025-06-05
[ST-210]  Part 1 - Upgrading to Django 5.0

glanzel:
**Tasks**
- [ x] Changelog

Only very small changes but it is needed for the upgrade of a-plus


partizipation:
LGTM!

PR #1799, opened 2025-06-05
[ST-210] Part 2 - Upgrading to Django 5.1

glanzel:
**Tasks**
- [ x] Changelog

** Changes **
Upgrading to Django 5.1 and removing index_together from the migration Files as this seem the only way to prevent migration under django 5.1 from crashing 

grafik


partizipation (approved):
LGTM

PR #1867, opened 2026-06-04
[ST-2119] polls: Allow Private Questions

glanzel:
**Tasks**
- [x] Changelog

### DIFF
--- a/adhocracy4/polls/migrations/0008_question_is_confidential.py
+++ b/adhocracy4/polls/migrations/0008_question_is_confidential.py
@@ -1,0 +1,23 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("a4polls", "0007_alter_poll_options"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="question",
+            name="is_confidential",
+            field=models.BooleanField(
+                default=False,
+                help_text=(
+                    "Individual responses are not shown publicly; only the number of "
+                    "submissions is visible."
+                ),
+                verbose_name="Confidential answers",
+            ),
+        ),
+    ]
--- a/adhocracy4/polls/serializers.py
+++ b/adhocracy4/polls/serializers.py
@@ -106,8 +106,9 @@ class QuestionSerializer(serializers.ModelSerializer):
 
     def get_answers(self, question: Question):
         answers = question.answers.all()
-        if question.is_confidential:
-            answers = self._filter_to_own_answers(answers)
+        if not question.is_confidential:
+            return AnswerSerializer(instance=answers, many=True).data
+        answers = self._filter_to_own_answers(answers)
         return AnswerSerializer(instance=answers, many=True).data
 
     def _filter_to_own_answers(self, answers):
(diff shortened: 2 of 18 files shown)


adhocracy4/polls/migrations/0008_question_is_confidential.py
```diff
@@ -0,0 +1,23 @@
...
+        ("a4polls", "0007_alter_poll_options"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="question",
+            name="is_confidential",
+            field=models.BooleanField(
+                default=False,
+                help_text=(
```
shn-liqd, line 16: Needs translation

adhocracy4/polls/static/PollDashboard/EditPollQuestion.jsx
```diff
@@ -34,6 +34,20 @@ export const EditPollQuestion = React.forwardRef((props, ref) => {
             ? <HelptextForm id={props.id} question={props.question} onHelptextChange={props.onHelptextChange} errors={props.errors} />
             : null}
 
+          <div className="form-check">
```
shn-liqd: looks like this can be a small component since its the same above?

adhocracy4/polls/serializers.py
```diff
@@ -102,6 +104,20 @@ def get_user_choices(self, question: Question) -> [int]:
...
+        answers = question.answers.all()
+        if question.is_confidential:
+            answers = self._filter_to_own_answers(answers)
+        return AnswerSerializer(instance=answers, many=True).data
+
+    def _filter_to_own_answers(self, answers):
+        if "request" not in self.context:
+            return answers.none()
+        user = self.context["request"].user
+        if user.is_authenticated:
```
shn-liqd, line 118: is user always an object here or can it be None?

adhocracy4/polls/serializers.py
```diff
@@ -111,6 +127,17 @@ def get_user_answer(self, question: Question) -> str | int:
 
     def get_other_choice_answers(self, question):
         other_choice_answers = question.other_choice_answers()
+        if question.is_confidential:
```
shn-liqd: Return early would be good here to improve readability

`if not question.is_confidential: return`

adhocracy4/polls/migrations/0008_question_is_confidential.py
glanzel: thats the migration file - we dont need transaltions here. the transaltion is in the models file

adhocracy4/polls/serializers.py
glanzel: User is always an Object, if not logged in its an AnonymousUser

adhocracy4/polls/static/PollDashboard/EditPollQuestion.jsx
glanzel: might be a good idea. i did that

adhocracy4/polls/serializers.py
glanzel: not really sure if this improves readabillity but i changed it

shn-liqd (approved):

